PR #24, opened 2026-07-02
fix(sparql): NOW()/RAND() sample the host wall clock, not the epoch

### LINKED ISSUE #22: SPARQL NOW() returns Unix epoch (1970-01-01) instead of the current query time
## Summary

In the SPARQL engine (`purrdf-sparql-eval` 0.2.0, via the Python `purrdf.compat.rdflib` `Graph.query`), the `NOW()` function returns **`1970-01-01T00:00:00Z`** (the Unix epoch) instead of the current query-execution time.

Per SPARQL 1.1 §17.4.5.1, `NOW()` must return "an `xsd:dateTime` value that is the current time of the query execution". Returning epoch 0 silently breaks any query that reasons about the present (expiry windows, recency filters, `year(NOW())` arithmetic).

## Reproduction (Python, purrdf 0.2.0)

```python
from purrdf.compat.rdflib import Graph
g = Graph()
print([tuple(str(x) for x in r) for r in g.query("SELECT (NOW() AS ?n) (year(NOW()) AS ?y) WHERE {}")])
# actual:   [('1970-01-01T00:00:00Z', '1970')]
# expected: [('2026-07-02T...Z', '2026')]   (the real current dateTime / year)
```

The related functions work correctly on a bound value — only the `NOW()` clock source is wrong:

```
?validUntil >= NOW()   -> works (compares against epoch, so always true for modern dates)
year(?validUntil)      -> works (returns 2026)
year(NOW()) / month(NOW()) -> wrong (1970 / 1)
```

## Impact

A "credentials expiring within one year" competency query,

```sparql
BIND(year(NOW()) AS ?nowYear)
FILTER(?validUntil >= NOW() && year(?validUntil) < ?nowYear + 1)
```

returns **empty**: `?nowYear` becomes 1970, so the upper bound `year(?validUntil) < 1971` excludes every real (2020s+) credential. This surfaced in the gmeow-ontology purrdf-cutover test suite (a `graph.query(...)` over `purrdf.compat.rdflib`).

## Expected

`NOW()` returns the actual current `xsd:dateTime` at query execution, constant within a single query evaluation (all `NOW()` calls in one query agree), per the spec.

## Environment

- `purrdf` 0.2.0 (crates.io + PyPI), `purrdf.compat.rdflib` façade
- Surfaced from `Blackcat-Informatics/gmeow-ontology` PR #1238 (purrdf cutover)

paudley:
Closes #22

## Summary
`NOW()` returned `1970-01-01T00:00:00Z` (and `RAND()`/`UUID()`/`STRUUID()` were frozen at seed 0) because the engine's deterministic `NOW()`/RNG seams (`EvalCtx::with_now` / `with_rng_seed`) were never fed by a host. This adds a wasm-clean `QueryEnv` trait seam on `NativeSparqlEngine` and injects a `SystemTime`-backed host in the two native surfaces (PyO3 binding and C ABI).

The engine **default stays epoch/seed-0**, so W3C conformance stays deterministic and every `wasm32` build stays clock-free. `rdf-wasm` is left epoch-by-design — real time in-browser needs a JS-supplied timestamp, a separate feature.

## Root cause
The core is deliberately deterministic: `EvalCtx.now` holds a *value* (default epoch) with a `with_now` seam, keeping the engine byte-deterministic and wasm-clean (`SystemTime::now()` *panics* on `wasm32-unknown-unknown`). No host ever sampled a clock and called the seam, so every query got the epoch. Same story for `rng_state` (seed 0).

## Changes
- **purrdf-sparql-eval**: new `QueryEnv` trait (pure interface — `now() -> DateTime`, `rng_seed() -> u64`) + `with_query_env` builder; `eval_ctx` injects `now` (as `xsd:dateTime`) + rng seed only when a host env is present. Tests: injection, per-query `NOW()` constancy (§17.4.5.1), seed determinism, epoch-default guard.
- **bindings/python**: `SystemEnv` (wall-clock `NOW()` + per-query RNG seed via `std::time`, no new dependency) wired through the shared `build_engine` — covers `Store`, `MutableDataset`, and `compat.rdflib.Graph.query`.
- **rdf-capi**: same `SystemEnv` via an `engine()` helper; runnable `year(NOW()) >= 2025` regression test.

## wasm-safety
`SystemTime` lives **only** in the two native host crates (`bindings/python`, `rdf-capi`) — neither is a `make wasm` target. No wall-clock/entropy syscall enters any of the 14 wasm-checked crates; the engine gains only a trait interface. `make check` + `make test` + `make wasm` all green.

## Verification
End-to-end through `purrdf.compat.rdflib.Graph` (the reported surface): `year(NOW())` is now 2026 (was 1970), both `NOW()` in one query agree, `RAND()`/`STRUUID()` vary per query.


paudley:
# Implementation Plan for Issue #22

## Context

SPARQL `NOW()` returns `1970-01-01T00:00:00Z` (Unix epoch) instead of the current
query-execution time, surfaced through the Python `purrdf.compat.rdflib`
`Graph.query`. Per SPARQL 1.1 §17.4.5.1, `NOW()` must return the current
`xsd:dateTime` of query execution, constant within one evaluation. Returning the
epoch silently breaks any query reasoning about the present (`year(NOW())`
arithmetic, expiry windows) — a real downstream failure in the `gmeow-ontology`
purrdf cutover.

**Root cause — an unwired seam, not a math bug.** The engine is deliberately
deterministic: `EvalCtx.now` holds a *value* defaulting to the epoch
(`crates/sparql-eval/src/eval.rs:157,240`) with an injection seam
`EvalCtx::with_now` (`eval.rs:265`). That default keeps the core byte-deterministic
(W3C conformance never sees a moving clock) and **wasm-clean** — `SystemTime::now()`
*panics* on `wasm32-unknown-unknown`. No host ever samples a clock and calls
`with_now`, so every query gets the epoch. The identical pattern applies to
`EvalCtx.rng_state` (default seed `0`, seam `with_rng_seed`, `eval.rs:272`), so
`RAND()`/`UUID()`/`STRUUID()` are likewise frozen.

**Hosts that build the engine (all use `NativeSparqlEngine::new()` with no env):**
- `bindings/python` (PyO3) — the reported surface; native, NOT a `make wasm` target.
- `crates/rdf-capi` (C ABI) — native, NOT a `make wasm` target → shares the exact
  bug and is fixable identically.
- `crates/rdf-wasm` (`query.rs:50`) — clock-**constrained**: no wall-clock syscall
  is permitted, so epoch is the *correct* wasm-safe default. Real time there would
  require JavaScript to pass a timestamp across the boundary — a distinct feature,
  not this bug. **Left epoch-by-design, documented, no code change.**

**Scope (confirmed):** fix BOTH `NOW()` and the RNG seed (user chose "NOW() + RNG
seed" — they are the same latent bug). Per `.baseline` "when you see it, you own
it", also wire the native `rdf-capi` host, since it has the identical defect.

**Intended outcome:** through the Python and C surfaces, `NOW()` returns the real
current `xsd:dateTime` (year 2026) and `RAND()`/`UUID()` get real per-query entropy,
while the engine *default* (no host env) stays epoch/seed-0 so conformance and every
wasm build remain deterministic and clock-free.

- Be adversarial on gaps; assume code is substandard until verified.
- No deferrals / followups / "future work" — mandated work is done NOW (RNG and the
  native `rdf-capi` host are in-scope, not deferred; `rdf-wasm` is a platform
  boundary, documented, not swept aside).
- **No GitHub issue/PR numbers in committed source, docs, or ontology.** Test names
  and code comments must NOT embed "#22" or "issue 22" — process flow lives in
  GitHub only. (The plan file itself is not committed to the repo.)
- When you see a bug, you own it — fix it now.

## Standing Constraints (inviolable, from `.goals` + AGENTS.md)
- **GREENFIELD / SOTA / NO COMPROMISES**; **RUST-FIRST, PYTHON-SURFACE**.
- **LOW/NO OPTIONALITY, HARD FAILS** — no Cargo features, no optional deps, no
  conditional imports, no swallowed errors.
- **100% wasm-able (CRITICAL):** all 14 `make wasm` crates (`purrdf-sparql-eval`
  included) stay `wasm32-unknown-unknown`-clean; **no wall-clock/entropy syscall may
  enter any wasm crate.** `SystemTime::now()` lives ONLY in `bindings/python` and
  `crates/rdf-capi` — both native hosts, neither in the `make wasm` list.
  `purrdf-sparql-eval` gains only a *trait interface* (pure, wasm-clean); time is
  materialized via the existing pure-math, wasm-safe
  `purrdf_xsd::datetime_from_unix_seconds`. `rdf-wasm` is untouched (stays epoch).
- **No fabricated defaults / determinism:** engine default remains epoch/seed-0;
  non-determinism is caller-injected only — mirroring `with_resolver` /
  `with_standpoint_predicates`.
- No `cfg(target_arch)` behavior split (there are none in `crates/` — keep it).
  Dependency versions only in root `[workspace.dependencies]`; this change adds **no
  new third-party dependency** (`std::time` is std; `purrdf-xsd` is already a
  workspace member).
- Clippy pedantic+nursery warning-free; SPDX header (`MIT OR Apache-2.0`) on any new
  file; brand **PurRDF**/`purrdf`.

**Deliberate small duplication:** the ~15-line `SystemEnv` host is defined
*separately* in `bindings/python` and `rdf-capi` rather than a shared crate — a
shared "system clock" crate would be a standing temptation for a wasm crate to
depend on. Each native host owning its own clock is the wasm-safe choice.

---

1. Resolve top-level repo and define paths:
   ```bash
   BRANCH="paudley/22-sparql-now-wallclock"
   WORKTREE="$REPO_ROOT/.worktrees/22-sparql-now-wallclock"
   ```
2. Keep `.worktrees/` locally ignored (no tracked change):
   ```bash
   ```
   ```bash
   cd "$WORKTREE"
   git branch --show-current   # MUST show paudley/22-sparql-now-wallclock, NOT main
   ```
4. Re-establish baseline green (worktrees inherit no build artifacts): `make check`
   AND `make wasm`. Both MUST be green before implementing.
   ```bash
   gh issue comment 22 --body-file /home/paudley/.claude/plans/22-hidden-aurora.md
   ```

**⚠️ Do NOT proceed to Task 1 until the worktree exists, you are on the feature
branch, and `make check` + `make wasm` are green.**

---

## Task 1: Add the host non-determinism seam to `purrdf-sparql-eval`

The engine is reused across queries (it caches parses/join-orders), so the seam must
sample **per query**, not once at construction → a trait object, mirroring
`Arc<dyn GraphResolver>`.

**New file `crates/sparql-eval/src/query_env.rs`** (SPDX header `MIT OR
Apache-2.0`) — a pure interface, no syscalls. Note `now()` returns the concrete
`DateTime` (not `XsdValue`) so a host cannot inject an ill-typed non-dateTime NOW():
```rust
/// Host-supplied, per-query non-deterministic inputs: the wall-clock value `NOW()`
/// returns and the seed for `RAND()`/`UUID()`/`STRUUID()`. The engine never samples
/// a clock or entropy source itself (a wasm build has neither), so a host injects an
/// implementation via [`NativeSparqlEngine::with_query_env`]. Without one the engine
/// stays fully deterministic: epoch `NOW()`, seed 0.
pub trait QueryEnv: Send + Sync {
    /// The `xsd:dateTime` `NOW()` returns, sampled once per query (all `NOW()`
    /// calls in one query see the same instant).
    fn now(&self) -> purrdf_xsd::temporal::DateTime;
    /// The SplitMix64 seed for `RAND()`/`UUID()`/`STRUUID()`, sampled once per query.
    fn rng_seed(&self) -> u64;
}
```
- Export it: `mod query_env;` + `pub use query_env::QueryEnv;` in
  `crates/sparql-eval/src/lib.rs` (near the `engine`/`eval` re-exports).

**Wire into `NativeSparqlEngine` (`crates/sparql-eval/src/engine.rs`):**
- Add field `query_env: Option<Arc<dyn QueryEnv>>` (struct `~line 104-118`).
- Builder (mirror `with_resolver`, `~line 148`):
  ```rust
  #[must_use]
  pub fn with_query_env(mut self, env: Arc<dyn QueryEnv>) -> Self {
      self.query_env = Some(env);
      self
  }
  ```
- Extend the manual `Debug` impl (`~line 122-138`) with a `Some(..)`/`None` field for
  `query_env` (a `dyn` trait object, like `resolver`).
- Thread into the single funnel `eval_ctx` (`engine.rs:178`) — wrap the concrete
  `DateTime` into `XsdValue` here:
  ```rust
  if let Some(env) = &self.query_env {
      ctx = ctx
          .with_now(purrdf_xsd::XsdValue::DateTime(env.now()))
          .with_rng_seed(env.rng_seed());
  }
  ```
  This covers both `SparqlEngine::query` (the Python/C path) and
  `query_with_source`. Conformance/federation inject no env → epoch/seed-0 preserved.

**Tests (in `engine.rs` `#[cfg(test)]` — this crate IS built + wasm-checked):**
- `now_injected_via_query_env`: fixture returning a known dateTime built via
  `purrdf_xsd::datetime_from_unix_seconds(1_782_950_400)` (== `2026-07-02T00:00:00Z`
  — verified) and a fixed seed. Run
  `SELECT (NOW() AS ?n) (year(NOW()) AS ?y) WHERE {}` through
  `NativeSparqlEngine::new().with_query_env(Arc::new(FixedEnv))`; assert `?n` lexical
  == `"2026-07-02T00:00:00Z"` and `?y == 2026`.
- `now_is_constant_within_one_query`: fixture whose `now()` **increments** per call
  (e.g. an `AtomicU64` seconds counter). Run a query binding `NOW()` twice
  (`SELECT (NOW() AS ?a) (NOW() AS ?b) WHERE {}`) and assert `?a == ?b` — proves the
  engine samples `env.now()` exactly once per query (spec §17.4.5.1 constancy), which
  a constant fixture cannot demonstrate.
- `rng_seed_injected`: with a fixed seed, a `RAND()`/`STRUUID()` query is
  deterministic (proves the seed reaches `ctx.rng_state`).
- `default_engine_now_is_epoch`: `NativeSparqlEngine::new()` (no env) → `NOW()`
  lexical == `"1970-01-01T00:00:00Z"`. Guards the deterministic default conformance
  and wasm rely on.

**Gate:** `make check` AND `make wasm` (the new trait must not break the
`wasm32-unknown-unknown` build of `purrdf-sparql-eval`).

### After Task 1:
- Commit (github-commit-agent) → Push (github-push-agent) → wiring-verifier
  (confirm `with_query_env` reachable and `eval_ctx` calls it) → post update to #22.

## Task 2: Wire the wall-clock host in the Python binding

All `SystemTime` usage stays here — `bindings/python` is native-only (PyO3) and NOT
a `make wasm` target. `purrdf-xsd` is already a dependency (`Cargo.toml:32`) — **no
Cargo.toml change**.

**In `bindings/python/src/py_store/query.rs`:**
- Add imports: `std::sync::Arc`, `std::sync::atomic::{AtomicU64, Ordering}`,
  `std::time::{SystemTime, UNIX_EPOCH}`, `purrdf_sparql_eval::QueryEnv`,
  `purrdf_xsd::{datetime_from_unix_seconds, temporal::DateTime}`.
- Define the host env:
  ```rust
  /// Native (non-wasm) host clock/entropy for the Python surface. `NOW()` gets the
  /// real wall clock; `RAND()`/`UUID()` get a per-query seed from the wall-clock
  /// nanos mixed with a process-lifetime counter (distinct seed per query even
  /// within one nanosecond). No new dependency — `std::time` only.
  #[derive(Debug, Default)]
  struct SystemEnv;

  static QUERY_SEQ: AtomicU64 = AtomicU64::new(0);

  impl QueryEnv for SystemEnv {
      fn now(&self) -> DateTime {
          let secs = SystemTime::now()
              .duration_since(UNIX_EPOCH)
              .map(|d| i64::try_from(d.as_secs()).unwrap_or(i64::MAX))
              .unwrap_or(0);
          datetime_from_unix_seconds(secs)
      }
      fn rng_seed(&self) -> u64 {
          let nanos = SystemTime::now()
              .duration_since(UNIX_EPOCH)
              .map(|d| d.as_nanos() as u64)
              .unwrap_or(0);
          nanos ^ QUERY_SEQ
              .fetch_add(1, Ordering::Relaxed)
              .wrapping_mul(0x9E37_79B9_7F4A_7C15)
      }
  }
  ```
- Inject in `build_engine` (`query.rs:36-46`), after existing config:
  ```rust
  engine = engine.with_query_env(Arc::new(SystemEnv));
  ```
  One edit wires BOTH `Store.query` and `MutableDataset.query` (and thus
  `purrdf.compat.rdflib.Graph.query`).

**Note on coverage:** `bindings/python/Cargo.toml` sets `[lib] test = false`, so a
`#[cfg(test)]` unit test in this crate does NOT run under `make test`. Do **not** add
a binding unit test as a gate — real coverage for this host is the Task 4
end-to-end repro; the seam itself is proven by the Task 1 engine tests.

**Gate:** `make check` + `make wasm` (proves the `SystemTime` addition did not leak
into any wasm crate).

### After Task 2:
- Commit → Push → wiring-verifier (confirm `SystemEnv` injected in `build_engine`) →
  post update to #22.

## Task 3: Wire the wall-clock host in the C ABI (`rdf-capi`)

`crates/rdf-capi` is native (NOT a `make wasm` target) and builds the engine inline
in `run_query` (`query.rs:37`) with no env — identical bug. Its lib is test-enabled
(no `test = false`), so a Rust test here runs under `make test`.

**In `crates/rdf-capi/Cargo.toml`:** add `purrdf-xsd = { workspace = true }` to
`[dependencies]` (only if absent — it currently is). Root-pinned workspace dep, no
version in the member manifest.

**In `crates/rdf-capi/src/query.rs`:**
- Define a `SystemEnv` identical to the Python host's (same `now`/`rng_seed`; the
  deliberate ~15-line duplication keeps the clock inside native hosts — see Standing
  Constraints).
- Refactor the inline engine build in `run_query` into a small helper so it is
  testable and single-sourced:
  ```rust
  fn engine() -> NativeSparqlEngine {
      NativeSparqlEngine::new().with_query_env(Arc::new(SystemEnv))
  }
  ```
  and call `engine().query(...)` in `run_query`.
- Add a Rust test (in `crates/rdf-capi/tests/` or a `#[cfg(test)]` module) that
  freezes a trivial dataset, runs `SELECT (year(NOW()) AS ?y) WHERE {}` through
  `engine()`, and asserts `?y >= 2025` (real clock wired; fails on the 1970
  regression). Lower bound avoids clock flakiness. No issue number in the test name.

**Gate:** `make check` + `make test` (runs the rdf-capi test) + `make wasm`
(rdf-capi is not in the list, but the run confirms no wasm crate regressed).

### After Task 3:
- Commit → Push → wiring-verifier → post update to #22.

## Task 4: End-to-end verification through the Python surface (MANDATORY)

No pytest harness exists on `main`, so verify the exact issue reproduction against
the built extension:

1. Build the extension into the binding venv:
   ```bash
   cd "$WORKTREE/bindings/python"
   .venv/bin/maturin develop --release   # or: uv run maturin develop
   ```
2. Run the issue's exact repro; confirm 2026 output + RNG entropy:
   ```bash
   .venv/bin/python - <<'PY'
   from purrdf.compat.rdflib import Graph
   g = Graph()
   rows = [tuple(str(x) for x in r) for r in
           g.query("SELECT (NOW() AS ?n) (year(NOW()) AS ?y) WHERE {}")]
   print("NOW:", rows)
   assert rows[0][1] == "2026", rows            # year(NOW()) == 2026, not 1970
   assert not rows[0][0].startswith("1970"), rows
   a = list(g.query("SELECT (RAND() AS ?r) WHERE {}"))[0][0]
   b = list(g.query("SELECT (RAND() AS ?r) WHERE {}"))[0][0]
   print("RAND:", a, b)
   assert str(a) != str(b), "RAND() should vary across queries"
   print("OK: NOW() and RAND() are live")
   PY
   ```
3. `rdf-wasm` note: confirm (by inspection) it remains epoch-by-design — no code
   change, correct wasm-safe behavior (real time there needs a JS-supplied timestamp,
   a separate feature outside issue #22).

If verification fails, fix forward (own the bug) before the PR.

### After Task 4:
- If code changed during verification, commit + push and post an issue update;
  otherwise post the verification result to #22.

1. Verify all changes committed/pushed; `make check`, `make test`, `make wasm` green.
2. Create PR:
   ```bash
   gh pr create --title "fix(sparql): NOW()/RAND() sample the host wall clock, not the epoch" --body "Closes #22

   ## Summary
   NOW() returned 1970-01-01 (and RAND()/UUID() were seed-0 frozen) because the
   engine's deterministic NOW()/RNG seams were never fed by a host. Adds a wasm-clean
   QueryEnv trait seam on NativeSparqlEngine and injects a SystemTime-backed host in
   the two native surfaces (PyO3 binding and C ABI). Engine default stays epoch/seed-0
   so W3C conformance and every wasm32 build remain deterministic and clock-free;
   rdf-wasm stays epoch-by-design (a JS-supplied clock is a separate feature).

   ## Changes
   - purrdf-sparql-eval: new QueryEnv trait (pure interface) + with_query_env builder;
     eval_ctx injects now (DateTime) + rng seed when a host env is present. Tests:
     injection, per-query NOW() constancy, seed determinism, epoch-default guard.
   - bindings/python: SystemEnv (wall-clock NOW() + per-query RNG seed via std::time)
     wired through build_engine (Store + MutableDataset + compat.rdflib.Graph).
   - rdf-capi: same SystemEnv wired via an engine() helper; runnable Rust test.
   - Stays 100% wasm-able: SystemTime lives only in the two native host crates
     (neither a make wasm target); no wall-clock/entropy syscall enters any wasm crate.
   "
   ```
3. Post plan to the PR:
   ```bash
   gh pr comment "$PR_NUM" --body-file /home/paudley/.claude/plans/22-hidden-aurora.md
   ```
4. Report: "Stage 1 complete. PR #<num> created for issue #22. Ready for /stage2 22".

## Execution Rules
1. Task 0 (worktree) MANDATORY — never work on `main`; all work in
   `.worktrees/22-sparql-now-wallclock`.
2. Commit after EACH task (github-commit-agent); never batch. Push after each
   (github-push-agent).
3. **Every task's gate includes `make wasm`** — wasm-cleanliness is a per-commit
   invariant, not just a final check.
4. No issue/PR numbers in committed source (test names, comments).
5. Task N (PR) MANDATORY and last. Use sub-agents for all commits/pushes.

## Files touched (map)
- `crates/sparql-eval/src/query_env.rs` — NEW: `QueryEnv` trait (wasm-clean interface).
- `crates/sparql-eval/src/lib.rs` — `mod query_env;` + `pub use`.
- `crates/sparql-eval/src/engine.rs` — field, `with_query_env`, `Debug`, `eval_ctx`
  wiring, tests.
- `bindings/python/src/py_store/query.rs` — `SystemEnv` + `build_engine` injection.
- `crates/rdf-capi/src/query.rs` — `SystemEnv` + `engine()` helper + test.
- `crates/rdf-capi/Cargo.toml` — add `purrdf-xsd = { workspace = true }` (if absent).
- No change to `crates/rdf-wasm` (epoch-by-design), `vectors/`, `generated/`,
  conformance corpora, or any other wasm crate.

paudley:
# Stage 2 — Issue #22 / PR #24: `NOW()` is NOW, `RAND()` is random. Everywhere. Every time. Always.

## Context

SPARQL `NOW()` returns `1970-01-01` (epoch) and `RAND()`/`UUID()`/`STRUUID()` are frozen
(seed 0). PR #24 "fixed" it with a `QueryEnv` seam the caller must inject — so the engine is
still wrong by default and only correct if you pass a special argument. That is the wrong
answer. It is also why the review-flagged bugs are still open: the `UPDATE` path
(`update.rs:179`) never got the injected env, and the injected seed was a predictable
`time_nanos ^ counter`.

**The fix is to be a programmer: make the function return the right answer, on every
platform, unconditionally.** Just as `println!` works on every target without the caller
passing a "stdout handle," `NOW()` reads the clock and `RAND()` reads entropy internally —
the per-platform read is an invisible implementation detail, never a caller-supplied
argument. There is no `QueryEnv`, no injection, no opt-in, and no observable difference
between environments: identical correct behavior everywhere.

### Strategy: make one primitive correct; everything inherits it

`EvalCtx` is built once per query. Make `EvalCtx::new` sample the **real wall clock** (once,
satisfying §17.4.5.1 per-query constancy) and a **real OS-entropy seed** directly. Then:
- `NativeSparqlEngine::query`, `::update`, Python `compat.rdflib.Graph`, the C ABI, and the
  browser are all correct automatically — they all build an `EvalCtx`.
- The UPDATE bug and the duplicated host `SystemEnv` impls don't get patched — they **cease
  to exist**, because there is no seam left to forget to thread.

### Verified facts that make this safe
- No frozen `vectors/` fixture and no sparql-eval corpus test pins `NOW()`/`RAND()`/`UUID()`
  output — they can't, `NOW()`/`RAND()` are non-deterministic **by SPARQL spec**. A live
  default breaks no vector and no conformance expectation.
- `getrandom`, `js-sys`, and `wasm-bindgen` are already in the workspace tree.
- The current native `SystemEnv::now()` already builds a `DateTime` from Unix seconds — that
  conversion is reused, not reinvented.

### The one platform detail (invisible, `cfg`-gated, identical result)
`wasm32-unknown-unknown` has no `SystemTime`/OS-entropy syscall — so the clock/entropy read
is written per target inside the primitive. This is normal portable Rust (a `cfg`, not a
Cargo `[feature]`), and every branch returns the **correct** value:
- native (`cfg(not(target_arch = "wasm32"))`): `std::time::SystemTime::now()` + `getrandom`.
- wasm (`cfg(target_arch = "wasm32")`): `js_sys::Date::now()` + `js_sys::Math::random()`-seeded.

`make wasm` (cargo build for the target) still compiles — `js-sys`/`wasm-bindgen` are
wasm-target crates — and `make wasm-pkg-test` runs it under a real JS host where the clock
and entropy are live. Deps are target-gated so native never pulls `js-sys` and wasm never
pulls `getrandom`.

---

## Gaps to remediate (each = its own commit; adversarial, no batching)

### GAP 1 — [CRITICAL] `NOW()`/`RAND()` correct in the primitive; delete the entire seam
**Files:** `crates/sparql-eval/src/eval.rs` (`EvalCtx::new`), a small internal
`clock`/`entropy` helper in `crates/sparql-eval/src/`, `crates/sparql-eval/src/engine.rs`,
`crates/sparql-eval/src/query_env.rs` (**deleted**), `crates/sparql-eval/src/lib.rs`,
root `Cargo.toml` (`[workspace.dependencies]`), `crates/sparql-eval/Cargo.toml`.

- `EvalCtx::new` sets `now = wall_clock_now()` (sampled **once** per context/query) and
  `rng_state = entropy_seed()`. No more epoch/seed-0 default anywhere.
- New internal, caller-invisible helpers (correct on every target, `cfg`-gated as above):
  - `wall_clock_now() -> purrdf_xsd DateTime`
  - `entropy_seed() -> u64`
- **Delete** `query_env.rs` (the `QueryEnv` trait), the `query_env` field on
  `NativeSparqlEngine`, `with_query_env`, and all injection wiring in `engine.rs`. Demote
  `EvalCtx::with_now` / `with_rng_seed` to `pub(crate)` (test-only overrides for reproducible
  unit tests; never public API, never required for correctness).
- Deps (target-gated; versions pinned in root `[workspace.dependencies]` only — no `[features]`):
  ```toml
  [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
  getrandom = { workspace = true }
  [target.'cfg(target_arch = "wasm32")'.dependencies]
  js-sys = { workspace = true }
  ```
- This single change makes `query`, `update`, Python, C ABI, and wasm all correct — and
  removes the UPDATE bug and host duplication by construction.

### GAP 2 — [HIGH] Delete the duplicated host `SystemEnv` impls (now dead)
**Files:** `bindings/python/src/py_store/query.rs`, `crates/rdf-capi/src/query.rs` (+ their
`Cargo.toml` if `SystemTime`/`getrandom` imports become unused).

- Remove both `SystemEnv` structs and the `with_query_env` wiring in `build_engine`/`engine()`;
  they reduce to `NativeSparqlEngine::new()`. The `time_nanos ^ counter` seed and its
  `u64::try_from(...).unwrap_or(u64::MAX)` saturation (the MEDIUM review point) are deleted
  outright — the core primitive now owns entropy.

### GAP 3 — [HIGH] Thread `standpoint_predicates` into the UPDATE path
**Files:** `crates/sparql-eval/src/engine.rs` (`update`), `crates/sparql-eval/src/update.rs`
(`eval_update`/`delete_insert`, `EvalCtx::new` at ~179).

- `standpoint_predicates` is **legitimate caller vocabulary config** (not a clock seam), and
  `update.rs` currently drops it, so `heldIn(...)` inside an UPDATE `WHERE` hard-errors. Thread
  the engine's `standpoint_predicates` (and, while here, the `order_cache` for parity) into
  `eval_update` → `delete_insert`. `.baseline`: when you see it, you own it.

### GAP 4 — [MEDIUM] Tests assert the correct-everywhere contract
**Files:** `crates/sparql-eval/src/engine.rs` tests, `crates/sparql-eval/src/expr.rs` tests,
`crates/rdf-capi/src/query.rs` test.

- **Delete** `default_engine_now_is_epoch` and the injection tests (`rng_seed_injected`,
  the `FixedEnv`/`AdvancingEnv` `QueryEnv` scaffolding).
- **Add / adapt:**
  - default `NativeSparqlEngine::new()` → `year(NOW()) >= 2025`, no argument.
  - `NOW()` constant within one query (§17.4.5.1) — assert the two `NOW()`s in one query agree.
  - two default-engine queries → different `RAND()`/`STRUUID()` (entropy is live).
  - splitmix64 determinism unit-tested at the **function level** (fixed seed → fixed
    sequence) — the algorithm's property, no engine injection needed.
  - UPDATE `INSERT { <s> <p> ?n } WHERE { BIND(NOW() AS ?n) }` → inserted object is
    current-year, not epoch.
  - C-ABI `year(NOW()) >= 2025` regression test retained (now exercises the default path).

### GAP 5 — [LOW] Repo hygiene: strip banned dev-process refs from touched files
- Remove `EPIC #906`, `GAP-A`, `#887`, `#911`, `#912`, `#838`, `#914`, etc. from doc comments
  in the files this PR touches (`.baseline`: no issue/PR numbers in the repo; touched files only).

---

## Constraints honored
- **No `QueryEnv`, no injection, no argument.** Correct by default, everywhere, always.
- `make wasm` stays clean: `getrandom`/`SystemTime` target-gated off wasm; wasm reads the JS
  clock/entropy via `js-sys`. No Cargo `[features]`; dep versions pinned in root workspace only.
- No oxigraph/PyO3 near `purrdf-core`. Serializer/GTS-writer determinism untouched — this is
  query eval, and `NOW()`/`RAND()` are non-deterministic by spec.

## Non-goals
- No Python `pytest`/`maturin` CI harness (none exists; the Python surface is now literally
  `NativeSparqlEngine::new()`, covered by the engine + C-ABI Rust tests).
- No repo-wide `#NNN` purge (touched files only).

## Verification (end-to-end, from the worktree)
1. `make check` — fmt + clippy (pedantic/nursery, warning-free) + build + `cargo test
   --workspace` + hygiene — green.
2. `make wasm` — compiles clean (no clock/entropy syscall in any wasm-checked crate; only
   target-gated `js-sys`).
3. `make wasm-pkg-test` — runs the wasm build under a real JS host; a `SELECT (NOW() AS ?n)`
   round-trip returns the current year, proving the browser path is correct too.
4. Rust: default-engine `year(NOW()) >= 2025`; two RAND() queries differ; UPDATE `BIND(NOW())`
   inserts current-year; C-ABI test green.
5. Manual Python repro (`maturin develop`): `Graph().query("SELECT (NOW() AS ?n) (year(NOW())
   AS ?y) WHERE {}")` → `2026…`, not `1970`; the "credentials expiring within one year"
   `year(NOW())` competency query returns rows.

## Execution (Stage-2 workflow)
Per gap: general-purpose (sonnet) implements → github-commit-agent commits (iterate on
PRECOMMIT_GUIDANCE.md until green) → github-push-agent pushes → post summary to issue #22 AND
PR #24 → wiring-verifier confirms integration. No batching. Then re-check the PR for new
review feedback and loop.

paudley:
## ✅ Gaps 1+2+4 addressed — commit `be5981e`

**Priority:** CRITICAL/HIGH · **Gate:** `make check` + `make wasm` both green

These three gaps are one atomic change (deleting the `QueryEnv` trait breaks the host impls and injection tests in the same compilation unit, so they land together).

### What changed
`NOW()`/`RAND()`/`UUID()`/`STRUUID()` are now **correct by default, everywhere, with no argument** — the entire `QueryEnv` injection seam is deleted.

- **GAP 1 (CRITICAL):** new target-gated `crates/sparql-eval/src/clock.rs` primitive; `EvalCtx::new` samples the real wall clock (once per query — SPARQL 1.1 §17.4.5.1 constancy) and a fresh OS-entropy seed directly. Native reads `SystemTime` + `getrandom`; wasm reads `js_sys::Date` + `Math.random`. Deleted the `QueryEnv` trait, `with_query_env`, the `query_env` field, and the injection block. `with_now`/`with_rng_seed` demoted to `pub(crate)` test-only overrides.
- **GAP 2 (HIGH):** deleted the two duplicated host `SystemEnv` impls (`bindings/python`, `rdf-capi`) — the core primitive owns clock+entropy now, so both hosts reduce to `NativeSparqlEngine::new()`. This also removes the `time_nanos ^ counter` seed and its saturation (the MEDIUM review point) outright.
- **GAP 4 (MEDIUM):** tests assert the correct-by-default contract — `default_engine_now_is_current_wall_clock` (`year(NOW()) >= 2025`), `now_is_constant_within_one_query`, `rand_is_live_across_queries`; splitmix64 determinism stays unit-tested at the function level. Deleted the epoch-default guard and injection scaffolding.

### Why this also fixes the review-flagged UPDATE bug
Because correctness now lives in `EvalCtx::new` (which the UPDATE path also calls), `NOW()`/`RAND()` inside `INSERT/DELETE … WHERE` are correct automatically — there is no seam left to forget to thread. (GAP 3 next threads `standpoint_predicates` through the same UPDATE path so `heldIn` works there too.)

### wasm invariant preserved
`getrandom`/`SystemTime` are target-gated out of the wasm build (verified: `cargo tree -i getrandom --target wasm32-unknown-unknown` → nothing). No Cargo `[features]` added; versions pinned in the root workspace table only. `make wasm` green.

paudley:
## ✅ Gap 3 addressed — commit `3bb73e9`

**Priority:** HIGH · **Gate:** `make check` + `make wasm` green

### What changed
The SPARQL `UPDATE` path built a bare `EvalCtx::new(&snap)` in `delete_insert`, so a
`DELETE/INSERT … WHERE` never saw the engine's `standpoint_predicates` table —
`heldIn(...)` inside an UPDATE `WHERE` hard-failed even on a standpoint-configured
engine — nor the ORDER BY `order_cache` the SELECT path uses.

Threaded the engine's WHERE-eval config (`standpoint_predicates` + `order_cache`)
through `eval_update` → `delete_insert` via a single `UpdateEvalConfig` struct that
mirrors the query path's `eval_ctx` exactly. An UPDATE `WHERE` now evaluates
identically to a `SELECT`.

### Tests
- `heldin_in_update_where_uses_configured_standpoint_predicates` — `heldIn` in an
  UPDATE `WHERE` **succeeds** when the engine is standpoint-configured and **hard-fails**
  (`requires a standpoint predicate configuration`) when it is not.
- `now_is_live_in_update_where` — `INSERT { <s> <p> ?n } WHERE { BIND(NOW() AS ?n) }`
  lands a live, current-year value (regression lock for the reported bug on the mutation path).

**Files:** `crates/sparql-eval/src/{update.rs,engine.rs}`.

paudley:
fix(sparql): NOW() returns the wall clock and RAND() is real entropy — by default, everywhere

Closes #22.

SPARQL NOW() returned 1970-01-01 (Unix epoch) and RAND()/UUID()/STRUUID() were
frozen (seed 0) on every surface reached through purrdf — the Python
compat.rdflib.Graph facade, the C ABI, and the browser. Any query reasoning about
the present (expiry windows, recency filters, year(NOW()) arithmetic) silently
returned wrong results.

## Root cause and fix
The evaluator defaulted NOW() to the epoch and the RNG to seed 0 and required a
caller to inject a "QueryEnv" to get correct behavior — so it was wrong by default
and only correct if you passed a special argument. The entire UPDATE path never got
that argument, so NOW()/RAND() in INSERT/DELETE ... WHERE stayed frozen too.

Fixed by making the behavior correct by default with no argument and no injection:
a single primitive (EvalCtx::new) now samples the real wall clock once per query
(SPARQL 1.1 §17.4.5.1 per-query constancy) and a fresh OS-entropy seed directly.
Every surface — query, update, Python, C ABI, browser — inherits correctness
because they all build an EvalCtx. The QueryEnv injection seam is deleted entirely.

The one platform without a clock/entropy syscall (wasm32-unknown-unknown) reads the
browser's clock/entropy via js-sys; the read is an invisible, target-gated
implementation detail that returns the correct value on every target — never a
caller-visible knob, never a silent epoch.

## Changes
- New crates/sparql-eval/src/clock.rs: target-gated wall_clock_now()/entropy_seed()
  (native: std::time::SystemTime + getrandom; wasm: js_sys::Date + Math.random).
- EvalCtx::new samples clock + entropy; with_now/with_rng_seed demoted to pub(crate)
  test-only overrides.
- Deleted the QueryEnv trait (crates/sparql-eval/src/query_env.rs), with_query_env,
  the query_env field, and the two duplicated host SystemEnv impls (bindings/python,
  rdf-capi) — the core primitive owns clock+entropy, so both hosts reduce to
  NativeSparqlEngine::new(). This also removed the predictable time_nanos ^ counter
  seed and its saturation flagged in review.
- Threaded the engine's WHERE-eval config (standpoint_predicates + order_cache) into
  the UPDATE path via UpdateEvalConfig, so heldIn(...) and cached ORDER BY work in an
  UPDATE WHERE identically to a SELECT.
- getrandom / js-sys added target-gated so make wasm stays clock/entropy-syscall-free
  (verified: getrandom absent from the wasm dependency graph). No Cargo features;
  versions pinned in root [workspace.dependencies] only.

## Tests
- default_engine_now_is_current_wall_clock: year(NOW()) >= 2025 on a plain engine.
- now_is_constant_within_one_query: two NOW() in one query agree (§17.4.5.1).
- rand_is_live_across_queries: STRUUID()/RAND() vary across queries.
- now_is_live_in_update_where: BIND(NOW()) in an UPDATE WHERE inserts a current-year value.
- heldin_in_update_where_uses_configured_standpoint_predicates: heldIn in an UPDATE
  WHERE succeeds when configured, hard-fails when not.
- C-ABI now_reports_the_real_wall_clock_year retained (now exercises the default path).
- splitmix64 determinism stays unit-tested at the function level.

## Standing goals (.goals) respected
- GREENFIELD / no compromises: correct by default, everywhere, always — no opt-in knob.
- RUST-FIRST: all logic in the Rust core; the Python/C/wasm surfaces inherit it.
- SUBSUME/DRY: one primitive replaces three duplicated host clock impls.
- LOW/NO OPTIONALITY, HARD FAILS: no injection seam, no silent epoch fallback anywhere.
- wasm invariant preserved: make check + make wasm green.

- To make a query deterministic in a unit test, use EvalCtx::with_now /
  with_rng_seed (pub(crate)); there is deliberately no public determinism knob.
- Reproducibility for W3C conformance is unaffected: NOW()/RAND() are
  non-deterministic by spec and no vector pins their output.

Diff stat (origin/main..HEAD):
 Cargo.lock                       |  49 ++++++++++++
 Cargo.toml                       |   6 ++
 crates/rdf-capi/src/query.rs     |  48 ++++++++++-
 crates/sparql-eval/Cargo.toml    |  10 +++
 crates/sparql-eval/src/clock.rs  |  44 ++++++++++
 crates/sparql-eval/src/engine.rs | 169 ++++++++++++++++++++++++++++++++++++++-
 crates/sparql-eval/src/eval.rs   |  27 ++++---
 crates/sparql-eval/src/expr.rs   |   6 +-
 crates/sparql-eval/src/lib.rs    |   1 +
 crates/sparql-eval/src/update.rs |  62 ++++++++++++--
 10 files changed, 396 insertions(+), 26 deletions(-)

Issue #1, opened 2026-07-02
CI-gated wasm size budget for the purrdf-wasm build


paudley:
## Goal

Turn the current best-effort `wasm-opt -Oz` optimization into a **gated** CI size budget for the `purrdf-wasm` (`crates/rdf-wasm`) artifact: a defined `.wasm` byte ceiling that fails CI on regression.

## Current state

- `purrdf-wasm` builds on `wasm32-unknown-unknown` and ships as npm `@blackcatinformatics/purrdf`.
- CI already has a dedicated wasm job (the whole workspace must stay wasm32-clean), but artifact size is measured nowhere and nothing gates it.

## Deliverables

- [ ] Measure the optimized `.wasm` artifact size in the CI wasm job (plain byte check; optionally a twiggy breakdown for diagnostics).
- [ ] Define a budget ceiling with modest headroom over the current size; document how to raise it deliberately (a one-line constant in CI, changed only with justification).
- [ ] Fail the wasm job on overshoot; report current size vs budget in the job summary.

## Acceptance

CI fails on a size regression past the budget; the budget and current size are visible in every wasm job run.

---
*Migrated from [gmeow-ontology#1212](https://github.com/Blackcat-Informatics/gmeow-ontology/issues/1212); rewritten for the purrdf extraction.*

paudley:
# Implementation Plan for Issue #1

## Context

Issue #1 (`enhancement`, `performance`, `wasm`, `complexity-hard`) asks us to turn
the existing best-effort `wasm-opt -Oz` step into a **hard CI gate**: measure the
optimized `.wasm` artifact's byte size, compare it against a documented budget
ceiling, fail CI on regression past that ceiling, and surface `current vs budget`
in every wasm job run.

Why now: the wasm artifact ships to browsers and Node; uncontrolled size creep is
a silent portability/perf regression that no current gate catches. The project's
`.goals` demand MAXIMAL PORTABILITY ("keep wasmable code as your core value") and
MAXIMAL PERFORMANCE — a size budget is the enforcement mechanism for both.

**Key finding grounding this plan:** the CI `wasm` job (`.github/workflows/ci.yaml`
lines 133–170) does **not** currently build the optimized artifact at all. It runs
`cargo check` + a plain `cargo build --lib` and only asserts the *raw*
`target/wasm32-unknown-unknown/release/purrdf_wasm.wasm` is non-empty. The
[...]

PR #27, opened 2026-07-02
perf(sparql-eval): single-threaded optimization backlog (items 3–7 + ORDER BY triple residual)

### LINKED ISSUE #20: SPARQL eval: single-threaded optimization backlog (profiled, ranked)
## Context

A measurement pass (perf + the end-to-end `benches/query_eval.rs` bench, 303k-quad synthetic dataset, 7 query shapes) produced a ranked, evidence-grounded optimization list for `purrdf-sparql-eval`. The engine is TermId-pure through joins as designed, but the id→value boundary leaks in FILTER/ORDER/aggregates, and per-row loop-invariant work dominates several shapes. Cases with heavy allocation spend 12–22% in malloc/free.

## Ranked items (file:line from the profile)

1. **Regex cache churn** (`expr.rs:1843`): per-row `Regex.clone()` re-initializes the lazy-DFA cache pool (12.4% of the filter case alone) and the `(String,String)` cache key allocates per row. Return a shared regex + borrowed-key probe. ~1.3× on regex filters.
2. **ORDER BY key parsing** (`modifier.rs:137,330`): `term_value_order` re-runs `parse_by_iri` per *comparison* — 52.6% of the sort case (~900k parses for 30k rows). Precompute parsed sort keys once per row. ~2× on ORDER BY.
3. **Pre-interned constants** (`expr.rs:229`): every FILTER sub-result allocates + hashes + reverse-index-probes a boolean literal per row. Cache true/false `SolutionTerm`s in `EvalCtx`; memoize `Expression::Literal`.
4. **Per-query id→value/parse memo**: `compare`/`equal`/`ebv_term` (`expr.rs:283,310`) materialize owned TermValues then re-parse per row; `materialize_solutions` re-resolves duplicate ids (13–23% of OPTIONAL/UNION cases). ~1.15–1.3×.
5. **Hash-join allocs** (`binop.rs:94,199`): pack single-column join keys as `u64` (`SolutionTerm` is 8B/Copy) instead of `Vec`; pre-size build maps. ~1.15× on OPTIONAL-heavy.
6. **Loop-invariant BGP probe hoisting** (`bgp.rs:135` + rdf-core `dataset.rs:666`): per partial row the engine re-runs full-table `partition_point` pairs — **73% of the selective-join case, 53% of aggregates, 47% of paths** — the single largest engine-wide sink. Tier (i): hoist permutation choice + constant-axis bounds out of the row loop. Tier (ii): sort-merge rows against the permutation run. 1.5–2.5× on join-heavy shapes.
7. **Path closure bookkeeping** (`path.rs:481`): BTreeMap frontier/visited → id-keyed DetHashSet/bitmaps (23% of the transitive case). ~1.2×.

Full breakdown tables + the five architecture answers live in the profiling report (bench in-tree at `crates/sparql-eval/benches/query_eval.rs`; baseline numbers in its header commit).

## Gates
W3C SPARQL conformance untouched; byte-identical results incl. ORDER BY tie-breaking; DetHasher for new maps; before/after `query_eval` numbers per item.

Parallel evaluation is deliberately separate (#18) — land these first.

paudley:
Closes #20

## Summary

Residual sweep of the profiled single-threaded optimization backlog for `purrdf-sparql-eval`.
Each item was implemented behind the issue's hard gates — **W3C conformance untouched,
byte-identical results, `DetHasher` for new maps, wasm-clean, no new deps, no Cargo features** —
and measured with **deterministic, contention-independent metrics** (see below).

Two backlog items were assessed and found **already implemented on `main`** (verified, not
skipped): **Item 1** (regex compile cache — two-level `DetHashMap`, borrowed-key probe, shared
`Rc<Regex>`) and **Item 2** common ORDER BY path (XSD parse already hoisted into `sort_key`).

## Measurement methodology

Wall-clock timing is meaningless on the shared build host (100+ tenants): back-to-back runs of
*identical* code swing ±17% at criterion `sample_size=10`. This PR uses metrics that count **work
done, not time**, and are exactly reproducible:
- **Heap-allocation counts/bytes** over the fixed `query_eval` dataset (allocation-reduction items).
- **Instructions retired** (`perf stat -e instructions:u`, run-to-run variance ~0.002%) for CPU items.

## Changes (one commit each, per-item evidence in each message)

| Item | Change | Deterministic result |
|---|---|---|
| 3 | Memoize constant `NamedNode`/`Literal` atoms per query | `b_scan_filter` −59,997 allocs (−25%) |
| 4 | Memoize dataset literal XSD parses (`TermId`→`XsdValue`) | `b_scan_filter` −5.5% instructions |
| 5 | Allocation-free single-column `u64` join keys + pre-sized build map | `c_optional_heavy` −213,038 allocs (−17.4%) |
| 7 | O(1) `DetHashSet` visited-set for path transitive closure | `f_path_transitive` −3.8% instructions |
| 6a | Hoist loop-invariant BGP probe permutation selection | `a` −1.4%, `e` −2.8% instructions |
| 2* | Pre-parse quoted-triple ORDER BY sort keys (residual) | correctness cleanup; equivalence unit test |

**Item 6 tier (ii)** (adaptive sort-merge) was profiled and **not shipped**: on every benched shape
the join is either selective (probe count ≪ run length → binary-search probing is already optimal)
or output-dominated (row materialization dominates; probe ≤6%), so an adaptive planner would never
select it — shipping it would be unreachable-on-this-workload code. The measurement decides the SOTA
shape (details on #20). Tier (i) is Item 6's deliverable.

## Cumulative allocations (origin/main → this branch, exact/reproducible)

- `b_scan_filter`: 240,453 → 180,462 (−24.9%)
- `c_optional_heavy`: 1,227,187 → 1,014,149 (−17.4%)
- `f_path_transitive`: 205,014 → 200,039 (−2.4%)

## Gates
- `make check` green (fmt + clippy pedantic/nursery + `cargo test --workspace` incl. W3C SPARQL
  conformance + hygiene + `make wasm`).
- **Byte-identical** results across all 7 `query_eval` shapes, diffed against origin/main row
  fixtures (BGP bag order + ORDER BY tie-breaks) after every commit.
- All new maps/sets use the crate's fixed-key `DetHasher`; no new dependencies; no Cargo features.

### DIFF
--- a/bindings/python/tests/rdflib_suite/xfail_ledger.toml
+++ b/bindings/python/tests/rdflib_suite/xfail_ledger.toml
@@ -57,7 +57,6 @@
 # --- Theme: SPARQL aggregate / filter evaluation semantics ---
 "test_agg_distinct.py::test_count_distinct" = "COUNT(DISTINCT ?x) grouped aggregation yields an empty Result under the native engine instead of the two expected grouped counts"
 "test_nested_filters.py::test_nested_filter_outer_binding_propagation" = "FILTER bindings from an outer group are not propagated into a nested group as rdflib does, so the computed solution set diverges from the expected set"
-"test_nested_filters.py::test_nested_filter_outermost_binding_propagation" = "outermost-group FILTER bindings are not propagated into nested groups as rdflib does, so the computed solution set diverges from the expected set"
 "test_conjunctive_graph.py::test_graph_ids[check-kws1]" = "ConjunctiveGraph.parse forwards a StringInputSource down a filesystem path in this keyword combination, raising FileNotFoundError instead of parsing the in-memory source"
 
 # --- Theme: ConjunctiveGraph legacy semantics ---
--- a/crates/rdf-core/src/ir/dataset.rs
+++ b/crates/rdf-core/src/ir/dataset.rs
@@ -266,7 +266,7 @@ fn compare_prefix(
 
 /// One of the six quad orderings. SPOG is the identity (the freeze-sorted table); the
 /// other five are materialized lazily as ordinal arrays.
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, Debug)]
 enum QuadPermutation {
     Spog,
     Pos,
@@ -301,6 +301,22 @@ impl QuadPermutation {
             Self::Gosp => AxisOrder::new([G, O, S, P]),
         }
     }
+}
+
+/// A loop-invariant probe plan: the permutation and prefix length chosen purely from
+/// *which* axes a pattern binds (not their bound values) plus the graph constraint.
+///
+/// The permutation choice in [`RdfDataset::pattern_candidate_run`] depends only on the
+/// bound-axis shape, which is constant across the rows of one index-nested-loop join
+/// slot (a variable bound by an earlier BGP pattern is bound for every row; an unbound
+/// one for none). So the join computes this **once** per slot via
+/// [`RdfDataset::probe_plan`] and reuses it for every probe row through
+/// [`RdfDataset::quads_for_pattern_with_plan`], instead of re-scanning all six
+/// permutations on each row. Opaque by design — callers only pass it back.
+#[derive(Clone, Copy, Debug)]
+pub struct QuadProbePlan {
+    perm: QuadPermutation,
+    prefix: usize,
 }
 
 /// The candidate-quad source for an indexed pattern query. Unifies the two access
@@ -670,49 +686,80 @@ impl RdfDataset {
         o: Option<TermId>,
         g: GraphMatch,
     ) -> (QuadPermutation, usize, usize) {
-        // The bound key for an axis, or `None` if the pattern leaves it free.
-        let bound = |axis: Axis| -> Option<u32> {
-            match axis {
-                Axis::S => s.map(|id| id.index() as u32),
-                Axis::P => p.map(|id| id.index() as u32),
-                Axis::O => o.map(|id| id.index() as u32),
-                Axis::G => match g {
-                    GraphMatch::Any => None,
-                    GraphMatch::Default => Some(0),
-                    GraphMatch::Named(id) => Some(id.index() as u32 + 1),
-                },
-            }
-        };
+        let plan = Self::probe_plan(s.is_some(), p.is_some(), o.is_some(), g);
+        self.candidate_run(&plan, s, p, o, g)
+    }
 
-        // Choose the permutation with the longest leading run of bound axes. SPOG is
-        // first in `ALL`, so it wins ties (it needs no array). The chosen `target`
-        // holds the bound key for each of the `prefix` leading axes.
+    /// Select the [`QuadProbePlan`] — permutation + prefix length — for a pattern's
+    /// bound-axis shape (which of s/p/o are bound, plus the graph constraint). This is
+    /// the **value-independent** half of [`Self::pattern_candidate_run`]: it reads only
+    /// the boundness of each axis, so an index-nested-loop join whose probe slot has a
+    /// fixed shape across rows computes it once and reuses it (see [`QuadProbePlan`]).
+    ///
+    /// Choose the permutation whose sort prefix covers the most leading bound axes;
+    /// SPOG is first in `ALL` so it wins ties (it needs no ordinal array).
+    #[must_use]
+    pub fn probe_plan(s_bound: bool, p_bound: bool, o_bound: bool, g: GraphMatch) -> QuadProbePlan {
+        let g_bound = !matches!(g, GraphMatch::Any);
+        let axis_bound = |axis: Axis| match axis {
+            Axis::S => s_bound,
+            Axis::P => p_bound,
+            Axis::O => o_bound,
+            Axis::G => g_bound,
+        };
         let mut best = QuadPermutation::Spog;
         let mut prefix = 0usize;
-        let mut target: QuadAxisKeys = [0; QUAD_ARITY];
         for perm in QuadPermutation::ALL {
             let axes = perm.axes();
             let mut k = 0;
-            let mut t: QuadAxisKeys = [0; QUAD_ARITY];
-            while k < QUAD_ARITY {
-                match bound(axes.axes[k]) {
-                    Some(v) => {
-                        t[k] = v;
-                        k += 1;
-                    }
-                    None => break,
-                }
+            while k < QUAD_ARITY && axis_bound(axes.axes[k]) {
+                k += 1;
             }
             if k > prefix {
                 prefix = k;
                 best = perm;
-                target = t;
             }
         }
+        QuadProbePlan { perm: best, prefix }
+    }
 
-        let axes = best.axes();
+    /// The contiguous `[lo, hi)` candidate run for this row's `(s, p, o, g)` values under
+    /// a precomputed [`QuadProbePlan`] — the **value-dependent** half of
+    /// [`Self::pattern_candidate_run`]. Builds the `target` key for the plan's `prefix`
+    /// leading (therefore bound) axes from the row's values, then binary-searches the
+    /// run. For SPOG the bounds index the freeze-sorted `quads` table directly;
+    /// otherwise the permutation's ordinal array.
+    fn candidate_run(
+        &self,
+        plan: &QuadProbePlan,
+        s: Option<TermId>,
+        p: Option<TermId>,
+        o: Option<TermId>,
+        g: GraphMatch,
+    ) -> (QuadPermutation, usize, usize) {
+        let axes = plan.perm.axes();
+        // The bound key for one of the `prefix` leading axes — each is bound by
+        // construction of the plan, so the `expect`/`unreachable` cannot fire.
+        let key_of = |axis: Axis| -> u32 {
+            match axis {
+                Axis::S => s.expect("plan prefix axis S is bound").index() as u32,
+                Axis::P => p.expect("plan prefix axis P is bound").index() as u32,
+                Axis::O => o.expect("plan prefix axis O is bound").index() as u32,
+                Axis::G => match g {
+                    GraphMatch::Default => 0,
+                    GraphMatch::Named(id) => id.index() as u32 + 1,
+                    GraphMatch::Any => unreachable!("plan prefix axis G is bound"),
+                },
+            }
+        };
+        let prefix = plan.prefix;
+        let mut target: QuadAxisKeys = [0; QUAD_ARITY];
+        for (slot, &axis) in target.iter_mut().zip(axes.axes.iter()).take(prefix) {
+            *slot = key_of(axis);
+        }
+
         // Binary-search the contiguous run whose `prefix` leading keys equal `target`.
-        match best {
+        match plan.perm {
             QuadPermutation::Spog => {
                 let lo = self
                     .quads
@@ -720,17 +767,17 @@ impl RdfDataset {
                 let hi = self
                     .quads
                     .partition_point(|q| compare_prefix(axes, q, &target, prefix).is_le());
-                (best, lo, hi)
+                (plan.perm, lo, hi)
             }
             _ => {
-                let arr = self.permutation(best);
+                let arr = self.permutation(plan.perm);
                 let lo = arr.partition_point(|&ord| {
                     compare_prefix(axes, &self.quads[ord as usize], &target, prefix).is_lt()
                 });
                 let hi = arr.partition_point(|&ord| {
                     compare_prefix(axes, &self.quads[ord as usize], &target, prefix).is_le()
                 });
-                (best, lo, hi)
+                (plan.perm, lo, hi)
             }
         }
     }
@@ -774,7 +821,25 @@ impl RdfDataset {
         o: Option<TermId>,
         g: GraphMatch,
     ) -> impl Iterator<Item = QuadIds> + '_ {
-        let (best, lo, hi) = self.pattern_candidate_run(s, p, o, g);
+        let plan = Self::probe_plan(s.is_some(), p.is_some(), o.is_some(), g);
+        self.quads_for_pattern_with_plan(&plan, s, p, o, g)
+    }
+
+    /// Like [`Self::quads_for_pattern_indexed`], but with a caller-precomputed
+    /// [`QuadProbePlan`] (see [`Self::probe_plan`]) so the per-call permutation
+    /// selection — loop-invariant across an index-nested-loop join slot — is skipped.
+    /// Behaviour is otherwise identical: the same selectivity guard and the same
+    /// residual `(s, p, o, g)` id-equality + [`GraphMatch`] filter, so the yielded
+    /// quads and their order are unchanged.
+    pub fn quads_for_pattern_with_plan(
+        &self,
+        plan: &QuadProbePlan,
+        s: Option<TermId>,
+        p: Option<TermId>,
+        o: Option<TermId>,
+        g: GraphMatch,
+    ) -> impl Iterator<Item = QuadIds> + '_ {
+        let (best, lo, hi) = self.candidate_run(plan, s, p, o, g);
         let candidates = match best {
             // For SPOG the run is a sub-slice of the freeze-sorted table (sequential).
             QuadPermutation::Spog => QuadCandidates::Slice(self.quads[lo..hi].iter()),
--- a/crates/rdf-core/src/ir/mod.rs
+++ b/crates/rdf-core/src/ir/mod.rs
@@ -40,7 +40,9 @@ pub use builder::{RdfDatasetBuilder, ValidatedRdfDatasetBuilder};
 pub use bundle::{GtsBundle, RdfEnvelope};
 pub use canon::{canonicalize, canonicalize_with, CanonHash, Canonicalized};
 pub use compare::{dataset_diff, datasets_isomorphic, DatasetDiff};
-pub use dataset::{QuadHandle, QuadIds, QuadRef, RdfDataset, RdfDatasetIter, TermRef};
+pub use dataset::{
+    QuadHandle, QuadIds, QuadProbePlan, QuadRef, RdfDataset, RdfDatasetIter, TermRef,
+};
 pub use event_sink::RdfDatasetVisitor;
 pub use ingest::{DatasetSink, FrozenDatasetSource};
 pub use mutable::{MutableDataset, QuadValues};
--- a/crates/rdf-core/src/lib.rs
+++ b/crates/rdf-core/src/lib.rs
@@ -79,9 +79,9 @@ pub use fno::{
 pub use ir::{
     canonicalize, canonicalize_with, dataset_diff, datasets_isomorphic, BlankScope, CanonHash,
     Canonicalized, DatasetDiff, DatasetSink, FrozenDatasetSource, GtsBundle, HandleEntry,
-    HandleKey, MutableDataset, PipelineBundle, PipelineBundleError, QuadHandle, QuadIds, QuadRef,
-    QuadValues, RdfDataset, RdfDatasetBuilder, RdfDatasetVisitor, RdfEnvelope, TermId, TermRef,
-    TermValue, ValidatedRdfDatasetBuilder,
+    HandleKey, MutableDataset, PipelineBundle, PipelineBundleError, QuadHandle, QuadIds,
+    QuadProbePlan, QuadRef, QuadValues, RdfDataset, RdfDatasetBuilder, RdfDatasetVisitor,
+    RdfEnvelope, TermId, TermRef, TermValue, ValidatedRdfDatasetBuilder,
 };
 pub use lookaside::{
     RdfBlobOrigin, RdfBlobRecord, RdfLookaside, RdfLookasideKind, RdfLookasideResource,
--- a/crates/sparql-eval/src/bgp.rs
+++ b/crates/sparql-eval/src/bgp.rs
@@ -31,7 +31,7 @@
 //! **projected away**, so two independent BGPs that happen to reuse the label `_:b`
 //! never accidentally share a join variable.
 
-use purrdf_core::{DatasetView, GraphMatch, QuadIds, RdfDataset, TermId, TermRef};
+use purrdf_core::{DatasetView, GraphMatch, QuadIds, QuadProbePlan, RdfDataset, TermId, TermRef};
 use purrdf_sparql_algebra::{NamedNodePattern, TermPattern, TriplePattern, Variable};
 
 use crate::convert::{ground_term_pattern_to_value, named_node_to_value};
@@ -137,6 +137,11 @@ pub(crate) fn eval_bgp(
     for &i in order.iter() {
         let cp = &compiled[i];
         let mut next = Vec::new();
+        // The probe's bound-axis shape is fixed across this slot's rows (a variable is
+        // bound by an earlier pattern for every row or for none), so the permutation
+        // choice is loop-invariant: compute the `QuadProbePlan` once (on the first
+        // single-graph probe) and reuse it, instead of re-selecting per row.
+        let mut probe_plan: Option<QuadProbePlan> = None;
         for row in &rows {
             let s = query_id(&cp.s, row);
             let p = query_id(&cp.p, row);
@@ -145,7 +150,10 @@ pub(crate) fn eval_bgp(
                 // Single-graph scope (store default / a named graph): the indexed
                 // partition_point read, unchanged — no de-dup overhead.
                 GraphScope::One(gm) => {
-                    for quad in ctx.dataset.quads_for_pattern(s, p, o, *gm) {
+                    let plan = *probe_plan.get_or_insert_with(|| {
+                        RdfDataset::probe_plan(s.is_some(), p.is_some(), o.is_some(), *gm)
+                    });
+                    for quad in ctx.dataset.quads_for_pattern_with_plan(&plan, s, p, o, *gm) {
                         if let Some(extended) = bind_row(row, cp, &quad, ctx.dataset) {
                             next.push(extended);
                         }
--- a/crates/sparql-eval/src/binop.rs
+++ b/crates/sparql-eval/src/binop.rs
@@ -24,7 +24,23 @@ use crate::error::EvalError;
 use crate::eval::{eval, EvalCtx};
 use crate::scratch::SolutionTerm;
 use crate::solution::{compatible, Solution, SolutionSeq, VarSchema};
-use crate::DetHashMap;
+use crate::{DetHashMap, DetHasher};
+
+/// A hash-join key over the shared columns of two solutions.
+///
+/// The overwhelmingly common case is a **single** shared variable (a star join on
+/// `?p`), so that case is specialized to a `Copy` `u64` ([`SolutionTerm::join_key_u64`])
+/// — no per-row heap allocation on either the build or probe side. Joins on zero or
+/// ≥2 shared columns keep the general owned-vector key. Within one join the shared
+/// column count is fixed, so every key is the same variant and the two never compare
+/// across variants.
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub(crate) enum JoinKey {
+    /// Exactly one shared column: the term encoded as a collision-free `u64`.
+    Single(u64),
+    /// Zero or ≥2 shared columns: the bound terms in shared-column order.
+    Multi(Vec<SolutionTerm>),
+}
 
 /// Evaluate `left . right` (algebra `Join`) as a hash join on shared variables.
 pub(crate) fn eval_join(
@@ -94,8 +110,11 @@ fn right_to_out_map(right: &VarSchema, out: &VarSchema) -> Vec<usize> {
 pub(crate) fn build_index(
     r: &SolutionSeq,
     shared: &[(usize, usize)],
-) -> (DetHashMap<Vec<SolutionTerm>, Vec<usize>>, Vec<usize>) {
-    let mut keyed: DetHashMap<Vec<SolutionTerm>, Vec<usize>> = DetHashMap::default();
+) -> (DetHashMap<JoinKey, Vec<usize>>, Vec<usize>) {
+    // Pre-size to the build-row count: the exact upper bound on distinct keys, so a
+    // large build side is filled without incremental rehash-and-reallocate churn.
+    let mut keyed: DetHashMap<JoinKey, Vec<usize>> =
+        DetHashMap::with_capacity_and_hasher(r.rows.len(), DetHasher::default());
     let mut wild: Vec<usize> = Vec::new();
     for (idx, rrow) in r.rows.iter().enumerate() {
         match bound_key(rrow, shared, KeySide::Right) {
@@ -121,7 +140,7 @@ pub(crate) fn build_index(
 pub(crate) fn probe_has_match(
     probe: &[Option<SolutionTerm>],
     shared: &[(usize, usize)],
-    keyed: &DetHashMap<Vec<SolutionTerm>, Vec<usize>>,
+    keyed: &DetHashMap<JoinKey, Vec<usize>>,
     wild: &[usize],
     r_rows: &[Solution],
 ) -> bool {
@@ -197,21 +216,27 @@ enum KeySide {
 /// The shared-column key of `row`, or `None` if any shared column is unbound.
 ///
 /// Both sides build the key in the same `shared` order, so a left key equals a
-/// right key iff the two rows agree on every (bound) shared column.
+/// right key iff the two rows agree on every (bound) shared column. A single shared
+/// column — the common star-join shape — produces an allocation-free
+/// [`JoinKey::Single`]; zero or ≥2 columns fall back to an owned [`JoinKey::Multi`].
 fn bound_key(
     row: &[Option<SolutionTerm>],
     shared: &[(usize, usize)],
     side: KeySide,
-) -> Option<Vec<SolutionTerm>> {
+) -> Option<JoinKey> {
+    let col_of = |ia: usize, ib: usize| match side {
+        KeySide::Left => ia,
+        KeySide::Right => ib,
+    };
+    if let [(ia, ib)] = *shared {
+        // Single shared column: no heap allocation for the key.
+        return Some(JoinKey::Single(row[col_of(ia, ib)]?.join_key_u64()));
+    }
     let mut key = Vec::with_capacity(shared.len());
     for &(ia, ib) in shared {
-        let col = match side {
-            KeySide::Left => ia,
-            KeySide::Right => ib,
-        };
-        key.push(row[col]?);
+        key.push(row[col_of(ia, ib)]?);
     }
-    Some(key)
+    Some(JoinKey::Multi(key))
 }
 
 /// Merge a compatible `(left_row, right_row)` pair into one solution over the output
--- a/crates/sparql-eval/src/eval.rs
+++ b/crates/sparql-eval/src/eval.rs
@@ -100,7 +100,7 @@ pub(crate) struct ExistsInner {
     /// `(outer_ordinal, inner_ordinal)` pairs (the probe's join key).
     pub shared: Vec<(usize, usize)>,
     /// Inner rows fully bound on the shared columns, grouped by their key.
-    pub keyed: DetHashMap<Vec<SolutionTerm>, Vec<usize>>,
+    pub keyed: DetHashMap<crate::binop::JoinKey, Vec<usize>>,
     /// Inner rows with an unbound shared column (compatible with any probe value).
     pub wild: Vec<usize>,
 }
--- a/crates/sparql-eval/src/eval.rs
+++ b/crates/sparql-eval/src/eval.rs
@@ -204,6 +204,16 @@ pub struct EvalCtx<'d> {
     /// fresh intern would produce. Naturally per-query: the algebra the addresses
     /// point into outlives this context's `query()` call.
     pub(crate) const_atom_cache: DetHashMap<usize, SolutionTerm>,
+    /// Per-query memo of the parsed XSD value of a dataset literal, keyed by its
+    /// `TermId`. `FILTER`/comparison hot paths (`compare`/`equal`/`ebv_term`) parse
+    /// the same `Existing(TermId)` literal's lexical form via `parse_by_iri` on
+    /// every row; a 30k-row `?age > 40` re-parses ~60 distinct ages 30k times. The
+    /// lexical form and datatype are immutable for a fixed id, so the parse is a
+    /// pure function of the id — memoizing it (including the `None` "not an XSD
+    /// value" outcome) collapses per-row re-parsing to one parse per distinct id.
+    /// Naturally per-query. Only dataset (`Existing`) ids are cached; computed
+    /// scratch values are ephemeral and stay on the borrowed-view path.
+    pub(crate) xsd_parse_cache: DetHashMap<purrdf_core::TermId, Option<purrdf_xsd::XsdValue>>,
     /// The `SERVICE` federation source, if one is injected. `None` in
     /// the default engine path: a non-silent `SERVICE` then hard-fails. Tests and
     /// the conformance harness inject an in-memory source via [`EvalCtx::with_remote`].
@@ -264,6 +274,7 @@ impl<'d> EvalCtx<'d> {
             regex_cache: DetHashMap::default(),
             cached_bool_terms: [None, None],
             const_atom_cache: DetHashMap::default(),
+            xsd_parse_cache: DetHashMap::default(),
             remote: None,
             bgp_order_cache: None,
             constructed: Vec::new(),
--- a/crates/sparql-eval/src/eval.rs
+++ b/crates/sparql-eval/src/eval.rs
@@ -194,6 +194,16 @@ pub struct EvalCtx<'d> {
     /// the scratch interner dedups by value — so the cached term is bit-identical
     /// to what a fresh intern would return.
     pub(crate) cached_bool_terms: [Option<SolutionTerm>; 2],
+    /// Per-query memo of interned constant expression atoms (`NamedNode` /
+    /// `Literal`), keyed by the atom node's immutable AST address. A constant atom
+    /// inside a `FILTER`/`BIND` is otherwise re-`to_owned()`'d into an owned
+    /// `TermValue` and re-interned (a dataset reverse-index probe) once per row;
+    /// this collapses that to a single intern per distinct atom node. Like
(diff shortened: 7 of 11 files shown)


paudley:
# Implementation Plan for Issue #20 — SPARQL eval single-threaded optimization backlog

## Context

Issue #20 is a profiled, ranked optimization backlog for `purrdf-sparql-eval`,
grounded in the in-tree end-to-end bench `crates/sparql-eval/benches/query_eval.rs`
(30k-person / ~303k-quad synthetic dataset, 7 query shapes `a`–`g`). The engine is
TermId-pure through joins, but the id→value boundary leaks in FILTER/aggregate paths
and per-row loop-invariant work dominates several shapes (12–22% in malloc on
allocation-heavy cases; item 6 is 73% of the selective-join case, 53% of aggregates,
47% of paths — the single largest engine-wide sink).

**Important reconciliation from codebase exploration:** the issue was filed against an
*earlier* snapshot. Two items are already implemented on `main` and need **verification
only**, not reimplementation:

* **Item 1 (regex cache)** — already a two-level
  `DetHashMap<String, DetHashMap<String, Option<Rc<Regex>>>>` (`expr.rs:1905-1919`,
  field `eval.rs:189`): borrowed-`&str` probe (no per-row key alloc), shared `Rc<Regex>`
  (per-row clone is a refcount bump, not a DFA-pool reinit), failures cached as `None`.
  Regression test at `expr.rs:2353`.
* **Item 2 (ORDER BY), common path** — already hoisted: `eval_order_by`
  (`modifier.rs:127-152`) precomputes a `SortKey` (with the XSD parse) once per row via
  `sort_key` (`modifier.rs:298-316`); the comparator `compare_sort_keys`
  (`modifier.rs:322-355`) is pure and re-parses nothing. A per-comparison `parse_by_iri`
  **residual survives only** in `term_value_order`→`literal_order` (`modifier.rs:366-422`),
  reached exclusively for the rare **quoted-triple** sort-key arm — addressed in Task 6.

Everything else (items 3, 4, 5, 6, 7) is genuine open/partial work and is implemented here.

* Gap analysis is adversarial, not sycophantic — assume code is substandard until proven otherwise.
* **No deferrals / followups / "left for future"** — owned work is done NOW.
* **No GitHub issue/PR numbers or process-flow references in repo source or docs** — no `// Per #20…` comments. Process flow lives in GitHub, not the tree.
* When you see a bug, you own it — fix it now, don't trace blame.

## Standing Constraints (inviolable, from `.goals` + AGENTS.md)

* **NO COMPROMISES / SOTA / MAXIMAL PERFORMANCE / SUBSUME-EXTEND-ENHANCE.** Optimizations
  are implemented fully and kept — never shipped as optional/deferrable components. Where a
  strategy is input-dependent (item 6 tier ii), the SOTA answer is an **adaptive per-slot
  planner choice** (consistent with the existing Selinger DP), not a fixed compromise and
  not a runtime feature flag.
* **LOW/NO OPTIONALITY, HARD FAILS.** No new Cargo features, no optional deps, no
  cfg-gated behavior. (`scripts/check-no-features.py` gates CI.)
* **MAXIMAL PORTABILITY / wasm-clean.** Every touched crate must stay
  `wasm32-unknown-unknown`-clean (`make wasm`). **No new dependency** — item 7's "bitmaps"
  use `DetHashSet<TermId>` (dep-free), not `roaring`/`fixedbitset` (neither exists in the
  workspace and adding one is out of scope).
* **Byte determinism is a hard gate** (issue + AGENTS.md §"Byte determinism"): results
  byte-identical incl. ORDER BY tie-breaking; W3C SPARQL conformance untouched; new maps use
  the crate's fixed-key `DetHasher` (`lib.rs:95-101`), never SipHash/RandomState.
* **Measure first** (AGENTS.md §4): each item carries before/after `query_eval` numbers;
  the bench is extended, never a bare assertion of speedup.
* Clippy pedantic+nursery warning-free; SPDX headers on any new file; deps only via root
  `[workspace.dependencies]`.

## Gate command per task

```bash
# correctness + hygiene gate (mandatory before every commit)
make check
# per-item perf evidence (report-only; capture before AND after)
cargo bench -p purrdf-sparql-eval --bench query_eval
```

The relevant W3C conformance harness (`crates/sparql-conformance`) runs inside
`make check` via `cargo test --workspace`; a green `make check` = conformance untouched.

---

**Complete before any implementation work.**

1. Resolve the top-level repo and derive paths:
   ```bash
   BRANCH="paudley/20-sparql-eval-optimization-backlog"
   WORKTREE="$REPO_ROOT/.worktrees/20-sparql-eval-optimization-backlog"
   ```
   ```bash
   ```
3. Create the worktree on a fresh branch off up-to-date main and enter it:
   ```bash
   cd "$WORKTREE"
   git branch --show-current   # MUST show the feature branch, NOT main
   ```
4. Re-establish the baseline in the fresh worktree: `make check` MUST be green before
   implementing.
5. **Capture the baseline bench numbers** (they live only in git history, not in the bench
   header — confirmed by exploration). Save criterion output for all 7 cases to compare
   against per-task:
   ```bash
   cargo bench -p purrdf-sparql-eval --bench query_eval 2>&1 | tee /tmp/issue20-baseline-timing.txt
   ```
   Treat only deltas outside criterion's reported noise band (its `change:` estimate /
   confidence interval) as real — a single sample can manufacture a false regression that
   would wrongly block a commit under Execution Rule 6. Re-run a case that lands in the noise.
6. **Capture the byte-identity result baseline (REQUIRED — this is the diff target the
   determinism gate in Tasks 4/5/7 depends on; criterion throws results away and cannot
   serve as it).** Add a tiny throwaway helper (a `#[test]` or `examples/` bin in the
   worktree, NOT committed) that runs each of the 7 bench queries `a`–`g` through
   `NativeSparqlEngine::query` on the *same* `people_dataset()` and writes a **canonical,
   stable serialization of the full result rows in returned order** (e.g. TSV of bound terms
   per row, rows in engine-emitted order) to per-case files:
   ```bash
   # produces /tmp/issue20-rows-baseline/{a..g}.tsv on origin/main HEAD
   ```
   These fixtures pin BGP bag-order and ORDER BY tie-break order for the exact synthetic
   queries — something `make check`/W3C conformance does not do. Every later "diff against
   baseline" compares regenerated `/tmp/issue20-rows-branch/{case}.tsv` against these.
7. Post this plan to the issue:
   ```bash
   gh issue comment 20 --body-file /home/paudley/.claude/plans/20-linked-whale.md
   ```

**⚠️ DO NOT proceed to Task 1 until the worktree exists, you are on the feature branch, `make check` is green, and BOTH baselines (timing + result-row fixtures for `a`–`g`) are captured.**

---

## Task 1: Item 3 — Memoize constant expression atoms

**Problem.** `eval_expr` atom arms (`expr.rs:58-59`) re-do, for **every row**,
`n.as_str().to_owned()` / `literal_to_value(l)` (allocating owned `String`s) plus a full
`intern` (`scratch.rs:92`: dataset reverse-index probe + scratch dedup) for the same
constant `Expression::NamedNode`/`Expression::Literal`. `bool_term` already caches
true/false (`expr.rs:235-243`); this extends that pattern to all literal/IRI constants.

**Approach.**
* Add a per-query constant-atom memo on `EvalCtx` (`eval.rs:135-218`, init in
  `EvalCtx::new` `eval.rs:242-259`), keyed by the `Expression` node identity — mirror the
  existing pointer-identity memo idiom (`exists_expr_vars_cache` keyed by `usize` AST
  pointer `eval.rs:181`; `path.rs` AST-pointer caches). Store the interned `SolutionTerm`
  (8-byte `Copy`) so a hit is a `Copy`, no alloc, no probe.
* The atom arms look up the memo before interning; on miss, intern once and record.
* Determinism/correctness: `SolutionTerm` interning is value-deterministic (promotes to
  `Existing(TermId)` when present, else stable `ScratchId`); memoizing it changes nothing
  observable. Keep the memo `pub(crate)` `DetHashMap` (fixed-key), init empty per query.
* Keep it minimal: memoize `NamedNode` + `Literal` (the hot constant atoms). Do NOT memoize
  variable lookups.

**Bench.** `b_scan_filter` (constant literals in FILTER) is the primary witness; also check
no regression across `a`–`g`. Record before/after.

### After Task 1 Implementation:
- Verify `make check` green + `query_eval` before/after captured.
- Commit: github-commit-agent sub-agent
- Push: github-push-agent sub-agent
- wiring-verifier sub-agent to confirm the memo is actually consulted on the atom paths.
- Post commit + bench delta to issue #20.

---

## Task 2: Item 4 — Per-id parse memo (`TermId → XsdValue`)

**Problem.** `xsd_of_term` (`expr.rs:305-319`) already avoids owned-`TermValue`
materialization (borrowed view), but still calls `parse_by_iri(lexical, iri)` **per row**
for the same `Existing(TermId)` in `compare`/`equal`/`ebv_term` (`expr.rs:281-408`). On
OPTIONAL/UNION shapes the same ids recur across many rows.

**Approach.**
* Add a per-query memo on `EvalCtx`: `DetHashMap<TermId, Option<XsdValue>>` mapping a
  dataset `Existing(TermId)` to its parsed XSD value (or cached `None` for non-XSD /
  parse-fail). Mirror the local `datatype_memo` pattern already used in
  `materialize_solutions` (`eval.rs:488,505,518`), but lift it to ctx so it spans the whole
  query, not one materialization call.
* `xsd_of_term` consults the memo for the `Existing` arm before `parse_by_iri`; the
  `Computed(sid)` arm stays as-is (scratch-local, cheap, and ids are ephemeral).
* Correctness: parsing is a pure function of `(lexical, datatype)`, both immutable for a
  fixed `TermId`, so memoization is behavior-preserving. `XsdValue` must be `Clone` (it is,
  used by value elsewhere); store owned, hand out clones — still far cheaper than re-parsing.
* Guard memory: keys are only ids that actually reach a comparison/EBV, bounded by the
  query's active term set. On a wide unselective scan that set can be large; holding one
  owned `XsdValue` per such id for the query lifetime is the accepted trade (re-parse cost
  dominates). If profiling ever shows this memo as a memory hotspot, that is a separate
  measured concern — but it is not deferred work here, just a noted bound.

**Bench.** `c_optional_heavy`, `d_union_4` (id reuse across rows); confirm `b_scan_filter`
(numeric `>` in FILTER) also benefits. Before/after recorded.

### After Task 2 Implementation:
- `make check` green + bench captured.
- Commit → github-commit-agent; Push → github-push-agent; wiring-verifier.
- Post commit + bench delta to issue #20.

---

## Task 3: Item 5 — Single-column `u64` join keys + pre-sized build map

**Problem.** `bound_key` (`binop.rs:201-215`) always allocates a `Vec<SolutionTerm>` — even
for a single shared column — on **both** build and probe sides, one heap alloc per row.
`build_index` (`binop.rs:94-107`) constructs `DetHashMap::default()` with no capacity hint
(rehash churn). `SolutionTerm` is 8-byte `Copy`, so a 1-column key fits a `u64` with zero
allocation.

**Approach.**
* Introduce an enum join-key representation that specializes the common single-column case:
  `Single(u64)` (bit-encode the 8-byte `Copy` `SolutionTerm` — the two-variant tag +
  4-byte id fits `u64`) vs `Multi(Vec<SolutionTerm>)` for `shared.len() > 1`. Keep the same
  `DetHashMap` value type (`Vec<usize>` bucket).
  * Encoding: derive a total, collision-free `u64` from `SolutionTerm` (e.g. a `#[repr]`
    transmute-free `match` mapping `Existing(id)`→`id.index() as u64` and
    `Computed(sid)`→`(1<<32) | sid as u64`, keeping variants disjoint). Must round-trip
    for equality/hash — do NOT need to decode, only compare, so equal keys ⇒ equal `u64`.
  * **Correctness invariant (mandatory `debug_assert`):** disjointness holds ONLY if both
    `id.index()` and `sid` are `< 2^32`. `TermId` is a 4-byte `NonZeroU32` (safe), and
    `ScratchId(u32)` is 4-byte — but a collision here is a *wrong join result*, not a perf
    regression, so assert `id.index() < (1<<32)` and `sid < (1<<32)` (or, better, encode via
    the width-typed ids so the invariant is enforced by construction). Add a unit test that a
    max-index `Existing` and a max-`ScratchId` `Computed` encode to distinct `u64`s.
* Both `build_index` and the probe loop in `hash_join` (`binop.rs:140-188`) use
  `bound_key`; update the single shared code path. `probe_has_match` (`binop.rs:121-138`)
  and the cached `ExistsInner.keyed` (`eval.rs:103`) share the map type — update them
  consistently so EXISTS keeps working (benched by `benches/exists_decorrelation.rs`).
* Pre-size the build map: `DetHashMap::with_capacity_and_hasher(r.rows.len(), DetHasher::default())`
  (right/build side row count is the exact upper bound on distinct keys).
* Determinism: `DetHasher` fixed-key; bucket iteration order already deterministic; output
  row order unchanged (same probe iteration, same emit sequence). Byte-identity preserved.

**Bench.** `c_optional_heavy` (OPTIONAL-heavy, single-column joins) primary; `a_selective_join`
secondary. Before/after recorded.

### After Task 3 Implementation:
- `make check` green (incl. EXISTS decorrelation tests) + bench captured.
- Commit → github-commit-agent; Push → github-push-agent; wiring-verifier.
- Post commit + bench delta to issue #20.

---

## Task 4: Item 7 — Path closure: `BTreeSet` → `DetHashSet<TermId>`

**Problem.** `closure` / `closure_multi` (`path.rs:473-527`) use `BTreeSet<TermId>` for
`visited` and `result` (O(log n) membership on the hot transitive-closure loop) — 23% of
the transitive case. `range_reach` (`path.rs:534+`) uses per-level `BTreeSet` too.

**Approach.**
* Switch `visited` to `DetHashSet<TermId>` (O(1) insert/contains; `TermId` has a
  hand-tuned dense-index `Hash` at `term.rs:57-64`). Frontier stays a `Vec<TermId>` stack.
* **Determinism guard (critical):** the closure currently returns an *ordered* `BTreeSet`.
  Downstream consumers may rely on that order for byte-identical output. Options, in order
  of preference:
  1. Keep the **return type** an ordered set (build `result` as `BTreeSet`, or collect the
     `DetHashSet` into a sorted `Vec`/`BTreeSet` at the boundary) so egress order is
     unchanged — accelerate only the internal `visited` membership checks.
  2. If profiling shows the `result` insert itself is hot, use `DetHashSet` internally and
     sort once at the return boundary.
  Verify with the conformance suite + `g`/`f` output equality that ORDER-sensitive results
  are byte-identical. Do NOT let hash iteration order leak into results.
* Consider a dense `Vec`-backed bitset keyed by `TermId::index()` ONLY if the bench shows
  `DetHashSet` insufficient AND the closure touches a dense id range — but default to
  `DetHashSet` (no new dep, no large per-closure allocation for sparse closures).
* **`range_reach` (`path.rs:534+`) is in scope too** — it uses per-level `BTreeSet<TermId>`.
  Its `current`/level sets feed the unbounded-tail `closure_multi` and ultimately egress
  ordering. Apply the SAME rule: accelerate internal membership with `DetHashSet` but keep
  the value that reaches egress an ordered set (sort at the boundary). Do NOT let a
  per-level `DetHashSet`'s iteration order leak into emitted results. If a level set is used
  only for membership (never iterated into output), it may be `DetHashSet` outright — state
  which for each site touched.

**Bench.** `f_path_transitive` primary. Before/after recorded; **diff regenerated
`/tmp/issue20-rows-branch/f.tsv` against `/tmp/issue20-rows-baseline/f.tsv`** (Task 0 step 6)
— row-set AND order must be byte-identical, including any path result consumed under ORDER BY.

### After Task 4 Implementation:
- `make check` green (path + conformance tests) + bench captured.
- Commit → github-commit-agent; Push → github-push-agent; wiring-verifier.
- Post commit + bench delta to issue #20.

---

## Task 5: Item 6 — BGP probe hoisting (biggest sink) — tier (i) + adaptive tier (ii)

**This task ships as TWO independent commits — 5a then 5b — never batched** (Execution
Rule 2). 5a is a provably order-preserving pure win; 5b is the highest-risk change in the
plan (reorders probe rows). Separating them keeps 5b independently benched, diffed, and
bisectable.

**Problem.** The per-row probe loop in `eval_bgp` (`bgp.rs:135-190`) calls, for every
partial row and every pattern in the plan, `quads_for_pattern` →
`quads_for_pattern_indexed` (`dataset.rs:770-810`) → `pattern_candidate_run`
(`dataset.rs:666-736`), which **rescans all 6 `QuadPermutation::ALL` entries** to pick the
permutation and **recomputes the `partition_point` bisection bounds** from scratch. For a
fixed plan slot the *set of bound axes* is loop-invariant (a slot bound by an earlier
pattern is `Some` for every row; an unbound slot is `None` for every row) — only the bound
axes' **values** change per row. So the permutation choice and the constant-axis structure
are pure loop-invariant waste.

**Commit 5a — Tier (i): loop-invariant hoist (pure win, identical emission order).**
* Add a dataset-side "probe plan" API: given the boundness mask (which of s/p/o/g are
  bound) — NOT the values — compute once the chosen `QuadPermutation`, its `axes()`, and
  the leading `prefix` length. Return a small `Copy` descriptor.
* In `eval_bgp`, compute this descriptor **once per plan slot** (outside the `for row`
  loop), from `CompiledPattern` + which slots are already bound in the working schema at
  that plan position (statically derivable from `order` + `Pos`).
* Per row, call a new `quads_for_pattern_with_plan(descriptor, s, p, o, g)` that skips the
  6-permutation scan and goes straight to building the `target` key from the row's values
  and running the two `partition_point` calls. Preserve the existing 1/4-table selectivity
  guard (`dataset.rs:789`) and the residual exact-match filter verbatim.
* Emission order is provably unchanged (same candidates, same order) ⇒ byte-identical.

*Commit 5a gate:* `make check` green; before/after `query_eval` on `a`/`e`/`f`; diff
regenerated `{a,e,f}.tsv` vs baseline fixtures — byte-identical. Commit → push →
wiring-verifier → post delta to #20, **before starting 5b**.

**Commit 5b — Tier (ii): adaptive sort-merge (SOTA; applied by cost, byte-identity preserved).**
Per `.goals` (NO COMPROMISES / MAXIMAL PERFORMANCE / SUBSUME-EXTEND-ENHANCE) sort-merge is
implemented as a first-class strategy, not a keep-if-convenient bolt-on. But measure-first
(AGENTS.md §4) governs *whether the branch is reachable in practice*, not just a threshold:
* When a plan slot has many probe rows whose keys fall in the same permutation, sort the
  probe rows by their join key so successive bisections walk the permutation's ordinal run
  monotonically (sort-merge instead of repeated independent bisections).
* **Byte-identity is non-negotiable and NOT a compromise against `.goals`** — it is a
  correctness invariant. Preserve it by carrying each probe row's **original index** through
  the sort and restoring the original emission order before `rows = next`. Downstream sees
  the identical row sequence.
* **Adaptive, not optional:** choose sort-merge vs index-nested-loop **per slot** by a cost
  comparison (probe-row count vs permutation-run length), consistent with the existing
  cost-based `plan_or_cached_order` Selinger DP (`bgp.rs:207`). This is runtime query
  planning (already how the engine works), not a banned Cargo feature / cfg switch. The
  capability always exists; the planner picks the cheaper strategy for the shape at hand.
* Calibrate the crossover with the bench; document the threshold rationale in a code comment
  (no issue/PR numbers per `.baseline`).
* **Ship criterion (resolves the measure-first tension honestly):** the adaptive branch is
  retained **iff there exists a benched shape (among `a`–`g` or a targeted large-fan-in
  join micro-shape added to the bench) where the planner actually selects sort-merge AND it
  wins outside criterion's noise band.** If, after calibration, no shape on this workload
  ever selects-and-wins, shipping the branch would add unreachable-in-practice code — which
  violates BOTH "measure first" and LOW/NO OPTIONALITY. In that case tier (i) stands as the
  Item 6 deliverable and 5b is not committed. This is not a deferral (the work is done and
  measured NOW, per `.baseline`); it is the measurement deciding the SOTA shape. If the
  existing 7 shapes under-exercise sort-merge, ADD a representative shape to `query_eval` so
  the decision is evidence-based, not assumed.

**Bench.** `a_selective_join` (73%), `e_group_aggregate` (53%), `f_path_transitive` (47%)
— all three must improve or hold. Capture before/after for 5a alone, then again after 5b.

*Commit 5b gate:* if retained — `make check` green; before/after `query_eval`; diff
`{a..g}.tsv` vs baseline fixtures byte-identical (this is the change that can reorder rows,
so the diff is the critical gate); the calibration evidence (which shape selects-and-wins)
posted to #20. If not retained, record the measurement conclusion on #20 and skip the commit.

---

## Task 6: Item 2 residual — hoist `parse_by_iri` out of quoted-triple ORDER BY comparator

**Problem.** The common ORDER BY path is already optimal (see Context). The lone residual:
`term_value_order` → `literal_order` (`modifier.rs:366-422`) calls `parse_by_iri` **twice
per comparison**, reached only for the `SortKey::Triple` arm (`compare_sort_keys`
line 352) — quoted-triple term ordering.

**Approach.**
* Precompute the parsed comparison keys for triple components once at `sort_key` build time
  (`modifier.rs:298-316`), mirroring how the top-level literal arm already stores its parsed
  `xsd` in `SortKey::Literal`. Extend `SortKey::Triple(TermValue)` to carry pre-parsed
  component keys (or a parsed representation) so `compare_sort_keys` performs no `parse_by_iri`.
* Determinism: ORDER BY tie-breaking must remain byte-identical — the parse result and the
  fallback `(datatype, language, lexical)` tuple compare are unchanged; only *when* the
  parse happens moves (build-time, once/row) — identical ordering.

**Bench.** `g_order_by_limit` (add or reuse a quoted-triple ORDER BY micro-shape if the
existing `g` doesn't exercise the `Triple` arm; if `g` sorts only literals, note that this
residual is not on `g`'s hot path and validate via a conformance ordering test instead).
Before/after recorded.

### After Task 6 Implementation:
- `make check` green (ORDER BY conformance) + bench/evidence captured.
- Commit → github-commit-agent; Push → github-push-agent; wiring-verifier.
- Post commit + evidence to issue #20.

---

## Task 7: Create Pull Request (MANDATORY FINAL STEP)

1. Verify all changes committed and pushed; `make check` green on the branch tip.
2. Confirm items 1 & 2 (common path) are verified-as-already-landed in the PR body (with
   the `expr.rs:1905` / `modifier.rs:127` references) so reviewers know they were assessed,
   not skipped.
3. Create PR:
   ```bash
   gh pr create --title "perf(sparql-eval): single-threaded optimization backlog (items 3–7 + ORDER BY triple residual)" --body "Closes #20

   ## Summary
   Residual sweep of the profiled optimization backlog for purrdf-sparql-eval.
   Items 1 (regex cache) and 2 (common ORDER BY path) were already implemented on
   main and are verified here; items 3, 4, 5, 6, 7 and the item-2 quoted-triple
   residual are implemented, each with before/after query_eval numbers.

   ## Changes
   - Item 3: per-query constant-atom intern memo on EvalCtx
   - Item 4: per-query TermId→XsdValue parse memo
   - Item 5: single-column u64 join keys + pre-sized build maps
   - Item 6: BGP probe hoisting — tier (i) loop-invariant permutation/bounds hoist
     + adaptive sort-merge tier (ii), byte-identity preserved
   - Item 7: path closure visited set BTreeSet→DetHashSet
   - Item 2 residual: hoist parse_by_iri out of quoted-triple ORDER BY comparator

   ## Gates
   - W3C SPARQL conformance untouched (make check green)
   - Byte-identical results incl. ORDER BY tie-breaking (verified per case)
   - DetHasher for all new maps; wasm-clean; no new deps; no Cargo features
   - Per-item before/after query_eval numbers in commit messages
   "
   ```
4. Post plan to PR:
   ```bash
   gh pr comment "$PR_NUM" --body-file /home/paudley/.claude/plans/20-linked-whale.md
   ```
5. Report: "Stage 1 complete. PR #<num> created for issue #20. Ready for /stage2 20".

---

## Verification (end-to-end)

* **Correctness/conformance:** `make check` (fmt + clippy pedantic/nursery + build +
  `cargo test --workspace` incl. `crates/sparql-conformance` + hygiene + `make wasm`) green
  after every task. This is the W3C-conformance-untouched gate.
* **Byte-identity:** diff the per-task regenerated `/tmp/issue20-rows-branch/{a..g}.tsv`
  against the `origin/main` fixtures `/tmp/issue20-rows-baseline/{a..g}.tsv` captured in
  Task 0 step 6 — must be byte-identical incl. row order (BGP bag order and ORDER BY
  tie-breaks). Any diff blocks the commit. This artifact — not the criterion bench, which
  discards results — is the determinism gate for items 6 (tier ii) and 7.
* **Performance:** `cargo bench -p purrdf-sparql-eval --bench query_eval` before (Task 0,
  `/tmp/issue20-baseline.txt`) and after each task; per-item deltas posted to issue #20 and
  carried in commit messages. Expected directional wins: `b` (items 3,4), `c` (items 4,5),
  `f` (item 7), `a`/`e`/`f` (item 6, largest).
* **No new deps / features / nightly:** `scripts/check-no-features.py` (in `make check`) and
  absence of any root-`Cargo.toml` dependency additions.

## Execution Rules

1. Task 0 (worktree setup + baseline bench) is MANDATORY first; never work on `main`.
2. Commit after EACH task (never batch) via github-commit-agent; push via github-push-agent.
3. Each task carries its own before/after `query_eval` evidence in the commit message.
4. Task 7 (PR creation) is MANDATORY last.
5. Use sub-agents for all commits/pushes; run implementation via general-purpose or
   domain-specific (performance-engineer) sub-agents.
6. If any task's bench shows a regression or a byte-identity diff, fix within that task —
   do not defer (`.baseline`).

paudley:
# Stage 2 — Gap Analysis & Remediation: Issue #20 / PR #27

**Branch:** `paudley/20-sparql-eval-optimization-backlog`
**Worktree:** `.worktrees/20-sparql-eval-optimization-backlog`
**PR:** #27 · **CI:** `pytest` **RED**, `conformance` pending, all else green.

## Context

Issue #20 is a profiled single-threaded optimization backlog for `purrdf-sparql-eval`
(items 1–7). The branch ships six perf commits. Independent verification found the branch
introduces **one CRITICAL correctness defect** whose blast radius includes silent query
corruption and a red conformance gate — plus two latent instances of the same bug class and
minor coverage/measurement gaps. The other five optimizations are correct, deterministic,
`DetHasher`-backed, wasm-clean, dependency-free, and feature-free.

The failing `pytest` check and the gemini-code-assist CRITICAL review are the **same root
cause**: `const_atom` (new, commit `4336a66`, Item 3) caches interned constants by the **heap
address** of an `Expression` node. Inside correlated `EXISTS`, `substitute_pattern`
(`expr.rs:722`) heap-allocates a temporary algebra tree per outer row; `eval(&substituted)`
runs `const_atom` over those temporaries; the tree is dropped each iteration; the allocator
recycles addresses → stale `addr→value` cache hits serve the **wrong constant** to later rows.
The strict-xfail conformance entry `test_nested_filter_outermost_binding_propagation`
(a doubly-nested correlated `FILTER NOT EXISTS`) flipped to XPASS because the corruption
coincidentally matched rdflib's answer for that fixture — an allocator-dependent, silently
wrong result, **not** a fix.

## Ranked Gaps

| # | Prio | Location | Defect |
|---|------|----------|--------|
| 1 | **CRITICAL** | `expr.rs:231-243` (`const_atom`) | Address-keyed cache over per-row substituted-EXISTS temporaries → stale constant → silently wrong solution set (allocator-dependent). Root cause of the red `pytest` gate + gemini's review. |
| 2 | **HIGH** | `expr.rs:696` (`exists_expr_vars_cache`), `expr.rs:732` (`exists_inner_cache`) | Same address-key-over-temporary hazard when a nested EXISTS is evaluated inside a substituted pattern (the failing query reaches both). Pre-existing on `main`, but owned per `.baseline` ("when you see it, you own it"). |
| 3 | MEDIUM | `crates/sparql-eval` tests | No correlated-EXISTS constant-substitution shape in the branch's own determinism gate; the 7-shape `query_eval` byte-identity corpus structurally could not catch gap #1. |
| 4 | MEDIUM | Item 6a measurement | Missing the `f_path_transitive` bench witness (issue named `a`/`e`/`f`; only `a`/`e` posted). |
| 5 | LOW | `eval.rs` `const_atom_cache` field doc | States a soundness invariant ("the algebra … outlives this context's `query()` call") that is **false** on the correlated path — a rationalized-correctness comment (ETHOS §21). |

Not gaps (verified sound): Items 4, 5, 7, 6a-mechanics, 2-residual are correct/deterministic/
`DetHasher`/wasm/dep/feature clean; Item 6b non-ship (adaptive sort-merge) is a justified
measurement-driven decision; no `.baseline` process-flow refs added by the diff.

## Remediation

Per stage2: one Task-tool sub-agent per fix → commit (github-commit-agent, iterate on
PRECOMMIT_GUIDANCE.md until green) → push (github-push-agent) → post summary to issue **and**
PR → wiring-verifier. Fixes ordered below.

### Fix A — ABA correctness guard (closes gaps 1, 2, 5) — one atomic commit

Gaps 1, 2, 5 are a **single coherent correctness fix**; splitting them would leave intermediate
broken/inconsistent states, so they land together (this is not "batching" unrelated work).

1. Add `in_substituted_exists: bool` to `EvalCtx` (init `false`).
2. In `exists()` correlated path, wrap the temporary eval with **save/restore** (handles the
   doubly-nested case correctly):
   ```rust
   let prev = ctx.in_substituted_exists;
   ctx.in_substituted_exists = true;
   let inner = eval(&substituted, ctx);
   ctx.in_substituted_exists = prev;
   let inner = inner?;
   ```
3. When `ctx.in_substituted_exists` is true, **bypass all three address-keyed caches** — they
   key on nodes that may be per-row temporaries:
   - `const_atom` (`expr.rs:236-241`): skip the read + write, return `intern(ctx, build())`.
   - `exists_expr_vars_cache` (`expr.rs:696-705`): recompute `pattern_expr_vars` without
     caching.
   - `exists_inner_cache` (`expr.rs:732-763`): skip the get/insert.
   Bypassing memo is always correctness-safe; the cost lands only on the already-expensive
   per-row-substitute correlated path. The common path (top-level algebra, uncorrelated EXISTS)
   keeps full memoization — original-algebra constants are truly invariant and stay fast.
4. Correct the `const_atom_cache` field doc: the outlives-query invariant holds only for the
   static algebra; substituted-EXISTS temporaries are the exception, which is why the guard
   exists. **Do not weaken it to a rationalization** — state the real invariant.

**Do NOT edit `xfail_ledger.toml`.** The XPASS is a true positive; the ledger entry must stay,
and the fix must make the query re-diverge (honor the xfail) by restoring pre-branch semantics.

Note: keying by a stable compile-time node id is **not** a valid alternative — substituted
constants legitimately differ per outer row at the same algebra position, so a position key
would memoize the wrong value deterministically. Bypass-in-substituted-eval is the correct SOTA
shape.

Files: `crates/sparql-eval/src/expr.rs`, `crates/sparql-eval/src/eval.rs`.

### Fix B — correlated-EXISTS regression test (closes gap 3) — separate commit

Add a Rust unit test in `expr.rs`'s test module (near the `exists_memo_*` tests, ~line 3080)
that builds a **correlated nested `FILTER NOT EXISTS`** with per-outer-row constant
substitution across multiple outer rows and asserts the exact solution set. This is the
greenfield-native regression guard (runs in `make check`, no Python). The in-tree rdflib
conformance test remains the end-to-end guard.

`crates/sparql-eval/src/expr.rs` (tests).

### Fix C — Item 6a `f_path_transitive` witness (closes gap 4) — measurement only

No code change: run the existing `query_eval` bench for the `f` shape on this branch vs.
`origin/main`, record the before/after (instructions-retired / alloc count, per the branch's
deterministic methodology), and post the number to the issue + PR to complete the "bench
witness per item" gate. If a number can't be produced, state why (as with the 2-residual item).

## Verification (end-to-end)

Run from inside the worktree after Fix A + Fix B:

1. **Rust gate:** `cargo test -p purrdf-sparql-eval` (new test + W3C SPARQL conformance) and
   `make check` (fmt + clippy pedantic/nursery + build + tests + hygiene + `make wasm`).
2. **Python conformance (the gate that caught it):**
   ```
   cd bindings/python
   uv run --group acceptance maturin develop
   uv run --group acceptance pytest tests -q
   ```
   Confirm `failed=0` and that `test_nested_filter_outermost_binding_propagation` is back to
   **xfail** (scoreboard `xpassed=0`), with the ledger entry unchanged.
3. **CI:** after push, confirm `pytest` and `conformance` both green on PR #27.
4. **Regression proof:** the new Rust test asserts the correct set for correlated EXISTS; the
   conformance test re-diverges as documented — together proving the corruption is gone and
   original-algebra memoization still works.

## Done criteria

All five gaps addressed; `pytest` + `conformance` green; `make check` green; summaries posted
to issue #20 and PR #27; wiring-verifier clean. Ready for `/stage3`.

paudley:
## Stage 2 gap remediation — root cause found and fixed (commit `3116390`)

The Stage-1 gap analysis proposed the right *fix* but mis-attributed the *symptom*. Empirical bisection + engine-level probing corrected the picture. Reporting the full story because it reverses the earlier "do not touch the ledger" call.

### What was actually wrong

The red `pytest` gate was a strict-xfail conformance case
(`test_nested_filters.py::test_nested_filter_outermost_binding_propagation` — a
doubly-nested `FILTER NOT EXISTS` passing an outermost binding into the innermost
`NOT EXISTS`). It flipped to XPASS. Investigation:

1. **The gate is not caused by the const-atom memo (Item 3).** Applying the guard alone didn't move it; bisecting the six commits pinned the flip to **Item 6a (`7be40d9`, BGP probe-plan hoist)**.
2. **But 6a is value-equivalent by inspection** (identical permutation/prefix selection, identical residual filter). It never *fixed* anything — it only perturbed allocation/evaluation order enough to *hide* the real bug for this one query. That XPASS was fragile masking.
3. **The real bug pre-dates the entire PR.** Direct engine probing: on `main` the query returns the **wrong (empty) set** on first evaluation and "warms" to correct — a deterministic-but-allocation-order-dependent corruption. `main` has no const-atom memo, so the pre-existing `exists_expr_vars_cache` / `exists_inner_cache` alone are sufficient to cause it.

### Root cause (gemini-code-assist was right about the mechanism)

Correlated EXISTS substitutes each outer row's bindings into a fresh
`substitute_pattern` **heap temporary** and evaluates it. Three per-query caches key
those temporary nodes by **raw address** — `const_atom_cache`,
`exists_expr_vars_cache`, `exists_inner_cache`. The temporary is dropped each row;
the allocator recycles its address; a later row hits a **stale entry** and reads the
prior row's constant / correlation decision / inner index → silently wrong solution
set. This is exactly the ABA hazard gemini flagged at `expr.rs`, generalized to the
two pre-existing `exists_*` caches as well.

**Proof it's the root cause:** the fix applied to the pre-6a tree (`75ccb13`, *without*
6a) makes the query deterministically correct (N=1) in every evaluation pattern —
first-run, repeated, fresh-graph, and in the full suite. It is correct **with or
without 6a**.

### The fix

`EvalCtx::in_substituted_exists` — set (save/restore, for doubly-nested correlated
EXISTS) around the substituted-pattern eval, and **bypass all three address-keyed
caches while set**. Address keying stays sound for the static query algebra (which
outlives the query); only the per-row temporaries are excluded, so the common path
keeps full memoization and zero perf regression.

Per the "no XPASS/XFAIL lies" principle: the vendored test now computes the correct
set, so its **stale xfail-ledger entry is removed** and it passes as an ordinary test.
A native `sparql-eval` regression (correlated EXISTS over multiple outer rows,
memo-invariant) guards the fix.

Item 6a stays — it is an independently valid perf win (byte-identical on the 7
`query_eval` shapes); it simply was never the correctness fix. (Item 6's path-case
coverage is the not-shipped tier-ii; 6a's own message already notes property paths
use `for_each_quad`, so `f_path_transitive` is unaffected by it.)

### Verification (local, all green)
- rdflib conformance: **62 passed / 24 xfailed, xpassed=0, failed=0, errors=0, ledger 24/24 applied, stale=0** (was `failed=1`).
- full python `pytest tests`: **325 passed, 1 skipped, 7 xfailed** (was `1 failed`).
- `cargo test -p purrdf-sparql-eval`: **262 passed, 0 failed** (incl. new regression).
- `make check`: green — fmt + clippy pedantic/nursery + build + workspace tests + rdf-core hygiene + `make wasm`.

Ready for /stage3 pending CI confirmation of `pytest` + `conformance`.

paudley:
perf(sparql-eval): single-threaded optimization backlog + correlated-EXISTS correctness fix

Closes #20.

## Summary

Works the profiled, ranked single-threaded optimization backlog for
`purrdf-sparql-eval` (issue #20), and — through adversarial gap analysis of the
result — finds and fixes a real pre-existing correctness defect in correlated
`EXISTS` evaluation that the optimization work surfaced. Every item lands behind
the issue's hard gates: W3C SPARQL conformance untouched, byte-identical results,
`DetHasher` for new maps/sets, wasm-clean, no new deps, no Cargo features, clippy
pedantic/nursery clean. Perf items are measured with deterministic,
contention-independent metrics (heap-allocation counts; `perf stat -e
instructions:u`) because the shared build host swings ±17% on wall-clock.

## Optimizations (one commit each, per-item evidence in each message)

- **Item 3** — memoize constant `NamedNode`/`Literal` expression atoms per query
  (`const_atom_cache` on `EvalCtx`). `b_scan_filter`: −59,997 allocs (−25%).
- **Item 4** — memoize dataset literal XSD parses per query, keyed by `TermId`
  (`xsd_parse_cache`; `None` cached too). `b_scan_filter`: −5.5% instructions.
- **Item 5** — allocation-free single-column `u64` hash-join keys (`JoinKey`) +
  pre-sized build map. `c_optional_heavy`: −213,038 allocs (−17.4%).
- **Item 7** — O(1) `DetHashSet` visited-set for property-path transitive closure
  (`result` stays an ordered `BTreeSet` for byte-identical egress).
  `f_path_transitive`: −3.8% instructions.
- **Item 6a** — hoist the loop-invariant BGP probe permutation selection out of the
  per-row loop: `RdfDataset::probe_plan` (value-independent) + `candidate_run`
  (value-dependent), plan computed once per index-nested-loop slot via the new
  `QuadProbePlan`. `a_selective_join` −1.4%, `e_group_aggregate` −2.8% instructions.
  Byte-identical on all 7 `query_eval` shapes. (Item 6 tier-ii adaptive sort-merge
  was measured and deliberately NOT shipped: on every benched shape the join is
  selective — binary-search probing already optimal — or output-dominated, so an
  adaptive planner would never select it; shipping it would be
  unreachable-on-workload code, contrary to LOW/NO-OPTIONALITY.)
- **Item 2 residual** — pre-parse quoted-triple ORDER BY sort keys
  (`SortKey::Triple` holds precomputed component keys; `compare_triple_keys`
  recurses through `compare_sort_keys`), removing a per-comparison `parse_by_iri`
  from the quoted-triple sort arm. Guarded by an equivalence unit test (no
  `query_eval` witness because `g_order_by_limit` sorts integers, not triples).

Items 1 (regex compile cache) and 2 common ORDER BY path were verified already
implemented on `main` — the issue was filed against an earlier snapshot.

## Correctness fix (gap analysis outcome)

The optimization sweep exposed a pre-existing defect: correlated `EXISTS`
substitutes each outer row's bindings into a fresh `substitute_pattern` heap
temporary and evaluates it, but three per-query caches keyed those temporary
`Expression`/`GraphPattern` nodes by raw address — `const_atom_cache`,
`exists_expr_vars_cache`, `exists_inner_cache`. The temporary is dropped each row;
the allocator recycles its address; a later row hits a stale entry and reads the
prior row's constant / correlation decision / inner index → a silently wrong
solution set for correlated (and doubly-nested) `EXISTS`. Deterministic but
allocation-order-dependent (correct in isolation, wrong once other evaluations
warm the allocator). The `exists_*` caches carry the hazard independently of the
Item 3 memo, so the defect predates the whole backlog.

Fix: `EvalCtx::in_substituted_exists`, set with save/restore around the
substituted-pattern eval (so doubly-nested correlated `EXISTS` is handled) and
bypassing all three address-keyed caches while set. Address keying stays sound for
the static query algebra, which outlives the query; only per-row temporaries are
excluded, so the common path keeps full memoization at zero perf cost.

Root cause was proven by applying the fix to the pre-Item-6a tree (without the BGP
hoist) and confirming the query is deterministically correct in every evaluation
pattern. Item 6a had only *masked* the symptom for one query by perturbing
allocation order; it is retained solely as a valid perf win.

A stale strict-xfail ledger entry
(`test_nested_filters.py::test_nested_filter_outermost_binding_propagation`) now
computes the correct set and is removed — the vendored rdflib conformance case
passes as an ordinary test (no XPASS/XFAIL deferral). A native `sparql-eval`
regression asserts a correlated `EXISTS` over multiple outer rows is correct and
memo-invariant.

## Files

- `crates/sparql-eval/src/expr.rs` — `const_atom`/`exists` cache bypass under the
  substituted-EXISTS window; the constant-atom memo; new correlated-EXISTS
  regression test.
- `crates/sparql-eval/src/eval.rs` — `EvalCtx` fields `const_atom_cache`,
  `xsd_parse_cache`, `in_substituted_exists`; the XSD-parse memo.
- `crates/sparql-eval/src/bgp.rs` — hoist the probe plan out of the row loop.
- `crates/sparql-eval/src/binop.rs`, `scratch.rs` — `JoinKey` u64 encoding +
  pre-sized build map.
- `crates/sparql-eval/src/modifier.rs` — precomputed quoted-triple sort keys.
- `crates/sparql-eval/src/path.rs` — `DetHashSet` visited-set.
- `crates/rdf-core/src/ir/dataset.rs`, `ir/mod.rs`, `lib.rs` — `QuadProbePlan` +
  `probe_plan`/`candidate_run` split, re-exported (opaque, private fields).
- `bindings/python/tests/rdflib_suite/xfail_ledger.toml` — remove the now-passing
  correlated-EXISTS conformance entry.

## Standing constraints (`.goals`) respected

- GREENFIELD/SOTA/NO COMPROMISES: root-caused and fixed the real defect rather than
  accepting the fragile 6a masking or a ledger lie.
- RUST-FIRST: all engine work in Rust; the Python surface is unchanged.
- LOW/NO OPTIONALITY, HARD FAILS: no Cargo features; Item 6 tier-ii not shipped
  precisely to avoid unreachable optionality; `in_substituted_exists` is a
  correctness invariant, not a config toggle.
- MAXIMAL PERFORMANCE / PORTABILITY: measured per-item gains; wasm-clean core
  (`make wasm` green).

## Verification

`make check` green (fmt + clippy pedantic/nursery + build + workspace tests +
rdf-core hygiene + `make wasm`). rdflib conformance: 62 passed / 24 xfailed,
xpassed=0, failed=0, errors=0, ledger 24/24 applied, stale=0. `pytest tests`: 325
passed, 1 skipped, 7 xfailed. `cargo test -p purrdf-sparql-eval`: 262 passed. All
PR CI checks green (pytest, conformance, test, workspace, wasm, msrv, benchmarks,
CodeQL, CodeRabbit).

## Conflicts

None — branch was already current with `origin/main` at merge time.

PR #44, opened 2026-07-04
feat(shapes): SHACL-SPARQL custom constraint components

paudley:
Closes #12

## Summary
Implement SHACL-SPARQL custom constraint components (sh:ConstraintComponent) with ASK and SELECT validators, parameter pre-binding, and /home/paudley/.local/bin:/home/paudley/.bun/bin:/home/paudley/perl5/bin:/home/paudley/.kimi-code/bin:/home/paudley/.local/bin:/run/user/1000/fnm_multishells/2250200_1783098922347/bin:/home/paudley/.local/share/fnm:/home/paudley/.local/bin:/run/user/1000/fnm_multishells/2250199_1783098922346/bin:/home/paudley/.local/share/fnm:/home/paudley/stage/root/bin:/home/paudley/.scripts:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/opt/rocm/bin substitution.

## Changes
- Add SHACL vocabulary constants for constraint components and validators.
- Add components module that parses the component registry from the shapes graph.
- Detect component usage during shape parsing and emit Constraint::Component.
- Evaluate ASK validators (false → violation, one per value node) and SELECT validators (one per solution row).
- Implement /home/paudley/.local/bin:/home/paudley/.bun/bin:/home/paudley/perl5/bin:/home/paudley/.kimi-code/bin:/home/paudley/.local/bin:/run/user/1000/fnm_multishells/2250200_1783098922347/bin:/home/paudley/.local/share/fnm:/home/paudley/.local/bin:/run/user/1000/fnm_multishells/2250199_1783098922346/bin:/home/paudley/.local/share/fnm:/home/paudley/stage/root/bin:/home/paudley/.scripts:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/opt/rocm/bin textual substitution for property-shape SELECT validators.
- Pre-bind component parameters by parameter name.
- Remove three passing component tests from the W3C SHACL XFAIL ledger.
- Reach 117/120 SHACL-SPARQL conformance tests passing.

## Summary by CodeRabbit

* **New Features**
  * Added support for custom SHACL-SPARQL constraint components, including ASK and SELECT validators.
  * Improved handling of component parameters, messages, severities, and property/node shape usage.
  * SPARQL queries now support substitution-based evaluation for more accurate validation results.

[...]

Issue #28, opened 2026-07-02
SHACL-AF: sh:expression node constraints + node-expression evaluator (validation scope)


paudley:
## Goal

Implement the SHACL Advanced Features **Expression Constraints** (`sh:expression`, `sh:ExpressionConstraintComponent`): each value node must make the associated *node expression* evaluate to `true`. This requires a reusable **node-expression evaluator** — the AF query sublanguage — over the frozen IR, staying within purrdf's validation-only remit (no triple production).

## Why this fits purrdf

Expression constraints are pure *validation*: they read the data graph and check a boolean, exactly like `sh:sparql`. The node-expression evaluator is the shared substrate the rest of the AF validation surface builds on. It does **not** infer or assert new triples, so it stays inside the "carrier, no inference" line — unlike SHACL Rules (see Non-goals).

## Deliverables

- [ ] Node-expression evaluator over `RdfDataset` (interned `TermId` space), covering the AF node-expression kinds:
  - constant term, focus node (`sh:this`)
  - path expressions (`sh:path`, reusing `crate::path`)
  - filter-shape expressions (`sh:filterShape` + `sh:nodes`)
  - `sh:union` / `sh:intersection`
  - `sh:if` / `sh:then` / `sh:else`
  - aggregations `sh:count` / `sh:distinct` / `sh:min` / `sh:max` / `sh:sum`
  - ordering / paging `sh:limit` / `sh:offset` / `sh:orderby`
  - `sh:exists`
  - function-call expressions (built-in + `sh:SPARQLFunction`, gated on the functions issue)
- [ ] `sh:expression` parsed from the shapes graph and wired as `sh:ExpressionConstraintComponent` in `crates/shapes/src/constraints.rs`, producing standard `ValidationResult`s (`sh:sourceConstraintComponent = sh:ExpressionConstraintComponent`)
- [ ] Vendor the W3C SHACL-AF `expression`/`node` manifests into `vectors/shacl/af/…` and extend the `w3c_conformance` harness (relates to #8); ledger any residuals with precise reasons
- [ ] Deterministic ordering/dedup for aggregations so validation reports stay byte-stable

## Non-goals

- SHACL **Rules** (`sh:rule` / `sh:TripleRule` / `sh:SPARQLRule`) — they *derive* triples; out of scope for the no-inference carrier.
[...]

PR #45, opened 2026-07-04
feat(shacl-af): sh:SPARQLFunction user-defined SPARQL functions

### LINKED ISSUE #29: SHACL-AF: SPARQL-based functions (sh:SPARQLFunction)
## Goal

Support user-declared **`sh:SPARQLFunction`** (and the surrounding `sh:Function` machinery): functions defined in the shapes graph with `sh:parameter` / `sh:returnType` and a `sh:select` or `sh:ask` body, callable from SHACL-SPARQL constraint/target queries and from node expressions (the `sh:expression` issue).

## Why this fits purrdf

These functions extend the *evaluation* surface the validator already runs — they compute values inside queries/expressions, they don't assert triples. That's validation-adjacent and in scope.

## Current state

The native engine's extension-function set is **CLOSED / exhaustive** today (built-in `rdf:List` functions under a caller-configured namespace — see `crates/sparql-eval/src/expr.rs:1402` and `crates/sparql-eval/src/list_fn.rs`). `sh:SPARQLFunction` needs a *dynamic* registry, populated from the shapes graph and threaded into evaluation via a host-provided callback seam.

## Deliverables

- [ ] Parse `sh:SPARQLFunction` declarations from the shapes graph (parameters with `sh:order`/`sh:optional`, `sh:returnType`, `sh:select`/`sh:ask` body)
- [ ] A dynamic function registry the shapes engine passes to the SPARQL evaluator (a callback seam, not a hardcoded set), reusing the caller-configured extension-function namespace policy
- [ ] Parameter binding + return-value extraction; arity / type checks per spec
- [ ] Reachable from `sh:sparql` / `sh:SPARQLTarget` queries and from node expressions
- [ ] Vendor the W3C SHACL-AF `function` manifests into `vectors/shacl/af/…` and extend the harness (relates to #8)

## Non-goals

- General-purpose user functions outside the SHACL validation surface — that's a SPARQL-engine feature; file separately if wanted.
- Rule-based triple derivation.

## Gates

`cargo test -p purrdf-shapes -p purrdf-sparql-eval` green; clippy `-D warnings`; wasm32-clean; no Cargo features.

Part of the SHACL-AF completion effort.

paudley:
Closes #29

## Summary
Implements SHACL-AF SPARQL-based functions (`sh:SPARQLFunction`): parse function declarations from the shapes graph (ordered `sh:parameter`s, `sh:optional`, `sh:returnType`, `sh:select`/`sh:ask` body), build a dynamic user-function registry, and resolve calls at eval time so the functions are callable from `sh:sparql` / `sh:SPARQLTarget` queries and `sh:expression` node expressions alike. Removes the prior capability stub.

## Changes
- **purrdf-sparql-eval**: `UserFunctionRegistry` (pure data) resolved in the `Function::Custom` eval arm — the open counterpart to the closed parse-time `PurrdfFn` dispatch. Registry threaded as a `Copy` borrow on `EvalCtx` (mirrors `remote`/`bgp_order_cache`), propagated through `fork_for_worker` and a recursion-bounded `child_for_user_fn`. New `NativeSparqlEngine::query_with_user_functions` (mirrors `query_with_source`); new `EvalError::Function` variant.
- **purrdf-shapes**: parse `sh:SPARQLFunction` declarations into the registry (variable = predicate local name, ordered by `sh:order` then IRI; hard-fails on missing predicate, colliding variable names, required-after-optional, both/neither body, unparsable body). Registry installed for a validation pass via a scoped thread-local guard; `run_select` routes through the user-function engine path. `FnCall::Builtin`/`UserDefined` node-expression arms collapse to one; stub removed.
- **corpus**: four first-party `example.org` cases (select-from-expression, sh:sparql constraint, ask boolean, optional parameter) + case-count ratchet 64→68; `vectors/shacl/af` seam stays empty (upstream ships no AF manifests).

## Verification
`make check` (fmt + clippy pedantic/nursery -D warnings + build + tests + hygiene) and `make wasm` both green; `cargo test -p purrdf-shapes -p purrdf-sparql-eval` green; W3C conformance unchanged (no XFAIL/XPASS drift). No Cargo features added; oxigraph/PyO3-free; wasm32-clean.

## Summary by CodeRabbit

* **New Features**
  * Added support for SHACL-AF SPARQL-based functions during validation, including function calls inside expressions and constraints.
  * Added handling for optional function parameters, boolean `ASK`-style functions, and return/type checks.
  * Expanded the conformance corpus with new function-focused validation scenarios.

* **Bug Fixes**
  * Validation now reports function-related errors more clearly when a function call is invalid, missing, or exceeds recursion limits.
  * Unknown function calls now fail explicitly instead of being handled ambiguously.


paudley:
# Implementation Plan for Issue #29 — SHACL-AF SPARQL-based functions (`sh:SPARQLFunction`)

## Context

SHACL-AF lets a shapes graph *declare its own functions* — `sh:SPARQLFunction`
declarations with ordered `sh:parameter`s, a `sh:returnType`, and a `sh:select`
or `sh:ask` body — that are then callable from SHACL-SPARQL constraint/target
queries (`sh:sparql`, `sh:SPARQLTarget`) and from SHACL-AF node expressions
(`sh:expression`, delivered by #28). They extend the *evaluation* surface (they
compute values inside queries/expressions), not the assertion surface, so they
are squarely in scope for PurRDF.

#28 deliberately left the seam stubbed: `FnCall::UserDefined` hard-errors with
"…requires the dynamic SPARQL function registry capability, which is not yet
available" (`crates/shapes/src/expression.rs:440`). This issue builds that
capability and removes the stub.

### Key discoveries from exploration (ground truth)

- **Parse side already exists.** `crates/shapes/src/shapes.rs:1942` already types
  a call IRI as `FnCall::UserDefined` when the IRI `has_type sh:SPARQLFunction`
  (or `sh:Function`). Missing: parsing the *declaration* (params/returnType/body).
- **`sh:sparql` UDF calls already parse.** The shapes SPARQL layer runs a **bare**
  `NativeSparqlEngine::new()` with **no** `extension_fn_namespaces`
  (`crates/shapes/src/sparql.rs:356`). Per `crates/sparql-algebra/src/parser.rs:2514`,
  an IRI in call position under *no* configured namespace parses to
  `Function::Custom(iri)` — no hard error, IRI preserved verbatim. So **no parser
  change is required**; we resolve at eval time.
- **Eval seam is one arm.** `crates/sparql-eval/src/expr.rs:1624` `Function::Custom`
  does XSD casts then `EvalError::unsupported`. We insert a registry lookup before
  the error.
- **`EvalCtx` threading pattern is established.** New optional fields are threaded
  exactly like `standpoint_predicates` (`crates/sparql-eval/src/eval.rs:184`) and
  `remote` (`eval.rs:243`), including propagation through the sub-context clone
  (`eval.rs:~510` — EXISTS/subquery contexts). `NativeSparqlEngine` mirrors with a
  `with_user_functions` builder (like `with_standpoint_predicates`,
  `engine.rs:171`) and threads it in `eval_ctx` (`engine.rs:180`).
- **Recursion discipline exists** as `RecursionGuard`/`MAX_RECURSION_DEPTH`
  (`expression.rs:173-183`, fail-closed at 64). We add an analogous UDF call-depth
  bound on `EvalCtx`.
- **W3C corpus has NO AF manifests** (`vectors/shacl/af/manifest.ttl` is a
  deliberately-empty seam). Coverage is authored as first-party `example.org`
  fixtures under `crates/shapes/corpus/NN-*` (three files: `shapes.ttl`, `data.nt`,
  `expected-report.nt`), with an exact case-count ratchet in
  `crates/shapes/tests/conformance.rs:49`.

### SHACL-AF binding semantics to implement (from the spec, not the repo)

- A `sh:SPARQLFunction` has zero or more `sh:parameter`s. Each parameter has an
  `sh:path` (a single predicate IRI); **the pre-bound SPARQL variable name is the
  local name of that predicate**. Parameters are ordered by `sh:order` (ascending;
  unspecified order sorts by IRI as a deterministic tiebreak). Positional call
  arguments bind to parameter variables in that order.
- `sh:optional true` marks a trailing optional parameter → arity is
  `[required_count, total_count]`.
- `sh:returnType` / per-parameter `sh:datatype`/`sh:nodeKind` are type constraints.
  Per `.goals` (LOW/NO OPTIONALITY, HARD FAILS) we **enforce** them: arity or type
  violation is a hard evaluation error, never a silent skip.
- `sh:ask` body → returns `xsd:boolean`. `sh:select` body → returns the single
  projected variable of the first solution row; empty result → no value
  (`Ok(None)`, mirroring `eval_scalar_expr`'s empty-is-no-value contract).

- Adversarial gap analysis (assume substandard until proven otherwise).
- **No deferrals / followups / "future work"** — everything in scope is done now.
- **No GitHub issue/PR numbers in repo source, ontology, or docs** — process flow
  lives in GitHub only. (Applies to code comments and corpus README additions.)
- Own every bug you touch; fix in place.

- GREENFIELD-FIRST / NO COMPROMISES — implement the full seam (callable from both
  `sh:sparql` queries and node expressions), not a node-expression-only shortcut.
- RUST-FIRST — all logic in Rust; no Python surface change required here.
- SUBSUME/EXTEND/ENHANCE — extend the native engine's function dispatch, don't bolt
  on a parallel path.
- LOW/NO OPTIONALITY, HARD FAILS — no Cargo features; arity/type violations and
  recursion overflow fail fast and loud.
- MAXIMAL UTILITY / PERFORMANCE / PORTABILITY — keep the seam pure-Rust,
  oxigraph-free, **wasm32-clean** (`make wasm`); registry lookup is off the hot path
  (only reached for `Function::Custom` IRIs, which are rare).

## Non-goals (respect)

- General-purpose user functions outside SHACL validation (the registry is
  populated only from the shapes graph during validation).
- Rule-based triple derivation.

---

```bash
BRANCH="paudley/29-shacl-af-sparql-function"
WORKTREE="$REPO_ROOT/.worktrees/29-shacl-af-sparql-function"

cd "$WORKTREE"
git branch --show-current   # MUST show paudley/29-…, NOT main
```

Re-establish a green baseline in the fresh worktree before implementing:

```bash
make check      # fmt + clippy (pedantic/nursery, -D warnings) + build + tests + hygiene
```

Post this plan to the issue:

```bash
gh issue comment 29 --body-file /home/paudley/.claude/plans/29-abundant-squirrel.md
```

**⚠️ Do NOT proceed to Task 1 until the worktree exists, you are on the feature branch, and `make check` is green.**

---

## Task 1: Dynamic user-function seam in `purrdf-sparql-eval`

Add the eval-time resolution path so any `Function::Custom` IRI known to an
injected registry executes as a SHACL-AF function. This is the foundation both
call surfaces reuse.

**Registry data model** (new module `crates/sparql-eval/src/user_fn.rs`, or a
type in `eval.rs`):

- `UserFunction { params: Vec<String>, required: usize, body: Arc<Query>, kind: Select | Ask, return_type: Option<NamedNode> }`
  where `params` are the pre-bound variable names in call order, `body` is the
  already-parsed `sh:select`/`sh:ask` query, `kind` selects extraction.
- `UserFunctionRegistry` = deterministic map `iri (String) -> UserFunction`
  (`DetHashMap`), wrapped `Arc` for cheap injection. Provide
  `fn resolve(&self, iri: &str) -> Option<&UserFunction>`.

**Request-scoped injection (keeps the thread-local engine + its plan cache).**
Rather than making the registry engine-level state (which would force replacing the
thread-local engine and lose cross-run cache reuse), thread it as a **borrow on the
request**, so the engine stays stateless w.r.t. shapes-graph data (better SRP):

- `SparqlRequest` (`crates/rdf-core/src/backend.rs:117`) is `Copy` via borrow-only
  fields. Add `pub user_functions: Option<&'a UserFunctionRegistry>` — a borrow keeps
  `Copy` intact. (`UserFunctionRegistry` lives in `sparql-eval`;
  `rdf-core` is a lower crate — if the type can't be named there, thread it via a
  small trait object `Option<&'a dyn UserFnResolver>` defined in `rdf-core`, or move
  the registry type down. Confirm the dependency direction in Task 1: `sparql-eval`
  already depends on `sparql-algebra`, `shapes` depends on both — no cycle from
  storing `Arc<purrdf_sparql_algebra::Query>` in the registry.)

**Thread on `EvalCtx`** (`crates/sparql-eval/src/eval.rs`):

- Add `pub(crate) user_functions: Option<&'d UserFunctionRegistry>` beside
  `standpoint_predicates` (`eval.rs:184`) and a `udf_depth: u32` counter, populated
  from the request in the engine's `query` path (`engine.rs:239`).
- **Propagate both fields in the sub-context clone at `eval.rs:~510`** — the block
  that copies `standpoint_predicates`/`remote`/`bgp_order_cache`. Missing this
  silently drops UDFs inside EXISTS/subqueries.
- Plan-cache key is unaffected: UDF calls parse to `Function::Custom` (IRI-preserving,
  registry-independent at parse time), so the existing key
  `(base, extension_fn_namespaces, text)` (`engine.rs:69`) stays correct. Add a code
  comment noting this invariant (no issue numbers).

**⚠️ Confirm before coding (the make-or-break detail).** The `Function::Custom` arm
holds `&mut EvalCtx`, not the engine. Executing a `sh:select`/`sh:ask` body means
evaluating a whole algebra `Query` **seeded with pre-bound parameter variables**
(`params[i] → vals[i]`) and projecting a scalar — which EXISTS does *not* do (EXISTS
returns a boolean over existing bindings, no seeded initial mapping). First verify
`eval.rs` exposes a lower-level entry point to (a) evaluate an algebra `Query`/pattern
from within expression context and (b) seed its initial solution mapping. If the only
seeding path is `SparqlRequest.substitutions` at the `engine.query` boundary, add the
lower-level seam. Do this spike first; the rest of Task 1 depends on it.

**Executor (`crates/sparql-eval/src/user_fn.rs`) + thin dispatch arm.** Keep the
`Function::Custom` arm (`expr.rs:1624`) thin — resolve then delegate:

1. Before the `unsupported` error, if `ctx.user_functions` resolves `iri`, call
   `user_fn::eval(func, &vals, ctx)`; the executor does:
   - Arity-check `vals.len()` against `[required, params.len()]` → hard `EvalError`.
   - Per-arg `sh:datatype`/`sh:nodeKind` and `return_type` checks → hard `EvalError`.
   - Bind each already-evaluated `vals[i]` to `params[i]`, evaluate `body` in a fresh
     sub-context (same dataset/graph, `udf_depth + 1`, registry carried). Guard
     `udf_depth` against a deliberately-chosen `MAX_UDF_DEPTH` (fail-closed). Note the
     two depth counters do not compose: shapes-side `RecursionGuard` (64) bounds
     shape re-entry; `MAX_UDF_DEPTH` bounds UDF→UDF chains inside SPARQL eval.
   - Extract: `Ask` → `Boolean` term; `Select` → first row's single projected column
     (`rows[0][0]`); empty → `Ok(None)`.
2. XSD-cast and `unsupported` fallbacks stay after the registry miss.

**Tests (in-crate):** a unit test that evaluates a query with a hand-built
`UserFunctionRegistry` supplied via `SparqlRequest.user_functions` (both SELECT- and
ASK-bodied UDFs), plus arity-error and recursion-depth-error cases.

### After Task 1:
- Commit (github-commit-agent) · Push (github-push-agent) · wiring-verifier · post summary to issue #29

---

## Task 2: Parse `sh:SPARQLFunction` declarations in `purrdf-shapes`

**IRI consts** (`crates/shapes/src/model.rs`, `sh` module): add the missing
`PARAMETER` (`#parameter`), `OPTIONAL` (`#optional`), `RETURN_TYPE`
(`#returnType`), and `PREDICATE` (`#predicate`). Reuse existing `ORDER`
(`model.rs:127`), `SELECT` (`:182`), `ASK` (`:185`), `PATH` (`:61`), `DATATYPE`
(`:79`), `NODE_KIND` (`:81`), `SPARQL_FUNCTION` (`:249`), `FUNCTION` (`:251`).

**Declaration parse** (`crates/shapes/src/shapes.rs`): a pass that finds every
subject `has_type sh:SPARQLFunction`, and for each builds a `UserFunction`:

- Parameters: `objects_of(fn, sh::PARAMETER)`; for each, read `sh:path`/`sh:predicate`
  (predicate IRI → variable = local name), `sh:order` (default: sort by predicate
  IRI), `sh:optional`, and optional `sh:datatype`/`sh:nodeKind`. Sort ascending by
  `(order, predicate-iri)`; `required` = count of non-optional. Reuse the existing
  helpers `objects_of` (`shapes.rs:526`), `first_object_of` (`:539`),
  `has_type` (`:508`). **Hard error on two parameters whose derived variable names
  (predicate local names) collide** — silent shadowing violates HARD-FAILS.
- Body: exactly one of `sh:select` / `sh:ask` (hard error if both/neither). Prepend
  the computed prefix header via `prefix_header` (`shapes.rs:577`) — same as
  `sh:sparql` parsing (`shapes.rs:1076`) — and parse with
  `purrdf_sparql_algebra::SparqlParser` into a `Query`.
- `sh:returnType` → optional `NamedNode`.
- Assemble a `UserFunctionRegistry` (Task 1 type). Store it on the compiled shapes /
  `ValidationEngine` so it is built **once per shapes graph**.

**Tests:** parse a shapes graph with a `sh:SPARQLFunction` and assert the registry
entry (params in order, required count, kind, return type).

### After Task 2:
- Commit · Push · wiring-verifier · post summary to issue #29

---

## Task 3: Wire the registry into evaluation + remove the stub

**Thread the registry through the shapes SPARQL seam**
(`crates/shapes/src/sparql.rs`): keep the thread-local bare engine (`sparql.rs:356`)
and its plan cache — the registry is **request-scoped**, not engine state. Pass
`Option<&UserFunctionRegistry>` down through the `eval_target` /
`eval_sparql_constraint` / `eval_scalar_expr` / `eval_aggregate` / `eval_order`
signatures (`sparql.rs:41,83,158,220,306`) into `run_select` (`sparql.rs:368`),
which sets it on the `SparqlRequest.user_functions` field. Empty/`None` registry →
current behavior, zero overhead. The registry is built once per shapes graph
(Task 2) and stored on the `ValidationEngine`, then borrowed into each eval call.

**Node-expression dispatch** (`crates/shapes/src/expression.rs`): thread the
registry through `eval_node_expr` (`expression.rs:317`) and its ~15 recursive call
sites plus the `aggregate` helper (`expression.rs:582`). Replace the
`FnCall::UserDefined` hard-error arm (`expression.rs:440`) so it renders
`<iri>(?a0, …)` and routes through `eval_scalar_expr` **exactly like the `Builtin`
arm** (`expression.rs:407-438`) — the engine now resolves the UDF via the registry,
so `Builtin` and `UserDefined` collapse to the same rendering with the same arity
logic (`expression.rs:411-421`).

**Fix the stale doc comment** on `eval_node_expr` (`expression.rs:313-316`), which
states it errors for the user-defined-function kind — a lie once the stub is gone
(own-every-bug-you-touch).

**Update the two tests that assert the old stub:**
`user_defined_call_is_capability_error` (`expression.rs:1013`) becomes a
success/dispatch test; `if_propagates_condition_error` (`expression.rs:790`) must be
re-pointed at a still-erroring construct (or updated to expect success).

### After Task 3:
- Commit · Push · wiring-verifier · post summary to issue #29

---

## Task 4: Conformance corpus fixtures + ratchet

The pinned W3C corpus ships no AF `function` manifests, so author first-party
`example.org` fixtures (same pattern as the #28 expression cases,
`crates/shapes/corpus/49-*…64-*`). Add new `crates/shapes/corpus/NN-*` dirs
(`shapes.ttl` + `data.nt` + `expected-report.nt`), covering:

- `sh:SPARQLFunction` with a `sh:select` body called from `sh:expression`.
- Same called from a `sh:sparql` constraint (proves the `sh:sparql` surface).
- `sh:ask`-bodied boolean function.
- Optional-parameter arity and a type/arity **violation** (expected error/report).

Bump the exact case-count const in `crates/shapes/tests/conformance.rs:49`. Update
`crates/shapes/corpus/README.md` with the new case descriptions (no GitHub/PR
numbers). Leave `vectors/shacl/af/manifest.ttl` as the empty seam (the harness
already auto-discovers it, `crates/shapes/tests/w3c_conformance.rs:482`); `TOTAL_TESTS`
(`w3c_conformance.rs:68`) is unchanged since no W3C AF manifest is added.

### After Task 4:
- Commit · Push · wiring-verifier · post summary to issue #29

---

## Task 5: Full gate + Create Pull Request (MANDATORY FINAL STEP)

1. Verify everything is committed/pushed and the full gate is green:
   ```bash
   make check    # fmt + clippy pedantic/nursery -D warnings + build + tests + hygiene
   make wasm     # every release crate stays wasm32-unknown-unknown-clean
   cargo test -p purrdf-shapes -p purrdf-sparql-eval
   ```
2. Create PR:
   ```bash
   gh pr create --title "feat(shacl-af): sh:SPARQLFunction user-defined SPARQL functions" --body "Closes #29

   ## Summary
   Adds SHACL-AF SPARQL-based functions: parse sh:SPARQLFunction declarations
   (params/order/optional/returnType, sh:select|sh:ask body), a dynamic user-function
   registry injected into the native SPARQL engine, and eval-time resolution making the
   functions callable from sh:sparql / sh:SPARQLTarget queries and sh:expression node
   expressions. Removes the #28 capability stub.

   ## Changes
   - purrdf-sparql-eval: UserFunctionRegistry + EvalCtx/engine seam + Function::Custom resolution + UDF-depth guard
   - purrdf-shapes: sh:SPARQLFunction declaration parsing; new sh:parameter/optional/returnType/predicate IRI consts
   - purrdf-shapes: registry threaded through sparql.rs / expression.rs; FnCall::UserDefined dispatch (stub removed)
   - corpus: first-party example.org sh:SPARQLFunction fixtures + case-count ratchet bump
   "
   ```
3. Post plan to PR:
   ```bash
   gh pr comment $PR_NUM --body-file /home/paudley/.claude/plans/29-abundant-squirrel.md
   ```
4. Post to PR and issue the answers to:
5. Report: "Stage 1 complete. PR #<num> created for issue #29. Ready for /stage2 29".

---

## Execution Rules

1. Task 0 (worktree) is MANDATORY — never work on `main`.
2. Commit after EACH task (github-commit-agent), never batch.
3. Push after EACH commit (github-push-agent).
4. Task 5 (PR) is MANDATORY final step.
5. Use sub-agents for all commits/pushes; wiring-verifier after each task.

## Verification (end-to-end)

- **Unit:** `cargo test -p purrdf-sparql-eval` (registry executor: SELECT/ASK bodies,
  arity error, recursion overflow) and `cargo test -p purrdf-shapes` (declaration
  parsing + dispatch).
- **Corpus:** `cargo test -p purrdf-shapes --test conformance` runs the new
  `example.org` fixtures; the exact case-count ratchet proves they are discovered.
- **W3C:** `cargo test -p purrdf-shapes --test w3c_conformance` stays at
  `TOTAL_TESTS` with no XFAIL/XPASS drift.
- **Gate:** `make check` (clippy pedantic/nursery `-D warnings`) and `make wasm`
  (wasm32-clean) both green. No Cargo features added; no oxigraph/PyO3 near
  `purrdf-core`; serializers/GTS determinism untouched.

paudley:
## Stage 1 reflection

**What I am least confident about, and why**

1. **Registry threading via a scoped thread-local (vs. explicit parameter threading).** `run_select` reads the in-scope `UserFunctionRegistry` from a thread-local guard installed at the top of `validate_with_focus_filter`, rather than threading `Option<&Registry>` through `eval_node_expr`'s ~15 recursive call sites + `constraints.rs`. This is correct because validation is serial (documented) and parallel FILTER workers get the registry through `EvalCtx` (propagated in `fork_for_worker`), and it is idiomatic to this module (which already holds the engine in a thread-local). But it is implicit state — if validation is ever made concurrent at the focus-node level, the guard must move onto the per-thread work, not just the driver thread.

2. **SHACL-AF parameter/return semantics have no upstream corpus to cross-check.** The pinned W3C data-shapes corpus ships **no** AF `function` manifests, so the pre-bound-variable convention (local name of the parameter's `sh:path`/`sh:predicate` predicate), the `[required, total]` optional-arity rule, and the SELECT-first-projected-value / ASK-boolean extraction are validated only against first-party `example.org` fixtures + unit tests, not an authoritative third-party suite. I implemented them to the spec text; a real-world shapes graph using an unusual parameter shape could surface a gap.

**What you do not know but should**

1. **`origin/main` is currently broken.** The tip of `origin/main` (`a7fb8ce`, "implement SHACL-SPARQL custom constraint components") has **committed, unresolved git conflict markers** (`<<<<<<< HEAD`) in `crates/shapes/src/constraints.rs` and `crates/shapes/src/shapes.rs` — that commit does not compile. It landed after this work started. **This PR is based on the older clean commit `950b63f`**, builds green, and is unaffected, but it will need a rebase once `main` is repaired, and merging is blocked until then. I did not touch `main` (repairing the shared trunk is an outward-facing action I am leaving to you).

2. **User functions are deliberately scoped to SHACL validation only.** The registry is populated solely from the shapes graph during a validation pass (per the issue's non-goals); general-purpose user functions in arbitrary SPARQL are out of scope. If a caller expects `myfn:foo()` to resolve in a bare `NativeSparqlEngine::query`, it will not — they must use `query_with_user_functions`.

paudley:
# Stage 2 Plan — Issue #29 (`sh:SPARQLFunction`) + urgent `main` repair

## Context

`/stage2 29` runs gap analysis + remediation on the SHACL-AF `sh:SPARQLFunction`
branch `paudley/29-shacl-af-sparql-function` (PR #45, worktree at
`.worktrees/29-shacl-af-sparql-function`, tip `9d4f2bf`, base `950b63f`).

Two independent problems surfaced:

1. **`origin/main` is broken.** Commit `a7fb8ce` ("feat(shapes): implement
   SHACL-SPARQL custom constraint components") committed **stray, unmatched
   `<<<<<<< HEAD` conflict-marker lines** into two files:
   - `crates/shapes/src/shapes.rs:218`
   - `crates/shapes/src/constraints.rs:1412`
   Each file has only the opening marker (no `=======`/`>>>>>>>`); surrounding
   code is intact. `main` therefore does not compile. This blocks Stage 3 for
   #29 (which merges `main` in) and any other branch. **User-approved to fix
   first.**

2. **The #29 branch has real gaps.** The worktree itself is marker-clean and the
   `sh:SPARQLFunction` seam is genuinely wired end-to-end (parse → `Shapes.functions`
   registry → thread-local scope guard → `EvalCtx.user_functions` →
   `Function::Custom` → `eval_user_function`; the #28 `FnCall::UserDefined` stub
   is removed). But **all three PR-review findings are unaddressed** (two are
   genuine spec bugs), and the negative-path conformance coverage the Stage-1
   plan promised was never authored.

Intended outcome: `main` compiles again, and the #29 branch is spec-correct,
review-clean, and ready for `/stage3`.

---

## Execution rules (from stage2 workflow + `.baseline`/`.goals`)

- **Commit after EACH fix — never batch.** Per gap: implement → `github-commit-agent`
  (reads `PRECOMMIT_GUIDANCE.md`, iterates to green) → `github-push-agent` →
  post summary to issue #29 AND PR #45 → `wiring-verifier`.
- Done-gate for every commit: `make check` (fmt + clippy pedantic/nursery
  warning-free + build + tests + hygiene) and wasm-cleanliness where touched.
- No Cargo features; no hand-edited `generated/`/`vectors/`; deterministic
  serialization; no issue/PR numbers in source or docs; no deferrals.

---

## Step 0 — CRITICAL: repair `origin/main` (do this FIRST, on `main`)

Operate in the primary tree `/home/paudley/Active/purrdf` on branch `main`.

1. Delete the single stray line `<<<<<<< HEAD` at:
   - `crates/shapes/src/shapes.rs:218` (between the `QualifiedValueShape`
     variant's closing `},` and the `Expression {` variant doc-comment)
   - `crates/shapes/src/constraints.rs:1412` (between the `sh:sparql` arm and the
     `// ── Expression (SHACL-AF §5.7)` comment / `Constraint::Expression` arm)
   No other edits — the code on both sides is already correct and complete.
2. Verify: `make check` (must build + pass; this is the proof `main` is healed).
3. Commit directly to `main` as a hotfix (user-approved direct-to-main):
   `fix(shapes): remove stray conflict markers left by a7fb8ce` and push to
   `origin/main`.

> This is the one direct-to-`main` action; everything below happens on the #29
> branch/worktree. Rebase of the #29 branch onto healed `main` is a Stage-3
> concern, not part of this plan.

---

## Step 1 — HIGH: unbound mandatory parameter must return no result

**File:** `crates/sparql-eval/src/user_fn.rs` (~`eval_user_function`, lines ~208-228).
`args: &[Option<TermValue>]`; unbound argument expressions yield `None`
(`expr.rs:1318-1320`). The binding loop `if let Some(value)`-skips a `None`,
leaving a *mandatory* parameter variable unbound and running the body anyway.
SHACL-AF §5.2/§9.5: a call missing any mandatory parameter **returns no result
node**.

**Fix:** in the binding loop, when the arg is `None` and `idx < func.required`,
`return Ok(None)` before executing the body (reviewer's exact patch). Optional
params (`idx >= func.required`) stay unbound as today.

**Covers reviewer finding A.** Add the mandatory-unbound fixture in Step 4.

---

## Step 2 — HIGH: stop enforcing `sh:returnType` as a runtime datatype constraint

**Files:** `crates/shapes/src/shapes.rs:621-624` (maps `sh:returnType` into a
`TypeConstraint.datatype`) and `crates/sparql-eval/src/user_fn.rs:258-260`
(runtime `return_constraint.check(...)`).

Per SHACL-AF §5.3, `sh:returnType` is **informational / for casting** and MAY be
a class IRI (e.g. `rdfs:Resource`) — not a literal datatype. Enforcing it as a
datatype hard-fails any function returning an IRI/blank node. This is latent only
because every current fixture returns `xsd:boolean`.

**Design-tension resolution (explicit):** the Stage-1 plan enforced this as a
"HARD FAILS" check. That is **spec-incorrect**, not strict — `.goals` HARD-FAILS
means fail fast on genuine errors, not manufacture a constraint the spec calls
informational. Spec correctness wins.

**Fix:** remove the runtime `return_constraint.check` call and stop mapping
`sh:returnType` into a datatype `TypeConstraint`. Revise the test
`return_type_violation_is_a_hard_error` (`user_fn.rs:446-480`) to codify the
correct behavior (an IRI/class-typed return is accepted; keep parsing
`sh:returnType` for informational retention only if it is still surfaced).

**Covers reviewer finding B.**

---

## Step 3 — MEDIUM: validate derived parameter variable names

**File:** `crates/shapes/src/shapes.rs:523` — `var = local_name(&predicate)` with
no guard. An empty local name produces an invalid `?`/`("", value)` binding; a
param predicate whose local name is `this` shadows SHACL's injected `$this`
focus-node binding.

**Fix:** hard-error on empty `var` and on the SHACL/SHACL-AF reserved set
(`this`, `path`, `PATH`, `value`, `shapesGraph`, `currentShape`) — reviewer's
patch. **Covers reviewer finding C.**

---

## Step 4 — MEDIUM: author the missing negative-path conformance fixtures

Existing corpus 65-68 cover only happy/violating paths; fixture 68 tests the
*optional* unbound case, never mandatory. The two HIGH bugs above have zero
conformance coverage.

**Add first-party `example.org` fixtures** (each `shapes.ttl` / `data.nt` /
`expected-report.nt`) under `crates/shapes/corpus/` for:
- mandatory parameter unbound → function yields no value (guards Step 1),
- return value that is an IRI with a class-valued `sh:returnType` → accepted
  (guards Step 2),
- arity mismatch and a per-parameter `sh:datatype`/`sh:nodeKind` violation →
  hard error,
- undefined-function call → hard error.

Then bump the count ratchet at `crates/shapes/tests/conformance.rs:49` and
**regenerate the frozen digest** `scripts/conformance-frozen/shapes-corpus.sha256`
via the repo's regeneration path (do not hand-edit the frozen file).

---

## Step 5 — LOW: tighten SELECT projection and malformed annotations

- `crates/sparql-eval/src/user_fn.rs:241-248` takes `rows[0][0]` with no check
  that the body projects exactly one variable → hard-error on multi-projection
  SELECT bodies at parse time (single result variable per SHACL-AF).
- `crates/shapes/src/shapes.rs:531-537` treats `sh:optional` as true only for the
  literal `"true"` (a `"1"^^xsd:boolean` is silently required) and maps any
  unparseable `sh:order` to `INFINITY` — make both **hard-fail** on malformed
  values, consistent with `.goals` HARD-FAILS.

---

## Post-approval bookkeeping

- After approval, post this plan to BOTH issue #29 and PR #45 (`--body-file`).
- Per-gap: post a summary (gap, priority, fix, commit hash) to issue AND PR.
- Feedback loop: after all gaps, re-check PR #45 comments (coderabbitai was
  rate-limited and never produced findings — expect a delayed pass); fold any new
  findings back as gaps.
- Final summary → issue AND PR; branch ready for `/stage3`.

## Verification

- Step 0: `make check` in the primary tree proves `main` compiles; confirm
  `git grep -n '<<<<<<< HEAD' HEAD` returns nothing after the hotfix.
- Each #29 step: `cargo test -p purrdf-shapes -p purrdf-sparql-eval` green,
  `make check` clean, `make wasm` clean; new/updated fixtures pass conformance
  and the frozen digest matches the regenerated corpus.
- End-to-end spot check: a `sh:SPARQLFunction` returning an IRI with a class
  `sh:returnType` validates without error; a constraint calling a function with a
  missing mandatory arg yields no result (no false conformance).

paudley:
### Gap 1 (HIGH) resolved — unbound mandatory `sh:SPARQLFunction` parameter yields no result

**Commit:** `e90bdd5`

Per SHACL-AF §5.2/§9.5, a function invoked without all mandatory parameters returns no result node. The binding loop in `eval_user_function` skipped unbound (`None`) arguments and executed the body anyway, so a mandatory parameter could evaluate unbound (reviewer finding A). Fix: return `Ok(None)` when a *required* parameter's argument is unbound; optional parameters still bind lazily. Added `unbound_mandatory_parameter_yields_no_value` (body ignores its parameter, so only the guard — not an unbound `?n` — can suppress the value). Test passes; `cargo clippy -p purrdf-sparql-eval -D warnings` and `cargo fmt --check` clean.

paudley:
### Gap 2 (HIGH) resolved — `sh:returnType` is informational, not an enforced datatype

**Commit:** `05594bc`

Per SHACL-AF §5.3, `sh:returnType` documents/casts a function's return and MAY be a class IRI, not a literal datatype. The code enforced it as a runtime datatype constraint, which spuriously hard-failed any function returning an IRI or blank node (reviewer finding B). This was latent only because every fixture returns `xsd:boolean`.

Fix: removed the return-value enforcement in `eval_user_function`. The parsed constraint is **retained** on `UserFunction` as informational metadata (a `pub` field, still surfaced/parsed), so nothing becomes dead code — consistent with §5.3. The former `return_type_violation_is_a_hard_error` test is replaced by `return_type_is_informational_not_enforced`, which asserts a class-typed IRI return is **accepted** (the pre-fix code wrongly errored). Test passes; clippy `-D warnings` and fmt clean.

**Design note:** this deliberately overrides the Stage-1 plan's "HARD FAILS" enforcement — that reading was spec-incorrect. HARD-FAILS means fail fast on genuine errors, not manufacture a constraint the spec calls informational.

paudley:
### Gap 3 (MEDIUM) resolved — reject empty / SHACL-reserved `sh:SPARQLFunction` parameter names

**Commit:** `837232b`

The parameter variable name derived from the predicate local name (`local_name(&predicate)`) was used unchecked (reviewer finding C). An empty local name yields an invalid `?` variable; a name like `this` shadows SHACL's injected focus-node binding (SHACL §3.2.1, SHACL-AF §5.2), silently corrupting evaluation. Fix: hard-error on an empty name and on the reserved set (`this`, `path`, `PATH`, `value`, `shapesGraph`, `currentShape`). Added `sparql_function_with_reserved_param_name_is_rejected`. Test passes; clippy `-D warnings` and fmt clean.

paudley:
### Gap 5 (LOW) resolved — hard-fail malformed `sh:order`/`sh:optional` and multi-projection SELECT bodies

**Commit:** `b9f9e8e`

Three fail-fast hardenings (consistent with `.goals` HARD-FAILS): a present-but-non-numeric `sh:order` and a non-boolean `sh:optional` now error instead of silently defaulting to `INFINITY`/`false` (a `"1"^^xsd:boolean` optional is now honoured); and a SELECT body projecting more than one variable is rejected at evaluation instead of silently returning column zero. Added `sparql_function_with_non_numeric_order_is_rejected` and `multi_projection_select_body_is_a_hard_error`.

paudley:
### Gap 4 (MEDIUM) resolved — negative-path coverage for `sh:SPARQLFunction`

**Commit:** `7b30eca`

**Harness note:** the conformance corpus runner treats a validation *error* as a test failure, so hard-error negative cases cannot be expressed as corpus fixtures — they are covered as unit tests.

- **Corpus fixture** `69-sparqlfn-nodekind-param` (conforming): exercises an `sh:nodeKind sh:IRI` parameter constraint end-to-end via `sh:expression` with an IRI argument. Ratchet bumped 68→69; frozen digest regenerated with `check-corpus-frozen.py --update` (vendored W3C manifests unchanged).
- **Unit tests** for the hard-error paths: `undefined_function_call_is_a_hard_error` and `parameter_datatype_violation_is_a_hard_error` (arity mismatch already covered by `wrong_arity_is_a_hard_error`).

The two HIGH bugs (Gaps 1–2) now have direct regression tests; combined with Gaps 3 and 5, every reviewer finding and negative path is covered.

paudley:
### Feedback-loop gap resolved — merge `sh:SPARQLFunction` body state back into the caller

**Commit:** `fb25f4d` (coderabbit review finding, `eval.rs`/`user_fn.rs`)

The function body was evaluated in a detached child `EvalCtx` and only the return term was interned back, so `constructed` rdf:List quads (`listSlice`/`listConcat`) were dropped and `bnode_counter`/`rng_state` reset per call — returned blanks could collide across calls and `RAND()`/`UUID()`/`STRUUID()` replayed the same stream (a determinism/data-integrity hole against `.goals`). Fix: the child now inherits the parent `bnode_counter`, and after evaluation its advanced `bnode_counter`/`rng_state` plus minted `constructed` quads are merged back into the parent (`eval_user_function` now takes `&mut EvalCtx`). Added `function_body_state_is_merged_back` (two `RAND()`-bodied calls within one expression must differ). Full `make check` green.

paudley:
## Stage 2 complete — issue #29 ready for `/stage3`

All gap-analysis items and reviewer feedback are resolved; the final branch (`fb25f4d`) passes the full `make check` (clippy `-D warnings`, all tests, hygiene, `make wasm`).

### Prerequisite: `origin/main` repaired
Commit `a7fb8ce` ("SHACL-SPARQL custom constraint components", PR #44) had been squash-merged in a **non-compiling** state and pushed to `main` — stray conflict markers, a missed `run_select`→`run_select_with_substitutions` rename, and an incomplete `Shapes`/`Parser` dataset-threading refactor. Repaired and pushed to `origin/main` as `12d28cf` (full `make check` green) so `/stage3`'s merge-of-main is unblocked.

### Gaps addressed (each committed + pushed + tested)
| Gap | Pri | Commit | Resolution |
|-----|-----|--------|------------|
| Unbound mandatory param yields no result (§5.2/§9.5) | HIGH | `e90bdd5` | `return Ok(None)` for a required unbound arg |
| `sh:returnType` is informational, not enforced (§5.3) | HIGH | `05594bc` | removed runtime datatype enforcement; retained as metadata |
| Reject empty/reserved derived param names | MED | `837232b` | hard-error on empty + `this`/`path`/`PATH`/`value`/`shapesGraph`/`currentShape` |
| Negative-path coverage | MED | `7b30eca` | conforming `nodeKind` fixture (#69, ratchet+digest) + unit tests for undefined-fn & param-datatype hard errors |
| Malformed `sh:order`/`sh:optional` + multi-projection SELECT | LOW | `b9f9e8e` | fail-fast instead of silent default; single-projection enforced |
| **Feedback loop:** merge UDF body state back into caller | MAJOR | `fb25f4d` | `bnode_counter`/`rng_state`/`constructed` merged back (determinism) |

### Reviewer feedback
All three **gemini-code-assist** findings map to Gaps 1–3 (resolved). Both **coderabbitai** findings resolved: the `shapes.rs:523` reserved-name finding = Gap 3; the `eval.rs` UDF-state finding = the merge-back commit.

### Design note (deliberate override)
Gap 2 overrides the Stage-1 plan's "enforce `sh:returnType` as a hard type check": per SHACL-AF §5.3 that annotation is informational and MAY be a class IRI, so enforcing it as a runtime datatype constraint was spec-incorrect. `.goals` HARD-FAILS means fail fast on genuine errors, not manufacture a constraint the spec calls informational.

Ready for `/stage3`.

paudley:
feat(shacl-af): SPARQL-based functions (sh:SPARQLFunction)

Implements user-declared SHACL-AF SPARQL functions: functions defined in the
shapes graph with sh:parameter / sh:returnType and a sh:select or sh:ask body,
callable from SHACL-SPARQL constraint/target queries and from sh:expression node
expressions.

## What this delivers (issue #29 acceptance criteria)

- Parse sh:SPARQLFunction declarations (parameters with sh:order/sh:optional,
  sh:returnType, sh:select/sh:ask body) into a function registry.
- A dynamic function registry threaded from the shapes engine into the SPARQL
  evaluator via a callback seam (EvalCtx.user_functions / query_with_user_functions),
  not a hardcoded set — the open counterpart of the closed native PurrdfFn set.
- Parameter binding + return-value extraction with arity and per-parameter type
  checks.
- Reachable from sh:sparql / sh:SPARQLTarget queries (run_select_with_substitutions
  routes through the user-function path when a registry is in scope) AND from
  sh:expression node expressions (FnCall::UserDefined resolves via the registry).
- Conformance coverage as first-party example.org fixtures (65-69) — the pinned
  W3C corpus ships no AF `function` manifests.

## Key files

- crates/sparql-eval/src/user_fn.rs (new): UserFunction / UserFunctionRegistry
  model and eval_user_function (arity check, parameter binding, ASK/SELECT body
  execution, recursion bound MAX_UDF_DEPTH, state merge-back).
- crates/sparql-eval/src/{eval,expr,engine,error,lib}.rs: EvalCtx.user_functions
  seam, child_for_user_fn recursion-bounded context, Function::Custom resolution.
- crates/shapes/src/shapes.rs: parse_one_sparql_function, Shapes.functions
  registry, local_name / node_kind_from_iri helpers.
- crates/shapes/src/sparql.rs: run_select_with_substitutions routes through the
  registry when in scope (CURRENT_FUNCTIONS thread-local guard).
- crates/shapes/corpus/65..69-sparqlfn-*: conformance fixtures; conformance.rs
  ratchet + scripts/conformance-frozen/shapes-corpus.sha256 regenerated.

## Spec-correctness decisions (Stage-2 gap analysis)

- Unbound mandatory parameter -> no result node (SHACL-AF §5.2/§9.5).
- sh:returnType is INFORMATIONAL (§5.3): may be a class IRI, so it is retained as
  metadata but NOT enforced as a runtime datatype constraint (enforcing it wrongly
  rejected IRI/blank-node returns).
- Reject empty / SHACL-reserved derived parameter names (this, path, PATH, value,
  shapesGraph, currentShape).
- Hard-fail malformed sh:order / sh:optional and multi-projection SELECT bodies.
- Function-body minted state (bnode_counter, rng_state, constructed rdf:List
  quads) is merged back into the caller so blanks stay unique across calls and
  RAND()/UUID()/STRUUID() advance rather than replay (determinism).

- RUST-FIRST: all functionality in purrdf-shapes / purrdf-sparql-eval; no Python.
- NO OPTIONALITY / HARD FAILS: undefined function, arity mismatch, parameter type
  violation, and malformed declarations are hard errors.
- MAXIMAL PORTABILITY: no new deps, no oxigraph/PyO3; make wasm clean.
- No Cargo features; determinism preserved (DetHashMap, sorted registry, merged
  RNG/bnode state); vendored corpora byte-frozen (digest regenerated, not
  hand-edited); fixtures use example.org.

## Merge conflicts resolved (main <- this branch)

main's SHACL-SPARQL custom-constraint-components feature and this branch both
extended the same files. Resolved keeping BOTH features:
- model.rs: union of the sh:SPARQLFunction vocab (RETURN_TYPE, PREDICATE) and the
  component vocab; sh:parameter property is sh::PARAMETER_PROPERTY, sh:Parameter
  class is sh::PARAMETER (this branch's sh::PARAMETER property usage rebound to
  sh::PARAMETER_PROPERTY).
- sparql.rs: kept this branch's user-function-aware body under main's
  run_select_with_substitutions name; renamed callers.
- shapes.rs: unioned imports, Shapes fields (functions + shapes_graph +
  shapes_dataset) and construction; severity_from_term now sourced from
  crate::components (main), keeping local_name / node_kind_from_iri.

Full `make check` (fmt, clippy -D warnings, tests, hygiene, wasm) green on the
merged result.

Issue #37, opened 2026-07-03
feat(entail): native OWL-DL reasoner + RIF engine for full entailment-regime conformance


paudley:
## Context

The W3C conformance gate (#9) landed `purrdf-entail`: a native, wasm-clean **forward-materialization** reasoner over RDFS + OWL-RL. It materializes a fixpoint closure and matches BGPs against it — the correct tool for the RDF/RDFS/OWL-RL/D entailment regimes, which it passes.

It is structurally the **wrong tool** for three entailment regimes the W3C SPARQL 1.1 entailment suite exercises, so **31 vendored W3C entailment cases remain typed expected-failures** in `crates/sparql-conformance/src/xfail.rs` (`XfailReason::Entailment`). Closing them requires reasoning engines a materialization chase cannot emulate — open-world class-expression retrieval and rule-language entailment.

This is deliberately deferred from #9 (which scoped the materialization reasoner only) and tracked here. It is a large, multi-part engine build.

### 1. Native OWL-DL reasoner (OWL Direct Semantics)
Open-world class-expression reasoning and instance retrieval — `owl:intersectionOf`, `owl:unionOf`, `owl:someValuesFrom`, `owl:allValuesFrom`, qualified/number restrictions — including the case where the class expression appears **only in the query** and is not asserted in the data (so materializing the data yields no membership to match). This is tableau- or consequence-based DL reasoning, not materialize-and-match.

Closes (26):
- `parent3`–`parent10` (8)
- `simple1`–`simple8` (8) — these declare `sd:entailmentRegime ent:OWL-Direct` / `pr:DL`; "simple" denotes small class-expression queries, **not** the simple-entailment regime
- `sparqldl-02/03/10/11/12/13` (6)
- `paper-sparqldl-Q1/Q2/Q3` (3)
- `owlds02` (1)

### 2. RIF-rule entailment engine
The RIF regime (RIF-Core / BLD rules) is a separate rule language, out of reach of the RDFS/OWL-RL rule set.

Closes (4): `rif01`, `rif03`, `rif04`, `rif06`

### 3. RDF axiomatic-triple entailment (bare RDF regime)
The infinite RDF axiomatic-triple schema under the `ent:RDF` regime.

Closes (1): `rdf01`

## Constraints (project mandate)
- **SUBSUME/EXTEND/ENHANCE**: subsume a suitable algorithm, do not take a heavyweight runtime dependency; relicense-clean (MIT/Apache), user owns copyright.
[...]

Issue #10, opened 2026-07-02
EPIC — purrdf public maturity: full conformance · optimization · ecosystem


paudley:
# EPIC — purrdf public maturity: conformance, optimization, ecosystem

The extraction and first publishes are done (crates.io 0.1.3, PyPI `purrdf`, npm `@blackcatinformatics/purrdf`, CI with fmt / clippy -D warnings / tests / MSRV 1.96 / wasm). This epic is the explicit home for the work that makes purrdf a *mature public* RDF stack — tracked as first-class issues, not buried deferral notes.

## Children

- [ ] #9 — Full W3C conformance (SPARQL 1.1/1.2 · RDFC-1.0 · IRI · syntax) + rdflib test-suite gate
- [ ] #6 — Vendor the full W3C SPARQL syntax suite as `purrdf-sparql-algebra` fixtures
- [ ] #8 — Vendored W3C SHACL + shexTest suites as engine conformance gates
- [ ] #7 — rdflib plugin entry-point discovery in `purrdf.compat.rdflib`
- [ ] #30 — SHACL-AF completion (validation scope)
- [ ] SPARQL engine completion: `SERVICE`, deep-subquery optimization, exotic aggregation (file when picked up)
- [ ] Cost-based query planning from cardinality statistics, measurably beating the selectivity heuristic on a criterion suite that exercises it (file when picked up)
- [ ] Published benchmarks vs rdflib (and peers) from the criterion suites

## Definition of done

All conformance children green in a single CI-published matrix; the planner win demonstrated by bench, not asserted; plugin discovery resolves the third-party acceptance matrix; benchmarks published.

---
*Migrated from [gmeow-ontology#927](https://github.com/Blackcat-Informatics/gmeow-ontology/issues/927); rewritten for the purrdf extraction.*

paudley:
# Implementation Plan for Issue #10

## Issue Summary

Issue #10 is the public-maturity epic for PurRDF: full W3C conformance, vendor
suites as hard gates, rdflib ecosystem integration, SHACL-AF completion, SPARQL
engine completion, cost-based query planning, and published benchmarks.

Reality on the ground: **all five filed child issues are already merged into
`main`** via prior PRs:

| Child | PR | Status |
|-------|-----|--------|
| #6 — Vendor full W3C SPARQL syntax suite | #39 | merged |
| #7 — rdflib plugin entry-point discovery | #42 | merged |
[...]

PR #52, opened 2026-07-04
feat(validate): SARIF 2.1.0 source-traced reporting

### LINKED ISSUE #48: SARIF 2.1.0 reporting: source-traced, actionable diagnostics and validation reports (Task 12)
## Goal

Uplift PurRDF's reporting layer to emit **SARIF 2.1.0** for errors, warnings, and SHACL/ShEx validation results — with faithful source tracing back to the input document (file, line, column, span) and rich, actionable messages. The groundwork was designed for exactly this: `RdfLocation` is documented as the anchor "that threads into a SARIF `physicalLocation` (Task 12)" (`crates/rdf-core/src/diagnostic.rs:59-62`), `RdfSeverity` mirrors LSP severities, and `AttributionRole` already reserves stable string ids "used in SARIF" (`crates/rdf-core/src/provenance.rs:184,209`). This issue is Task 12 made concrete.

## Current state (grounded)

- **Central diagnostic exists but is under-populated.** `RdfDiagnostic` + `RdfLocation { path, line, column, logical, gts_* }` live in `crates/rdf-core/src/diagnostic.rs:39,186`. The struct can carry `path:line:column`, but almost nothing fills it in.
- **Parsers report byte offsets or raw line text, not positions.**
  - SPARQL: `ParseError::Lex/Syntax { reason, at: usize }` (`crates/sparql-algebra/src/error.rs:24`) — `at` is a byte offset, rendered as "at byte {at}".
  - ShEx: same shape (`crates/shex/src/error.rs:15`).
  - IRI: `IriError::BadPercentEncoding(usize)`, `DisallowedChar(char, usize)` (`crates/iri/src/error.rs:15`).
  - Turtle/TriG/N-Triples/N-Quads: `text_parse.rs` errors embed the offending **line text** into the message string (`crates/rdf/src/native_codecs/text_parse.rs:251,284`) and never attach an `RdfLocation` with a line *number*.
- **SHACL results have no source location.** `ValidationResult` (`crates/shapes/src/report.rs:71`) points at RDF terms (`focus_node`, `source_shape`, `result_path`, `value`) plus slice `attributions` — but no file/line/column linking back to where the offending triple or shape was parsed from.
- **No SARIF, no serde on report types, no CLI.** Reports serialize only to the RDF report graph (`ValidationReport::to_ntriples()`, `report.rs:178`) and an ad-hoc Python dict (`bindings/python/src/shacl.rs:39-100`). `serde::Serialize` is not derived on `RdfDiagnostic`/`RdfLocation`/`ValidationReport`/`ValidationResult`. No `miette`/`codespan`/`ariadne` anywhere. The `purrdf-validate` boundary named in `provenance.rs` docs does not exist yet.

### Phase 1 — position fidelity (the real work)

1. **Line/column tracking in the text codecs.** Thread a position index (byte offset → line/col; a newline table built during lexing is enough and cache-friendly) through `native_codecs/text_parse.rs` lexer/parser so every parse error constructs `RdfLocation::file(path).with_line(l).with_column(c)` instead of interpolating line text into the message. Same for the SPARQL (`sparql-algebra`) and ShExC lexers: keep `at: usize` (byte-precise, useful for spans) and add resolved line/col + span length.
2. **Optional triple→source provenance map.** Parsers gain an opt-in side-table mapping each parsed triple/quad (by index or TermId tuple) to its source span. This is what lets a SHACL violation on `ex:Alice` point at *line 14 of data.ttl* rather than just the focus node IRI. Must be strictly opt-in (zero cost when off) to respect the perf discipline; datasets assembled programmatically simply have no physical location and fall back to logical locations.
3. **Logical locations for RDF-space results.** When no physical source exists, populate SARIF `logicalLocations` from what we uniquely have: focus node, result path (incl. `path_structure` rendered as SPARQL path syntax), source shape, constraint component, and slice `Attribution` roles (their stable ids were reserved for this).

### Phase 2 — SARIF emission

4. **New boundary crate `purrdf-validate`** (name already anticipated by `provenance.rs` docs) hosting the SARIF writer. Keeps `purrdf-core` "structured but SARIF-free" as its `lib.rs:11` promises, and keeps the kernel ring-fence intact. Hand-rolled serde structs for the SARIF subset we emit (runs/tool/rules/results/locations/relatedLocations/fixes) — no heavyweight SARIF crate needed; `serde`/`serde_json` are already workspace deps. Must stay `wasm32-unknown-unknown`-clean like every release crate.
5. **Mappings** (documented in the crate):
   - `RdfSeverity::{Error,Warning,Note,Info}` → SARIF `level` {error, warning, note, note}.
   - SHACL `Severity::{Violation,Warning,Info,Other(iri)}` → {error, warning, note, + `properties` carrying the verbatim severity IRI — SHACL severities are open-world and must not be lossy}.
   - `RdfDiagnostic.code` → SARIF `ruleId`; constraint-component IRIs → stable rule ids with `helpUri` pointing at the W3C SHACL spec anchors.
   - `RdfLocation` → `physicalLocation` (artifact URI + region with startLine/startColumn/byteOffset); GTS indices (`gts_quad_index` etc.) and attributions → `logicalLocations` + `properties`.
   - `ValidationResult.message` → `message.text`; when absent, synthesize an actionable message from component + path + value ("Value \"foo\"^^xsd:string on path ex:age fails sh:datatype xsd:integer at shape ex:PersonShape") — never a bare IRI dump.
   - Multi-location results: focus-node span as the primary location, source-shape span as a `relatedLocation` ("shape defined here").
6. **Determinism.** SARIF output must be byte-deterministic like every serializer here: stable result ordering (severity, then location, then rule id), no timestamps unless caller-supplied, `serde_json` with stable key order. Goldens under the usual golden discipline.
7. **Bindings surface.** Expose `to_sarif()` alongside `to_ntriples()` on `ValidationReport`, and thread through the Python/WASM/C-ABI bindings (the Python dict builder in `bindings/python/src/shacl.rs` gets a `sarif` sibling).

### Phase 3 — message quality pass

8. Audit `Display` texts across the error enums for actionability: every message should say *what* was expected, *what* was found, and *where* — e.g. `ShexError::Syntax` and Turtle `err()` messages that today read "unexpected token in {line:?}". Add SARIF rule metadata (`shortDescription`/`fullDescription`/`help`) per rule id.

## Non-goals

- No terminal pretty-rendering (ariadne/miette-style carets) — that's a separate, optional consumer of the same `RdfLocation` data.
- No changes to the RDF report-graph serialization (`to_ntriples()` stays canonical per the SHACL spec).
- SARIF `fixes` (auto-fix suggestions) are a follow-up once emission lands.

## Gates

- `make check` + `make wasm` clean (new crate included in the wasm sweep).
- W3C SHACL/ShEx/SPARQL conformance counts untouched (position tracking must not change accept/reject behavior).
- SARIF goldens validated against the SARIF 2.1.0 JSON schema in CI; byte-deterministic across runs.
- Zero measurable parse-throughput regression with provenance tracking **off** (existing `native_codecs` bench as the report-only reference, per repo perf discipline).
- No new deps beyond serde/serde_json already in the workspace tree; no Cargo features (the opt-in is a runtime option struct, not a feature flag).

paudley:
Closes #48

## Summary
Adds source-traced, actionable **SARIF 2.1.0** output for parse diagnostics and SHACL/ShEx validation reports, backed by a shared lazy source-position primitive and a new wasm-clean boundary crate `purrdf-validate`.

## Changes
- **`iri::pos`** — `LineIndex`/`Position` primitive (byte offset → 1-based line/column), built **lazily on the error path only**, hosted in the zero-dependency `purrdf-iri` leaf so every parser shares it (zero cost on the happy path).
- **Position fidelity** — `ParseError`/`ShexError`/`IriError` gain `byte_offset()`/`locate()`; the native text codecs (N-Triples/N-Quads/Turtle/TriG) attach real line/column `RdfLocation`s and no longer dump line text (parallel path keeps document-global line numbers; determinism proofs unchanged).
- **Opt-in span table** — a runtime `ParseOptions.track_source_spans` (never a Cargo feature) with a const-`ENABLED` ZST collector: **zero cost when off** (`parse_dataset` byte-identical). `parse_dataset_with` returns `(dataset, Option<SpanTable>)`.
- **`purrdf-validate`** — new wasm-clean boundary crate: hand-rolled **byte-deterministic** SARIF 2.1.0 serde model; severity/ruleId/message-synthesis mappings; physical (focus→span) + logical (focus/path/shape/component/attribution) + related (`shape defined here`) locations; `path_structure` rendered as SPARQL property-path syntax; provenance `UnitId` resolved to public slice IRIs at the boundary (S0.5).
- **`to_sarif()` across all bindings** — a `SarifReport` extension trait (Rust), `ValidationReport.to_sarif()` (Python), `shaclValidateToSarif` (WASM), `purrdf_shacl_validate_to_sarif` (C-ABI).
- **Schema validation** — vendored OASIS SARIF 2.1.0 schema + a `jsonschema` pytest lane (verified: real `to_sarif()` output validates), plus a dependency-free Rust structural check in the `make check` gate.
- **Rule metadata** — per-rule `reportingDescriptor` with W3C SHACL spec `helpUri` anchors.

## Gates
- `make check` (fmt + clippy pedantic/nursery `-D warnings` + workspace build + all tests + hygiene + ring-fence) ✅
- `make wasm` (incl. `-p purrdf-validate`) ✅
- W3C SHACL/ShEx/SPARQL conformance counts unchanged (position tracking is additive) ✅
- SARIF byte-deterministic; goldens + serialize-twice tests ✅
- No new Rust deps beyond serde/serde_json; no Cargo features ✅

## Summary by CodeRabbit

* **New Features**
  * Added SARIF 2.1.0 SHACL validation output across Python, WebAssembly, and C interfaces, including `ValidationReport.to_sarif()` in Python.
  * Introduced `purrdf-validate` SARIF rendering (including SHACL path-to-SPARQL rendering) and rule metadata generation.
  * Added opt-in source span tracking via `parse_dataset_with` + `ParseOptions` (`SpanTable`) for physical source locations.
* **Bug Fixes**
  * Improved diagnostics with `byte_offset()`/`locate()` helpers returning consistent 1-based line/column positions.
* **Tests**
  * Added SARIF schema-validated end-to-end tests and determinism checks, plus expanded location helper coverage.


paudley:
# Implementation Plan for Issue #48 — SARIF 2.1.0 reporting

## Context

PurRDF's reporting layer today points at *RDF terms* (focus node, source shape)
but has no faithful path back to the *input document* (file, line, column, span),
and it emits only an RDF report graph or an ad-hoc Python dict — no SARIF, no
serde on the report types. The groundwork was deliberately pre-staged for this:
`RdfLocation` is documented as the anchor "that threads into a SARIF
`physicalLocation` (Task 12)" (`crates/rdf-core/src/diagnostic.rs:59-62`),
`RdfSeverity` mirrors LSP levels, and `AttributionRole` reserves stable string
ids "used in SARIF" (`crates/rdf-core/src/provenance.rs:184,209`). This issue is
Task 12 made concrete: emit **SARIF 2.1.0** for errors/warnings and SHACL/ShEx
validation results, source-traced and actionable.

Outcome: a new wasm-clean boundary crate `purrdf-validate` that renders any
`ValidationReport` (and parse `RdfDiagnostic`s) to byte-deterministic SARIF
2.1.0, backed by a shared source-position primitive that finally populates the
line/column fields the parsers have always carried but never filled.

- Adversarial gap analysis; assume code is substandard until proven otherwise.
- **NO deferrals / follow-ups / future-work.** Every requirement is done NOW.
- **No issue/PR numbers in repo source, docs, or ontology.** Process flow lives
  in GitHub, never in the tree. No "Per #48…" comments.
- Own every bug you touch; fix it now, don't trace blame.

- GREENFIELD-FIRST / SOTA / no compromises.
- RUST-FIRST, PYTHON-SURFACE: core in Rust; bindings are a thin surface.
- SUBSUME, EXTEND, ENHANCE substandard building blocks.
- LOW/NO OPTIONALITY, HARD FAILS: no conditional imports, no swallowed errors.
  (The one sanctioned exception is the perf opt-in required by the issue — and
  it is a **runtime option struct, never a Cargo feature**, honoring both this
  and the repo's no-features rule.)
- MAXIMAL UTILITY / PERFORMANCE / PORTABILITY (wasm-clean core).

Plus CLAUDE.md/AGENTS.md hard rules: no Cargo features; never hand-edit
`generated/`/`vectors/`; root `[workspace.dependencies]` only; determinism in
serializers; no oxigraph/PyO3 near `purrdf-core`; every release crate stays
`wasm32-unknown-unknown`-clean (`make wasm`); brand casing **PurRDF**/`purrdf`;
PurRDF mints no vocabulary IRIs (fixtures use `example.org`).

## Completeness Contract

Every requirement and acceptance criterion from the issue maps to a numbered

| Req 1 — line/col tracking in text codecs (Turtle/TriG/NT/NQ) | 1, 3 |
| Req 1 — line/col + span for SPARQL & ShExC lexers (keep byte `at`) | 1, 2 |
| Req 1 — IRI error offsets → line/col | 1, 2 |
| Req 2 — opt-in triple→source span map, zero-cost when off | 4 |
| Req 3 — logical locations (focus/path/shape/component/attribution) | 8 |
| Req 3 — `path_structure` rendered as SPARQL property-path syntax | 8 |
| Req 4 — new boundary crate `purrdf-validate` hosting SARIF writer | 5 |
| Req 4 — hand-rolled serde SARIF subset; no heavyweight SARIF crate | 6 |
| Req 4 — new crate stays wasm32-clean (added to `make wasm`) | 5 |
| Req 5 — RdfSeverity→level mapping | 7 |
| Req 5 — SHACL Severity→level, `Other(iri)` non-lossy via properties | 7 |
| Req 5 — code→ruleId, constraint-component→stable rule id + helpUri | 7, 11 |
| Req 5 — RdfLocation→physicalLocation (uri + region + byteOffset) | 8 |
| Req 5 — GTS indices + attributions → logicalLocations + properties | 8 |
| Req 5 — message→message.text; synthesize actionable msg when absent | 7 |
| Req 5 — multi-location: focus primary, source-shape relatedLocation | 8 |
| Req 6 — determinism (result ordering, no timestamps, stable keys) | 6, 7 |
| Req 6 — SARIF goldens under golden discipline | 6, 7, 8 |
| Req 7 — `to_sarif()` alongside `to_ntriples()` (Rust surface) | 9 |
| Req 7 — Python binding `sarif` sibling | 9 |
| Req 7 — WASM binding SARIF surface | 10 |
| Req 7 — C-ABI binding SARIF surface | 10 |
| Req 8 — audit `Display` texts for actionability (what/where) | 11 |
| Req 8 — SARIF rule metadata (short/full description, help) per rule | 11 |
| Gate — `make check` + `make wasm` clean | every task; 12 |
| Gate — W3C SHACL/ShEx/SPARQL conformance counts untouched | 2, 3, 12 |
| Gate — SARIF goldens validated vs SARIF 2.1.0 JSON schema in CI | 9 |
| Gate — byte-deterministic SARIF across runs | 6 |
| Gate — zero parse-throughput regression with tracking OFF | 4 |
| Gate — no new Rust deps beyond serde/serde_json; no Cargo features | 5, 6 |

**Non-goals (explicit scoping by the issue author — NOT deferrals):** terminal
pretty-rendering (ariadne/miette carets); changes to `to_ntriples()` RDF
report-graph serialization; SARIF `fixes`/auto-fix suggestions. These are the
issue's own boundaries and are honored as written.

- **TRANSFORMATION — ADOPTED (restructures the plan):** rather than three ad-hoc
  position bolt-ons, build **one shared `LineIndex`/`Position` primitive** in
  `purrdf-iri` (the zero-dep leaf every parser already reaches), built **lazily
  on the error path only**. This subsumes the text-codec, SPARQL, ShExC, and IRI
  needs into a single tested primitive with a zero-cost happy path — directly
  serving `.goals` SUBSUME/EXTEND and the "zero measurable regression" gate.
  Tasks 1–3 are organized around it. *Why the primitive and not per-lexer
  counters:* threading a live line counter through hot lexer loops risks the
  parse-throughput gate; resolving byte-offset→line/col only when a diagnostic is
  already being constructed makes position fidelity free unless parsing fails.
- **ADOPTED — extension trait** `SarifReport for ValidationReport` in
  `purrdf-validate` so `report.to_sarif(&opts)` reads as a method (req 7) while
  the writer never leaves the boundary crate (req 4); `purrdf-shapes` stays
  SARIF/serde-free.
- **ADOPTED — ZST `SpanCollector`** (`NoSpans` monomorphized to nothing) so the
  opt-in span map adds no branch to the hot parse loop when off.
- **Declined — `purrdf-span` as its own new leaf crate:** would force a new
  dependency edge onto `iri`/`xsd`/`sparql-algebra` and trip the zero-dep-leaf
  gate. Hosting `pos` inside `iri` adds zero edges. Lesser structure suffices.
- **Declined — vendoring a Rust `jsonschema` crate** for schema validation:
  a new Rust workspace dep, violating the "no new deps" gate. Schema validation
  runs in the Python CI lane (`jsonschema` is a pip dep, not a Rust dep).
- **Declined — deriving `Serialize` on `purrdf-core`/`purrdf-shapes` report
  types:** would pull serde-derive/SARIF concerns down a ring and break the
  `purrdf-core` "SARIF-free" promise. The SARIF structs are defined and owned
  entirely in `purrdf-validate`.

1. Resolve the top-level repo and define paths:
   ```bash
   BRANCH="paudley/48-sarif-reporting"
   WORKTREE="$REPO_ROOT/.worktrees/48-sarif-reporting"
   ```
   ```bash
   ```
   ```bash
   cd "$WORKTREE"
   git branch --show-current   # MUST show paudley/48-sarif-reporting, NOT main
   ```
4. Re-establish a green baseline in the fresh worktree before implementing:
   `make check` (and `make wasm` if the wasm target is installed). Confirm green.
   ```bash
   gh issue comment 48 --body-file /home/paudley/.claude/plans/48-zippy-locket.md
   ```

**⚠️ DO NOT proceed to Task 1 until the worktree exists, you are on the feature
branch, and the baseline is green.**

## Task 1: Shared source-position primitive `iri::pos` (TRANSFORMATION)

New module `crates/iri/src/pos.rs` (`pub mod pos;` in `crates/iri/src/lib.rs`),
SPDX header, zero new deps (keeps `purrdf-iri` a zero-dep leaf — the Makefile
zero-dep-leaf gate must stay green):

```rust
/// 1-based line/column plus the originating byte offset (for SARIF Region.byteOffset).
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Default)]
pub struct Position { pub line: u32, pub column: u32, pub byte_offset: usize }

/// Newline table over one source document. Built ONCE, O(n), on the ERROR path.
#[derive(Clone, Debug)]
pub struct LineIndex { line_starts: Vec<usize> } // [0] == 0

impl LineIndex {
    pub fn new(src: &str) -> Self;                       // single byte scan, no memchr
    pub fn locate(&self, src: &str, byte_offset: usize) -> Position; // binary search
    #[inline] pub fn line_col(&self, src: &str, byte_offset: usize) -> (u32, u32);
}
```

- Columns count Unicode scalar values (code points) from line start — SARIF
  `region` columns are code-point based. Clamp out-of-range offsets to EOF; never
  panic.
- Unit + property tests: `locate` is monotonic in byte offset; column counts
  scalar values (multi-byte UTF-8 fixture); offset past EOF clamps.
- No other crate changes here; `pos` is standalone.

### After Task 1 Implementation:
- Commit: github-commit-agent. Push: github-push-agent. Post update to issue #48.

## Task 2: SPARQL / ShExC / IRI position resolution

At the diagnostic-construction seams (where the full source text + path are in
hand), resolve the existing byte `at`/offset into line/col via
`iri::pos::LineIndex`. The lexer enums KEEP `at: usize` (byte-precise, needed for
span length) — this is purely additive.

- `crates/sparql-algebra`: resolve `ParseError::{Lex,Syntax}{at}` → line/col at
  the boundary that owns the query string; carry span length where the token end
  is known. Improve `Display` to include line:col.
- `crates/shex`: same for `ShexError::{Lex,Syntax}{at}`.
- `crates/iri`: `IriError::{BadPercentEncoding(at),DisallowedChar(_,at)}` gain a
  resolution helper so callers that own the lexical string can attach line/col.
- **Conformance gate:** accept/reject behavior is unchanged — only message/location
  content changes. Run the SPARQL/ShEx conformance suites; counts must match.
- Goldens/insta for a representative error message per lexer.

### After Task 2 Implementation:
- Commit: github-commit-agent. Push: github-push-agent. Post update to issue #48.

## Task 3: Text-codec position fidelity (`text_parse.rs`)

Rewrite the error sites in `crates/rdf/src/native_codecs/text_parse.rs` to attach
a real `RdfLocation::file(path).with_line(l).with_column(c)` instead of
interpolating offending line **text** into the message (lines ~251,264,284 and the
`validate_*`/`TokenCursor`/`DocParser` error sites):

- NT/NQ line loop (`parse_lines_sequential`, `text.lines()` at ~246): accumulate a
  running document byte offset per line so per-line-local token `Spanned{start,end}`
  become document-global; resolve via `LineIndex` on error.
- Turtle/TriG `DocParser`: `Spanned{start,end}` are already document-global — feed
  directly to `LineIndex` at each error site.
- Build the `LineIndex` lazily (only when an error is being constructed).
- The `err()` helper gains a location-carrying variant; messages state what was
  expected/found (no raw line text dumps — this also front-loads part of Req 8).
- Goldens for the new diagnostic messages+locations.

### After Task 3 Implementation:
- Commit: github-commit-agent. Push: github-push-agent. Post update to issue #48.

## Task 4: Opt-in triple→source span side-table (perf-critical)

In `crates/rdf` (native_codecs), add a monomorphized, branch-free collector:

```rust
pub trait SpanCollector { fn record(&mut self, quad_index: usize, span: iri::pos::Position); }
pub struct NoSpans;      // ZST
impl SpanCollector for NoSpans { #[inline(always)] fn record(&mut self,_:usize,_:_){} }
pub struct SpanTable { spans: Vec<iri::pos::Position> } // dense; index == quad emission index
impl SpanCollector for SpanTable { #[inline] fn record(&mut self, i, s){ debug_assert_eq!(i,self.spans.len()); self.spans.push(s); } }

#[derive(Default, Clone, Copy)]
pub struct ParseOptions { /* existing base/etc + */ pub track_source_spans: bool }
```

- The public parse entry checks `track_source_spans` **once**, then dispatches to a
  generic inner parse over `S: SpanCollector`. With `NoSpans`, `record` compiles to
  nothing — hot loop byte-identical to today. Runtime option struct, **not** a
  Cargo feature.
- Join key is the **quad emission index (`QuadHandle`)** — the same key the
  existing provenance sidecar (`AssertionOccurrence`) uses. Programmatically
  assembled datasets simply have no `SpanTable` (→ logical-location fallback).
- **Gate:** extend/rerun the `crates/rdf/benches/native_codecs.rs` criterion bench
  (report-only) and reason from code that the `NoSpans` path is unchanged; assert
  no branch added to the hot loop. Per repo perf discipline, do NOT assert a
  speedup — the bench is the report-only reference that tracking-off is unchanged.

### After Task 4 Implementation:
- Commit: github-commit-agent. Push: github-push-agent. Post update to issue #48.

## Task 5: Scaffold `purrdf-validate` boundary crate

- `crates/validate/` → package `purrdf-validate`. Add to root `Cargo.toml`
  `[workspace] members` and `[workspace.dependencies]`
  (`purrdf-validate = { path = "crates/validate", version = "0.2.1" }`).
- Inherit `[workspace.package]` + `[lints] workspace = true`; SPDX header on every
  file. Deps (workspace-pinned only): `purrdf-shapes`, `purrdf` (=purrdf-rdf, for
  `RdfDiagnostic`/`RdfLocation`/provenance interners), `purrdf-iri`, `serde`,
  `serde_json`. No PyO3, no oxigraph, no I/O → wasm-clean.
- Add `-p purrdf-validate` to the `make wasm` `-p` enumeration (Makefile ~L71).
- Empty `lib.rs` that compiles clean under pedantic/nursery `-D warnings` and in
  the wasm sweep.

### After Task 5 Implementation:
- Commit: github-commit-agent. Push: github-push-agent. Post update to issue #48.

## Task 6: SARIF data model + determinism (`model.rs`)

Hand-rolled `#[derive(Serialize)]` structs in `crates/validate/src/model.rs`,
`#[serde(rename_all="camelCase")]`, `skip_serializing_if` on Options/empty Vecs:
`SarifLog{version:"2.1.0",$schema,runs}`, `Run{tool,results,artifacts?,invocations?}`,
`Tool{driver}`, `Driver{name:"purrdf",version?,informationUri?,rules}`,
`ReportingDescriptor{id,name?,shortDescription?,fullDescription?,help?,helpUri?,
defaultConfiguration{level}}`, `SarifResult{ruleId,ruleIndex,level,message,
locations,relatedLocations,properties}`, `Message{text,id?,arguments?}`,
`Location{physicalLocation?,logicalLocations,message?}`,
`PhysicalLocation{artifactLocation,region}`, `ArtifactLocation{uri,uriBaseId?}`,
`Region{startLine,startColumn?,endColumn?,byteOffset?,byteLength?}`,
`LogicalLocation{name,fullyQualifiedName?,kind}`,
`PropertyBag(BTreeMap<String,serde_json::Value>)`.

Determinism (Req 6, hard gate): field order = declaration order; property bags are
`BTreeMap` (sorted keys); **no** serde_json `preserve_order`; no timestamps unless
caller-supplied. Test: serialize a hard-coded exemplar twice → **byte-identical**;
byte golden of the exemplar.

### After Task 6 Implementation:
- Commit: github-commit-agent. Push: github-push-agent. Post update to issue #48.

## Task 7: `build_sarif` core mapping (diagnostics + reports)

`crates/validate/src/lib.rs`: `SarifOptions{tool_version, invocation_times,
source_root_uri}`, `build_sarif(&ValidationReport,&SarifOptions)->model::SarifLog`,
`to_sarif_string(...)->String`, and a diagnostics entry accepting `&[RdfDiagnostic]`
(parse errors/warnings). Mappings (Req 5):

- `RdfSeverity::{Error,Warning,Note,Info}` → level `{error,warning,note,note}`.
- SHACL `Severity::{Violation,Warning,Info}` → `{error,warning,note}`;
  `Other(iri)` → `note` **plus** `properties["shaclSeverity"]=<verbatim iri>`
  (open-world, non-lossy).
- `RdfDiagnostic.code` → `ruleId`; constraint-component IRIs → stable rule ids
  (rule metadata + helpUri finalized in Task 11).
- `ValidationResult.message` → `message.text`; when absent, synthesize an
  actionable message from component + path + value + shape (e.g. `Value
  "foo"^^xsd:string on path ex:age fails sh:datatype xsd:integer at shape
  ex:PersonShape`) — **never a bare IRI dump**.
- Result sort key applied before serialize: `(level_rank, artifact_uri, start_line,
  start_column, rule_id, message_text)` — subsumes "severity → location → ruleId".
- Goldens over synthetic `ValidationReport`s (fixtures use `example.org`).

### After Task 7 Implementation:
- Commit: github-commit-agent. Push: github-push-agent. Post update to issue #48.

## Task 8: Physical + logical + related locations

- **Physical:** join `ValidationResult.focus_node` → `SpanTable` (min `QuadHandle`
  with the focus as subject; prefer the value-bearing quad for `sh:value` results)
  → `RdfLocation` → `physicalLocation{artifactLocation.uri, region{startLine,
  startColumn, byteOffset}}`. No span table → skip physical, emit logical only.
- **Logical (Req 3):** `logicalLocations` from focus node, result path, source
  shape, constraint component, GTS indices, and `Attribution` roles. Attribution
  `kind = AttributionRole::as_str()` (reserved stable ids); `name` = slice IRI
  resolved via `UnitInterner::name()` — **numeric `UnitId` MUST NOT appear**
  (S0.5). GTS indices + attributions also mirrored into `properties`.
- **`path_structure` → SPARQL property-path syntax** (`crates/validate/src/path_syntax.rs`
  over `crate::shapes::Path`): `Predicate=>iri`, `Inverse=>"^"+inner`,
  `Sequence=>join("/")`, `Alternative=>join("|")`, `ZeroOrMore=>"*"`,
  `OneOrMore=>"+"`, `ZeroOrOne=>"?"`, parenthesizing composites. Reuses the same
  `Path` enum `emit_path_structure` walks (SUBSUME, don't re-model).
- **Related (Req 5):** focus-node span = `locations[0]`; source-shape span =
  `relatedLocations` entry with `message.text = "shape defined here"`.
- Goldens covering physical, logical-only, and multi-location results.

### After Task 8 Implementation:
- Commit: github-commit-agent. Push: github-push-agent. Post update to issue #48.

## Task 9: Rust `to_sarif()` surface + Python binding + schema validation

- **Rust surface (Req 7):** extension trait in `purrdf-validate`:
  `pub trait SarifReport { fn to_sarif(&self,&SarifOptions)->String; }` +
  `impl SarifReport for ValidationReport`. `report.to_sarif(&opts)` compiles once
  `purrdf_validate::SarifReport` is in scope — method ergonomics, writer stays in
  the boundary crate, `purrdf-shapes` untouched. Rust doc example.
- **Python (Req 7):** `bindings/python` gains `purrdf-validate` dep;
  `PyValidationReport` (`bindings/python/src/shacl.rs`) gets a `to_sarif(&self)->String`
  sibling next to `to_ntriples`, delegating detached (`py.detach`) to
  `purrdf_validate::to_sarif_string`. Add a `sarif` key to the dict builder where
  natural.
- **Schema validation (Gate):** vendor the official SARIF 2.1.0 JSON schema under
  `crates/validate/tests/data/sarif-schema-2.1.0.json`; a Python test in
  `bindings/python/tests/` generates SARIF via `to_sarif()` and runs
  `jsonschema.validate(instance, schema)`. (`jsonschema` is a pip test dep, not a
  Rust workspace dep — honors "no new Rust deps".)

### After Task 9 Implementation:
- Commit: github-commit-agent. Push: github-push-agent. Post update to issue #48.

## Task 10: WASM + C-ABI SARIF surfaces (full binding parity)

Per approved scope (all three bindings). Since WASM/C-ABI have no
`ValidationReport` surface today, add a minimal validate-and-emit entry point to
each (the SHACL engine already exists in `purrdf-shapes`):

- **WASM** (`crates/rdf-wasm`): a `wasm-bindgen` fn taking shapes + data strings,
  running validation, returning the SARIF 2.1.0 string via
  `purrdf_validate::to_sarif_string`. Must stay in `make wasm` sweep.
- **C-ABI** (`crates/rdf-capi`): a `cargo-c` extern `"C"` fn with the buffer/cursor
  conventions already used in that crate, returning the SARIF string.
- Round-trip tests on both (validate a fixture → SARIF → schema/golden check).

### After Task 10 Implementation:
- Commit: github-commit-agent. Push: github-push-agent. Post update to issue #48.

## Task 11: Message-quality + SARIF rule-metadata pass (Req 8)

- Audit `Display` texts across `ParseError`, `ShexError`, `IriError`, and the
  text-codec messages: every message states *what was expected*, *what was found*,
  and *where* (line:col). Fix the "unexpected token in {line}" style.
- Add SARIF `ReportingDescriptor` metadata per rule id: `shortDescription`,
  `fullDescription`, `help`, and `helpUri` pointing at the W3C SHACL spec anchors
  for constraint-component rule ids. Rule table lives in `purrdf-validate`.
- Goldens updated to reflect final messages + rule metadata.

### After Task 11 Implementation:
- Commit: github-commit-agent. Push: github-push-agent. Post update to issue #48.

## Task 12: Create Pull Request (MANDATORY FINAL STEP)

1. Run the Mechanical Deferral Scan over the branch diff (no TODO/FIXME/"future"/
   "follow-up"/stub language; no issue/PR numbers in source per `.baseline`).
   Resolve every hit.
2. Full gate: `make check` **and** `make wasm` clean; SPARQL/ShEx/SHACL
   conformance counts unchanged; SARIF schema + determinism tests green.
3. Verify all changes committed and pushed.
4. Create PR:
   ```bash
   gh pr create --title "feat(validate): SARIF 2.1.0 source-traced reporting" --body "Closes #48

   ## Summary
   Adds source-traced, actionable SARIF 2.1.0 output for parse diagnostics and
   SHACL/ShEx validation reports, backed by a shared lazy source-position
   primitive and a new wasm-clean boundary crate purrdf-validate.

   ## Changes
   - iri::pos LineIndex/Position primitive (lazy, error-path only)
   - line/col fidelity in text codecs + SPARQL/ShExC/IRI diagnostics
   - opt-in zero-cost triple->source span table (runtime option, not a feature)
   - purrdf-validate crate: hand-rolled deterministic SARIF 2.1.0 serde model
   - severity/ruleId/message/physical+logical+related location mappings
   - to_sarif() Rust surface + Python/WASM/C-ABI bindings
   - vendored SARIF 2.1.0 schema + Python jsonschema CI validation
   - message-quality + rule-metadata pass
   "
   ```
5. Post plan + the two reflection answers to PR and issue:
   ```bash
   gh pr comment $PR_NUM --body-file /home/paudley/.claude/plans/48-zippy-locket.md
   ```
6. Report: "Stage 1 complete. PR #<num> created for issue #48. Ready for /stage2".

## Execution Rules

1. Task 0 (worktree) is MANDATORY — never work on `main`; all work in
   `.worktrees/48-sarif-reporting`.
4. Run wiring-verifier after each implementation task; treat any deferral
   language in a sub-agent summary as FAILED and re-dispatch.
5. Task 12 (PR) is MANDATORY — never skip.

## Verification

- **Gate:** `make check` (fmt + clippy pedantic/nursery `-D warnings` + build +
  tests + hygiene: no-features/licenses/corpus-frozen/generated/issue-refs +
  rdf-core ring-fence) and `make wasm` (now including `-p purrdf-validate`).
- **Conformance:** run the SPARQL/ShEx/SHACL conformance harnesses; accept/reject
  counts must be byte-for-byte unchanged from `origin/main` (position tracking is
  additive only).
- **SARIF validity:** Python `jsonschema` test validates generated SARIF against
  the vendored SARIF 2.1.0 schema.
- **Determinism:** serialize-twice byte-equality test + committed byte goldens.
- **Perf:** `cargo bench -p purrdf --bench native_codecs` as the report-only
  reference; confirm the `NoSpans` path is unchanged (no speedup asserted).
- **End-to-end:** build the Python extension in the worktree; validate an
  `example.org` fixture with a known violation; assert `to_sarif()` emits a
  physical location (line:col) tracing to the input document and an actionable
  message; validate the output against the schema.

paudley:
## Reflection

1. **The parallel-path line numbering under span tracking.** Document-global line numbers in the chunk-parallel N-Triples/N-Quads path rely on a per-chunk base-line prefix sum (count of `\n` in preceding chunks). Span tracking pins the sequential pipeline, so the tracked path is exercised, but the *parallel* line-number path only triggers on ≥1 MiB inputs and is covered by the first-error-wins determinism tests rather than a dedicated large-input line-number assertion. I'm confident in the arithmetic (chunks split exactly after `\n`), less so that there's no off-by-one on a file whose final line lacks a trailing newline — though that's the last chunk, so it can't shift later chunks.

2. **Focus-node → span join coverage.** The physical-location join keys on the subject's *first* assertion position (bare IRI / `_:label`). It's correct for the common case (focus node is a subject in the data), but a focus node that only appears in object position, or a blank-node focus whose label was rescoped during interning, would fall back to logical-only. That fallback is intentional and schema-valid, but I haven't exhaustively characterized every SHACL focus-node shape against the SpanTable key convention.

3. **The `helpUri` anchor format.** I derive `https://www.w3.org/TR/shacl/#<LocalName>` for constraint components. The common ones (Datatype, MinCount, …) match the W3C SHACL spec anchors, but I did not verify every one of the ~40 components resolves to a live anchor; a few may 404. It never affects SARIF validity (it's an optional hint).

- **`make check` was fmt-dirty on the committed branch until the final commit.** The commit hooks did not enforce `cargo fmt --check`, so several earlier commits landed rustfmt-noncompliant; the final `style:` commit normalized all of it and `make check` is now green. Worth knowing the per-commit hooks are weaker than the `make check` gate.
- **The Python schema-validation lane runs under `make pytest`, not `make check`.** `make check` (the PR gate) covers a dependency-free Rust structural check; the full OASIS-schema validation needs `uv run pytest` (jsonschema was added to the dev group). Both were run and pass here, but CI must run the pytest lane to exercise the real schema.
- **Span tracking forgoes parallelism.** When `track_source_spans` is on, parsing pins the sequential pipeline (the parallel path is NoSpans-only by construction). For very large tracked inputs that trades throughput for source fidelity — a deliberate, documented choice, not a bug.
- **The vendored SARIF schema is third-party.** It sits under `bindings/python/tests/data/` (not a license-scanned vendored root) with a provenance `.license` sidecar; it's the OASIS SARIF 2.1.0 schema redistributed for test validation only.

paudley:
# Stage 2 Gap Analysis & Remediation — Issue #48 (SARIF 2.1.0 reporting), PR #52

## Context

PR #52 (`paudley/48-sarif-reporting`) implements Issue #48 — source-traced SARIF
2.1.0 output for parse diagnostics and SHACL/ShEx validation reports. All 12
planned tasks landed **real, wired code** (not stubs): the deferral hunt is
clean (zero TODO/FIXME/stub/future-work markers, zero issue/PR refs in source),
no Cargo features were added, no new external Rust deps, the `make wasm` sweep
includes `-p purrdf-validate`, and the `purrdf-core`/`purrdf-shapes` ring-fence
is intact.

**But the feature's headline promise — faithful `line:column`/`byteOffset`
source tracing — ships silently wrong.** Two independent bot reviewers
(CodeRabbit, Gemini) plus an adversarial gap-analysis pass converged on a
correctness cluster in the text-codec parser and the SARIF region builder. The
golden tests assert only `line`, never exact `column` or `byteOffset`, so the
defects passed CI unnoticed. This stage fixes every confirmed gap NOW (no
deferrals) and adds the tests/bench that should have caught them.

All work happens in the worktree `/home/paudley/Active/purrdf/.worktrees/48-sarif-reporting`
on branch `paudley/48-sarif-reporting`. Each gap is fixed → committed → pushed
individually (no batching), with a summary posted to issue #48 and PR #52.

## Confirmed gaps (verified against HEAD)

### HIGH — correctness of the core feature

**Gap 1 — TokenCursor located diagnostics point one token PAST the offending token.**
`crates/rdf/src/native_codecs/text_parse.rs`. `bump()` advances `self.pos` and
`col()` reads the token at `self.pos`, so every `col()` computed *after* a
`bump()` reports the *next* token's column. Sites: `expect_dot` (425),
`term()` IRI branch (440), `literal()` LangTag (484-485) and HatHat datatype
(492, 494, 499).
*Fix:* capture `let col = self.col();` **before** the consuming `bump()` at each
site; add tests asserting the exact `column`, not just `line`.

**Gap 2 — Same off-by-one in `DocParser` (`loc()` after consume).**
Same file: `base_directive` (~877, after `expect_iri_raw()` already bumped),
`literal()` LangTag branch (~1166, after `bump()`), `resolve_prefixed`
unknown-prefix error (~1435, caller already bumped the `PrefixedName`).
*Fix:* capture `loc()`/`(line,col)` before the consuming call; thread the
prefixed-name position into `resolve_prefixed` from its callers. Add
exact-column tests. (Correct sites to mirror: `version_string`,
`expect_prefix_ns`, `expect_iri_raw`, HatHat `datatype_iri`.)

**Gap 3 — `region.byteOffset` is never emitted for N-Triples/N-Quads, and dropped when 0.**
Two compounding bugs. `text_parse.rs:351` hardcodes `byte_offset: 0` in the
NT/NQ span record (Turtle/TriG correctly resolve the real offset via
`index.locate`). `build.rs:325` uses `(position.byte_offset > 0).then_some(...)`,
which drops a genuine offset-0 (file start). Req 5 explicitly requires
`region { … byteOffset }`.
*Fix:* thread the real document-global byte offset of the line into the NT/NQ
`Position` record (the per-line running offset already exists in
`parse_lines_sequential`); change `build.rs:325` to
`byte_offset: Some(position.byte_offset)`. Add a test asserting a non-omitted
`byteOffset` (including the offset-0 case).

### MEDIUM

**Gap 4 — "true zero-cost when off" claimed with no Criterion bench.**
`span.rs:12,20` assert byte-identical/zero-cost, but `crates/rdf/benches/native_codecs.rs`
is unchanged. Repo perf discipline (CLAUDE.md/AGENTS.md) requires extending the
criterion benches, not asserting a claim in prose.
*Fix:* add a **report-only** bench arm exercising `parse_dataset` vs
`parse_dataset_with(track_source_spans=false)` on the existing corpus. Per the
project's perf memory, **do not assert a speedup** — the bench is the report-only
reference that tracking-off is unchanged.

**Gap 5 — S0.5 attribution→`UnitInterner` resolution path has zero test coverage.**
`build.rs:266-275` is the sole enforcement point that a numeric `UnitId` never
leaks (resolved via `units.name(...)`), yet every existing test sets
`units: None`.
*Fix:* add a test with `units: Some(&interner)` asserting the emitted
`LogicalLocation.name` is the public slice IRI and the numeric id never appears.

**Gap 6 — `SarifOptions::source_root_uri` is dead (LOW/NO-OPTIONALITY violation).**
Declared at `build.rs:66`, documented as the artifact base URI, but never
threaded — `uri_base_id` is hardcoded `None` (build.rs:320, 394; model.rs:366).
The issue's Task 7 lists `source_root_uri` as a required `SarifOptions` field, so
it must be **wired**, not deleted.
*Fix:* thread `source_root_uri` into `ArtifactLocation.uri_base_id` and emit the
corresponding `run.originalUriBaseIds` entry so the SARIF stays schema-valid.
Add a golden/test covering a set `source_root_uri`.

**Gap 7 — Parallel-path line numbering under no-trailing-newline is untested.**
The chunk-parallel NT/NQ path derives document-global line numbers from a
per-chunk base-line prefix sum; span tracking forces the sequential path, so the
parallel line path is only exercised by first-error-wins tests, never for a
final line lacking a trailing `\n`.
*Fix:* add a >1 MiB fixture whose last line has no trailing newline, asserting
the parallel diagnostic line equals the forced-sequential line.

### LOW

**Gap 8 — Redundant UTF-8 validation for non-text formats when tracking on.**
`parse.rs:158` validates UTF-8 unconditionally; the `RdfXml | TriX | HexTuples`
arm never uses `text`. *Fix:* move `from_utf8` into the text-format match arm.

**Gap 9 — Broken intra-doc link `crate::locate`.** `build.rs:15` links a module
that does not exist (`lib.rs` declares only `build`, `model`, `path_syntax`,
`rules`). *Fix:* rewrite the doc line to point at the inline
`focus_physical_location`/`diagnostic_location`.

**Gap 10 — Duplicated validate-to-SARIF glue.** `crates/rdf-capi/src/shacl.rs:21`
and `crates/rdf-wasm/src/shacl.rs` open-code the same
`validate_graphs(...) → report_to_sarif_string(...)` sequence.
*Fix:* hoist a shared `validate_to_sarif_string(shapes, data)` helper into
`purrdf-validate`; both bindings call it.

**Gap 11 — helpUri anchors may 404.** `rules.rs` derives
`https://www.w3.org/TR/shacl/#<LocalName>`; ~20 of ~40 components curated.
*Fix:* verify the emitted anchors against the live SHACL ToC; drop or correct any
that do not resolve. (Optional hint; never affects SARIF validity.)

## False-positives / gates that PASS (adversarially checked)

Deferral language, issue/PR refs in source, Cargo features, new external Rust
deps, wasm sweep, rdf-core ring-fence, determinism (serialize-twice + byte
golden + BTreeMap property bags, no clock sampling), and the focus-node→span
logical-only fallback are all clean/by-design — no action.

## Remediation order (commit-per-gap, no batching)

1. Gap 1 — TokenCursor column off-by-one + exact-column tests.
2. Gap 2 — DocParser loc off-by-one + tests.
3. Gap 3 — real `byteOffset` (NT/NQ span + `build.rs` `Some(..)`) + test.
4. Gap 6 — wire `source_root_uri` → `uri_base_id` + `originalUriBaseIds` + golden.
5. Gap 5 — `units: Some(..)` S0.5 resolution test.
6. Gap 4 — report-only criterion bench arm (no speedup assertion).
7. Gap 7 — >1 MiB no-trailing-newline parallel-vs-sequential line-number test.
8. Gap 8 — move `from_utf8` into text-format arm.
9. Gap 9 — fix `crate::locate` doc link.
10. Gap 10 — hoist shared capi/wasm SARIF helper.
11. Gap 11 — verify/correct helpUri anchors.

Each: general-purpose sub-agent implements (NO-DEFERRALS clause) →
github-commit-agent (iterate on pre-commit) → github-push-agent → post summary
to issue #48 AND PR #52 → wiring-verifier confirms integration.

## Key files

- `crates/rdf/src/native_codecs/text_parse.rs` — Gaps 1, 2, 3 (parser), 7
- `crates/rdf/src/native_codecs/parse.rs` — Gaps 3 (offset threading), 8
- `crates/rdf/src/native_codecs/span.rs` — Gap 4 (doc claim)
- `crates/rdf/benches/native_codecs.rs` — Gap 4 (bench arm)
- `crates/validate/src/build.rs` — Gaps 3, 5, 6, 9
- `crates/validate/src/model.rs` — Gap 6 (`originalUriBaseIds`)
- `crates/validate/src/rules.rs` — Gap 11
- `crates/validate/src/lib.rs` — Gap 10 (shared helper)
- `crates/rdf-capi/src/shacl.rs`, `crates/rdf-wasm/src/shacl.rs` — Gap 10

## Verification

- `make check` (fmt + clippy pedantic/nursery `-D warnings` + build + all tests +
  hygiene + ring-fence) and `make wasm` clean after every commit.
- New exact-`column` and non-omitted-`byteOffset` assertions for the located
  diagnostics (Gaps 1-3) — these are the tests whose absence let the bugs ship.
- `cargo bench -p purrdf --bench native_codecs` runs as the report-only reference
  for Gap 4 (no speedup asserted).
- W3C SHACL/ShEx/SPARQL conformance counts unchanged (position fixes are message/
  location-only, additive to accept/reject behavior).
- End-to-end: `maturin develop` then validate an `example.org` fixture with a
  known violation at a non-zero byte offset; assert `to_sarif()` emits a physical
  location whose `startColumn` points at the *offending* token and whose
  `byteOffset` is present; run `jsonschema.validate` against the vendored OASIS
  SARIF 2.1.0 schema.

paudley:
### ✅ Gap 1 fixed — `899a525` (HIGH, correctness)

**Gap:** TokenCursor located diagnostics pointed one token *past* the offending token — `col()` was computed after `bump()` had already advanced `self.pos`, so N-Triples/N-Quads error columns reported the *next* token.

**Fix:** capture `let col = self.col()` **before** the consuming `bump()` in `expect_dot`, the `term()` IRI branch, and the `literal()` LangTag / HatHat-datatype branches (`crates/rdf/src/native_codecs/text_parse.rs`). The `term()` "unexpected token" wildcard branch was already correct (uses `peek()`, no prior bump) and is untouched.

**Tests:** 7 new exact-`column` regression tests (`expect_dot_column_points_at_offending_token`, `term_iri_validation_column_points_at_iri`, `langtag_direction_column_points_at_langtag`, `langtag_validation_column_points_at_langtag`, `datatype_non_iri_column_points_at_datatype`, `datatype_iri_validation_column_points_at_datatype`, `datatype_rdf_lang_string_column_points_at_datatype`) — these assert the exact 1-based column, the assertion the original goldens lacked.

**Gates:** `cargo test -p purrdf-rdf` ✅ (166 passed) · clippy `-D warnings` ✅ · fmt ✅.

paudley:
### ✅ Gap 2 fixed — `09cf443` (HIGH, correctness)

**Gap:** The same "position captured after consuming the token" off-by-one, in the `DocParser` (Turtle/TriG) path — `loc()` was read after a consuming call advanced the cursor, so errors pointed one token past the offender.

**Fix** (`crates/rdf/src/native_codecs/text_parse.rs`):
- `base_directive` — capture `(l, c)` before `expect_iri_raw()`; use it for the "base IRI must be absolute" error.
- `literal()` LangTag branch — move the `loc()` capture above `bump()` (mirroring the already-correct HatHat/`datatype_iri` path).
- `resolve_prefixed` — now takes the prefixed-name position as parameters; both callers (`term()` PrefixedName branch and `datatype_iri`) capture `loc()` **before** their `bump()` and pass it in, so "unknown prefix" points at the prefixed name.

**Tests:** 3 new exact-column tests — `base_directive_column_points_at_relative_iri`, `doc_langtag_column_points_at_langtag`, `doc_unknown_prefix_column_points_at_prefixed_name`.

**Gates:** `cargo test -p purrdf-rdf` ✅ (169 passed) · clippy `-D warnings` ✅ · fmt ✅.

paudley:
### ✅ Gap 3 fixed — `0547488` (HIGH, correctness)

**Gap:** `region.byteOffset` was never emitted for N-Triples/N-Quads (two compounding bugs) — Req 5 requires it.
- `text_parse.rs` hardcoded `byte_offset: 0` in the NT/NQ span recorder (Turtle/TriG already resolved the real offset).
- `build.rs:325` used `(position.byte_offset > 0).then_some(...)`, dropping a legitimate start-of-file offset of 0.

**Fix:**
- NT/NQ sequential span path now tracks a running **document-global line-start byte offset** (advanced by payload length + terminator, CRLF-aware) and records it as the subject `Position.byte_offset`. All accumulator maintenance is guarded by `if S::ENABLED`, so the `NoSpans` hot path and the parallel path stay byte-identical (determinism proofs still green).
- `build.rs` emits `byte_offset: Some(position.byte_offset)` unconditionally — 0 is a valid 0-based offset.

**Tests:** `nt_tracking_records_document_global_byte_offset` (asserts line-1 subject → 0, line-2/3 → cumulative offsets) and `physical_location_emits_zero_byte_offset` (SARIF region emits `byteOffset: 0`, not omitted). `example.org` fixtures.

**Gates:** `cargo test -p purrdf-rdf -p purrdf-validate` ✅ (170 + 22) · clippy `-D warnings` ✅ · fmt ✅ · determinism proofs unchanged.

paudley:
### ✅ Gap 6 fixed — `98b8f25` (MEDIUM, LOW/NO-OPTIONALITY)

**Gap:** `SarifOptions::source_root_uri` was declared and documented but inert — every `ArtifactLocation.uriBaseId` serialized as `None`. A dead public option violates the repo's LOW/NO-OPTIONALITY goal; the issue's Task 7 lists it as a required field, so it had to be **wired**, not deleted.

**Fix** (`crates/validate/src/{build,model}.rs`):
- Added `original_uri_base_ids: BTreeMap<String, ArtifactLocation>` to `Run` (serialized `originalUriBaseIds`, `skip_serializing_if` empty, BTreeMap → sorted deterministic keys).
- Threaded a `SRCROOT` base-id through `build_report_sarif_with` / `build_diagnostics_sarif` → `focus_physical_location` / `diagnostic_location`; emitted artifact locations now carry `uriBaseId: "SRCROOT"` when `source_root_uri` is set, and `run.originalUriBaseIds["SRCROOT"].uri` maps to the configured base URI (per SARIF 2.1.0 symbolic-base semantics). When unset, output is byte-unchanged.

**Tests:** `source_root_uri_wires_base_id_and_original_uri_base_ids`, `diagnostic_source_root_uri_stamps_base_id`, `default_emits_no_base_id_and_no_original_uri_base_ids`; structural-conformance test still green.

**Gates:** `cargo test -p purrdf-validate` ✅ (25) · clippy `-D warnings` ✅ · `wasm32` check ✅ · fmt ✅.

paudley:
### ✅ Gap 5 fixed — `d730250` (MEDIUM, test coverage)

**Gap:** The `units: Some(..)` branch in `build.rs` that resolves an attribution's numeric `UnitId` → public slice IRI (via `units.name(...)`) is the sole enforcement point of the S0.5 invariant, but every prior test used `units: None`, leaving it unverified.

**Fix:** Added `attribution_resolves_unit_id_to_slice_iri_and_never_leaks_numeric_id` — interns `http://example.org/slice/1`, attaches an `Attribution` to a result, builds SARIF with `units: Some(&interner)`, and asserts the `LogicalLocation` carries the resolved slice IRI + role (`shape-owner`) and that the numeric id (`unit#0` / `unit#`) appears **nowhere** in the serialized SARIF. No production bug found — the resolution path is correct.

**Gates:** `cargo test -p purrdf-validate` ✅ (26) · clippy `-D warnings` ✅ · fmt ✅.

paudley:
### ✅ Gap 4 fixed — `b36982e` (MEDIUM, perf discipline)

**Gap:** `span.rs` asserted the `NoSpans` path is "true zero-cost when off" in prose, but no Criterion bench backed it — the repo's perf discipline requires a bench extension, not a claim.

**Fix:** Added a **report-only** criterion arm `native_codecs_parse_span_tracking` to `crates/rdf/benches/native_codecs.rs` with `nquads_2k_spans_off` vs `nquads_2k_spans_on` over the existing 2k-row N-Quads fixture (`black_box`, `Throughput::Bytes`, no timing assertions). Reworded the `span.rs` doc to point at this bench as the report-only reference and to name the `parse::tests` that prove the identical-dataset guarantee. **No speedup asserted** (per repo perf policy — machine is not quiet, benches are report-only).

Indicative (non-authoritative) numbers: off ≈1.48 ms, on ≈1.78 ms.

**Gates:** `cargo bench -p purrdf-rdf --bench native_codecs --no-run` ✅ compiles · clippy `-D warnings` ✅ · fmt ✅.

paudley:
### ✅ Gap 7 fixed — `d28d22b` (MEDIUM, test coverage)

**Gap:** The chunk-parallel NT/NQ document-global line-number path (triggered ≥1 MiB) was exercised only by determinism tests, never for a final line lacking a trailing newline — a reflection-admitted off-by-one risk.

**Fix:** Added `parallel_final_line_without_newline_reports_correct_line` — a ~1.4 MiB fixture (20,000 valid rows over `PARALLEL_MIN_BYTES = 1<<20`) with a deliberate parse error on the newline-less final line (20,001). Asserts the `Auto`/parallel path and the forced-sequential path both report line 20,001 and produce byte-identical diagnostics.

**Bug status:** none — the per-chunk base-line prefix sum numbers a newline-less final line correctly; the test locks it in. Test-only addition.

**Gates:** `cargo test -p purrdf-rdf` ✅ (171 + integration) · determinism proofs still green · clippy `-D warnings` ✅ · fmt ✅.

paudley:
### ✅ Gaps 8 & 9 fixed — `620483d`, `30362c4` (LOW)

**Gap 8 (`620483d`)** — `parse_dataset_with` decoded bytes as UTF-8 *before* matching the format, but the RDF/XML/TriX/HexTuples arm never uses `text` (it re-parses from bytes). Moved the `from_utf8` check into the text-format match arm, so byte-oriented formats skip the redundant scan when span tracking is on.

**Gap 9 (`30362c4`)** — `build.rs` module docs referenced `crate::locate`, a module that does not exist (broken intra-doc link). Repointed the doc at the inline `focus_physical_location` / `diagnostic_location`. Verified with `RUSTDOCFLAGS="-D rustdoc::broken_intra_doc_links" cargo doc -p purrdf-validate` — clean.

**Gates:** clippy `-D warnings` (purrdf-rdf + purrdf-validate) ✅ · rustdoc broken-link check ✅ · fmt ✅.

paudley:
### ✅ Gap 10 fixed — `98ed037` (LOW, DRY)

**Gap:** The C-ABI (`validate_to_sarif_bytes`) and WASM (`validate_to_sarif_impl`) bindings each open-coded the same `validate_graphs → report_to_sarif_string` sequence.

**Fix:** Added `validate_to_sarif_string(shapes_ttl, data_nt, &SarifOptions) -> Result<String, String>` in a new `crates/validate/src/shacl.rs` (re-exported from `lib.rs`). Both bindings now reduce to a single call, keeping only their binding-specific wrapping (C-ABI buffer, WASM `JsValue`/String). Public ABI names (`purrdf_shacl_validate_to_sarif`, `shaclValidateToSarif`) are unchanged. Both binding crates dropped their now-unused direct `purrdf-shapes` dependency (reached transitively via `purrdf-validate`).

**Tests:** `validate_to_sarif_string_emits_2_1_0_error` + `malformed_shapes_is_an_error` (example.org) in `purrdf-validate`; existing capi/wasm round-trip tests still green.

**Gates:** `cargo test -p purrdf-validate -p purrdf-capi -p purrdf-wasm` ✅ · clippy `-D warnings` ✅ · `wasm32` check (validate + wasm) ✅ · fmt ✅.

paudley:
### ✅ Gap 11 resolved — `b3b0478` (LOW, verification)

**Gap:** The reflection admitted the derived `helpUri` anchors (`https://www.w3.org/TR/shacl/#<LocalName>`) "may 404".

**Verification:** Checked against the live W3C SHACL Recommendation — every constraint-component section is anchored uniformly at `#<LocalName>` across the value-type (4.1–4.3), string-based (4.4), shape-based (4.7), "other" (4.8), and SPARQL-based (§6) sections. All 20 curated component names in `rules.rs` match the spec's exact local names (incl. `NodeConstraintComponent` for `sh:node`, `PropertyConstraintComponent` for `sh:property`, `SPARQLConstraintComponent` for `sh:sparql`). **No 404s; no code fix needed.**

**Locked in:** added `curated_components_resolve_to_live_spec_anchors` asserting the exact `helpUri` for components *outside* §4.1, so a future edit to the anchor derivation can't silently emit a dead link.

**Gates:** `cargo test -p purrdf-validate` ✅ · clippy `-D warnings` ✅ · fmt ✅.

paudley:
## Stage 2 complete — all gaps remediated, full gate green

Every gap from the approved gap analysis is fixed, each in its own commit, and the full `make check` + `make wasm` gate passes on the branch tip (`b03c7b4`).

### Remediation ledger

| # | Pri | Gap | Fix commit |
|---|-----|-----|-----------|
| 1 | HIGH | TokenCursor located diagnostics pointed one token past the offender (`col()` after `bump()`) | `899a525` |
| 2 | HIGH | Same off-by-one in `DocParser` (`loc()` after consume) | `09cf443` |
| 3 | HIGH | `region.byteOffset` never emitted for NT/NQ (hardcoded 0) + dropped when 0 | `0547488` |
| 6 | MED | `SarifOptions::source_root_uri` was dead — wired into `uriBaseId` + `originalUriBaseIds` | `98b8f25` |
| 5 | MED | S0.5 attribution→`UnitInterner` resolution had no test | `d730250`, `b03c7b4` |
| 4 | MED | "zero-cost when off" claim had no criterion bench (report-only arm added) | `b36982e` |
| 7 | MED | Parallel line numbering untested for a newline-less final line | `d28d22b` |
| 8 | LOW | Redundant UTF-8 validation for byte-oriented formats | `620483d` |
| 9 | LOW | Broken intra-doc link `crate::locate` | `30362c4` |
| 10 | LOW | Duplicated validate-to-SARIF glue (hoisted shared helper) | `98ed037` |
| 11 | LOW | helpUri anchors verified against live W3C SHACL spec (all resolve) | `b3b0478` |

### Re: the re-surfaced CodeRabbit threads (posted 21:43Z)

Those comments repeat the **original** review verbatim — their prose still cites the pre-fix line numbers (`422-426`, `1166-1167`, `crate::locate`, "no criterion bench"). They are re-anchored to HEAD but describe the pre-fix baseline (`0a38947`). Each is **already fixed at HEAD** (`b03c7b4`), verified directly:

- col()/loc()-after-bump → `let col = self.col()` / `(l, c)` captured **before** the consuming call at every flagged site; new tests assert the **exact column** (the assertion the original goldens lacked).
- byteOffset → `byte_offset: Some(position.byte_offset)` unconditional; NT/NQ now record the real document-global line offset.
- crate::locate → removed; doc builds clean under `-D rustdoc::broken_intra_doc_links`.
- zero-cost bench → `native_codecs_parse_span_tracking` (off vs on) added, report-only.

### Gates

`make check` (fmt · clippy pedantic/nursery `-D warnings` · workspace build · all tests · hygiene: no-features/licenses/corpus-frozen/generated/issue-refs · rdf-core ring-fence) ✅ · `make wasm` incl. `-p purrdf-validate` ✅ · W3C conformance behavior unchanged (fixes are message/location-only) ✅.

**Ready for /stage3.**

paudley:
feat(validate): SARIF 2.1.0 source-traced reporting (closes #48)

Uplifts PurRDF's reporting layer to emit byte-deterministic SARIF 2.1.0 for
parse diagnostics and SHACL/ShEx validation reports, with faithful source
tracing (file, line, column, byteOffset) back to the input document and rich,
actionable messages. Task 12 made concrete.

## What this delivers

- **`iri::pos` primitive** — `LineIndex`/`Position` (byte offset → 1-based
  line/column), built lazily on the error path only, hosted in the zero-dep
  `purrdf-iri` leaf so every parser shares one tested primitive at zero happy-
  path cost.
- **Position fidelity** — `ParseError`/`ShexError`/`IriError` gain
  `byte_offset()`/`locate()`; the native text codecs (N-Triples/N-Quads/Turtle/
  TriG) attach real line/column `RdfLocation`s instead of dumping line text.
- **Opt-in span table** — a runtime `ParseOptions.track_source_spans` (a runtime
  option struct, NEVER a Cargo feature) with a const-`ENABLED` ZST collector:
  zero cost when off (`parse_dataset` byte-identical; the parallel path stays
  `NoSpans`-only, determinism proofs unchanged). Records real document-global
  line byte offsets when on.
- **`purrdf-validate`** — new wasm-clean boundary crate hosting a hand-rolled,
  byte-deterministic SARIF 2.1.0 serde model; severity/ruleId/message-synthesis
  mappings; physical (focus→span) + logical (focus/path/shape/component/
  attribution) + related (`shape defined here`) locations; `path_structure`
  rendered as SPARQL property-path syntax; provenance `UnitId` resolved to
  public slice IRIs at the boundary (S0.5, never leaked). `source_root_uri`
  wired through to `uriBaseId` + `run.originalUriBaseIds`.
- **`to_sarif()` across all bindings** — `SarifReport` extension trait (Rust),
  `ValidationReport.to_sarif()` (Python), `shaclValidateToSarif` (WASM),
  `purrdf_shacl_validate_to_sarif` (C-ABI); shared
  `validate_to_sarif_string` helper hoisted into `purrdf-validate`.
- **Schema validation** — vendored OASIS SARIF 2.1.0 schema + a `jsonschema`
  pytest lane, plus a dependency-free Rust structural check in `make check`.
- **Rule metadata** — per-rule `reportingDescriptor` with W3C SHACL spec
  `helpUri` anchors (verified against the live spec; all resolve).

## Stage 2 gap remediation (folded in)

Adversarial gap analysis + two bot reviewers surfaced a correctness cluster the
line-only golden tests missed; all fixed here with exact-column/byteOffset tests:

- HIGH — located diagnostics reported one token PAST the offender in both
  `TokenCursor` and `DocParser` (`col()`/`loc()` computed after the consuming
  `bump()`); positions now captured before consuming.
- HIGH — `region.byteOffset` never emitted for NT/NQ (hardcoded 0) and dropped
  when 0; now records the real line offset and emits `Some(byte_offset)`.
- MED — dead `source_root_uri` option wired; missing S0.5 attribution-resolution
  test added; report-only span-tracking criterion bench added (no speedup
  asserted, per repo perf discipline); parallel line-numbering test for a
  newline-less final line added.
- LOW — lazy UTF-8 validation for byte-oriented formats; broken `crate::locate`
  intra-doc link fixed; duplicated capi/wasm SARIF glue hoisted; helpUri anchors
  verified vs the live W3C SHACL spec.

- GREENFIELD-FIRST / SOTA — hand-rolled SARIF subset, one shared position
  primitive (SUBSUME/EXTEND), no compromises.
- RUST-FIRST, PYTHON-SURFACE — all logic in Rust; bindings are thin surfaces.
- LOW/NO OPTIONALITY, HARD FAILS — the sole opt-in (span tracking) is a runtime
  option struct, not a Cargo feature; no dead options remain (source_root_uri
  now wired); no swallowed errors.
- MAXIMAL PORTABILITY — `purrdf-validate` is wasm32-clean and in the `make wasm`
  sweep; `purrdf-core`/`purrdf-shapes` ring-fence intact (SARIF/serde stay in
  the boundary crate).

## Gates

- `make check` (fmt · clippy pedantic/nursery `-D warnings` · workspace build ·
  all tests · hygiene: no-features/licenses/corpus-frozen/generated/issue-refs ·
  rdf-core ring-fence) GREEN on the merged tree.
- `make wasm` (incl. `-p purrdf-validate`) GREEN.
- W3C SHACL/ShEx/SPARQL conformance behavior unchanged (position tracking is
  additive; message/location-only).
- SARIF byte-deterministic (serialize-twice + byte goldens; BTreeMap property
  bags; no clock sampling).
- No new external Rust deps beyond serde/serde_json; no Cargo features.

## Merge

`origin/main` (entailment epic #37, public-maturity epic #10) merged into the
branch with NO conflicts; full gate re-run GREEN post-merge. New `purrdf-entail`
crate coexists cleanly with `purrdf-validate` in the wasm sweep.

## Deferral gate

Branch diff scan CLEAN (no TODO/FIXME/stub/future-work). The three PR-text
matches are false positives: the Completeness Contract asserting "NOTHING is
deferred", the deferral-scan instruction text from the plan, and CodeRabbit's
"Out of Scope Changes check ✅ Passed" summary — none reference deferred work.

PR #66, opened 2026-07-05
feat(shapes): SHACL Rules — 100% SHACL-AF coverage

### LINKED ISSUE #64: feat(shapes): 100% SHACL-AF coverage — implement SHACL Rules and close every remaining Advanced-Features gap
## Goal
Bring `purrdf-shapes` to **100% SHACL Advanced Features (SHACL-AF) coverage**. Today the SHACL-AF *validation* surface is implemented and gated, but the **Rules** half of the spec is explicitly out of scope, and the vendored DASH corpus is validation-only. This issue closes the gap completely — Rules **and** any other skipped AF surface — so there is nothing left marked "out of scope" for SHACL-AF.

## Current state (what is already done — do not redo)
Per `docs/CONFORMANCE.md`:
- SHACL Core + SHACL-SPARQL: **126/126**, 0 ledgered xfails.
- SHACL-AF **validation** implemented and exercised: `sh:expression` node constraints (`sh:ExpressionConstraintComponent`), custom SPARQL constraint components (`sh:ConstraintComponent`), SPARQL-based functions (`sh:SPARQLFunction`), `sh:SPARQLTargetType`, pre-binding semantics, and the full normative node-expression surface (focus, constant, path, filter-shape, function-call, `sh:union`, `sh:intersection`) plus the PurRDF-owned expression extensions (`sh:if`/`then`/`else`, aggregations, `orderby`/`limit`/`offset`, `sh:exists`).

`crates/shapes/src` has **no** handling of the rules vocabulary, and the vendored `vectors/shacl/af/` corpus is validation-only (no `rules/` tests). SHACL Rules derive and **add** new triples to a graph, driven by a shape's targets.

### Rules vocabulary & execution
- [ ] `sh:rule` attaching one or more rules to a shape; rules run over the shape's **target focus nodes** (all target kinds).
- [ ] **`sh:TripleRule`** — `sh:subject`, `sh:predicate`, `sh:object` each a **node expression** (reuse the existing node-expression evaluator in `crates/shapes/src/expression.rs`); the cross-product of evaluated subject/predicate/object values yields the inferred triples.
- [ ] **`sh:SPARQLRule`** — `sh:construct` (a SPARQL 1.1 `CONSTRUCT` query) with SHACL pre-binding (`$this`, `$shapesGraph`, `$currentShape`); inferred triples = the CONSTRUCT result (reuse `crates/shapes/src/prebinding.rs` + the SPARQL eval crate).
- [ ] **`sh:condition`** — a rule fires for a focus node only if that node conforms to every condition shape.
- [ ] **`sh:order`** — deterministic rule ordering (ascending; ties broken deterministically); `sh:deactivated` on a rule skips it.
- [ ] **Iterative application to a fixpoint** — rules may feed each other; apply until no new triples are derived (with a cycle/termination guard), matching the SHACL-AF rules semantics.

### Architecture (reuse, don't rebuild)
- [ ] The core IR is immutable/value-interned, so rule application **materializes a new frozen `Arc<RdfDataset>`** (base graph + inferred triples) rather than mutating in place — the same shape as `purrdf-entail`'s `materialize`/`materialize_rif` (`crates/entail/src/`). Evaluate reusing `entail`'s forward-chaining chase for `sh:TripleRule` fixpoint instead of writing a second derivation engine; SPARQLRule reuses the SPARQL evaluator.
- [ ] Deterministic inferred-triple emission order (mirror `entail`'s "Deterministic RDFS/OWL inferred-triple emission order" guarantee) — no serializer/IR nondeterminism.
- [ ] Public API on the shapes surface to run rules and obtain the entailed dataset (name/shape TBD in implementation; must be documented).

### Node-expression completeness audit
- [ ] Audit every W3C-normative node-expression kind and confirm each is usable in `sh:subject`/`sh:predicate`/`sh:object` positions (they share the evaluator, but rule positions have subject/predicate legality constraints — a literal in subject/predicate position must hard-fail per the no-swallowed-errors rule).
- [ ] Confirm no other SHACL-AF surface is skipped: SPARQL-based constraints, constraint components, functions, target types, expression constraints are all done — verify none regressed and that this issue leaves **zero** "out of scope" SHACL-AF items.

### Conformance corpus & docs
- [ ] Vendor the **DASH rules** test corpus (the rules portion currently excluded from `vectors/shacl/af/`), byte-frozen and SHA-verified like the rest (`scripts/check-corpus-frozen.py`), discovered/gated by `crates/shapes/tests/w3c_conformance.rs`.
- [ ] Add first-party frozen-report rule fixtures for the PurRDF-owned/edge cases the vendored suite doesn't cover.
- [ ] `docs/CONFORMANCE.md`: remove the "SHACL Rules … remain out of scope" line; add Rules to the SHACL row/matrix with a real pass count; regenerate the conformance matrix (`scripts/conformance-matrix.py --write-doc`) and update the ratchet baseline.

## Acceptance criteria
- SHACL Rules (`sh:TripleRule` + `sh:SPARQLRule`, with `sh:condition`, `sh:order`, `sh:deactivated`, fixpoint) fully implemented and producing a correct entailed dataset.
- The vendored DASH rules corpus passes with **0 ledgered xfails** (or every xfail is a documented, upstream-spec-ambiguity with a ledger reason — no silent skips).
- No SHACL-AF feature remains documented as "out of scope."
- `make check` (fmt + pedantic/nursery clippy + build + tests + hygiene) and `make wasm` green; `make conformance` green with the new rules suite in the scoreboard.

## Constraints (`.goals`)
Greenfield/SOTA/no-compromises; Rust-first; subsume-extend-enhance (reuse the `entail` chase + the shapes node-expression evaluator + the SPARQL eval crate); low/no optionality, hard-fail on unsupported constructs; MAXIMAL performance (id-native, the immutable-IR materialization path); wasm-clean core; no fabricated vocabulary (standard `sh:` terms only); deterministic output.

## Notes
The hard part — a forward-chaining derivation engine over triple patterns — already exists in `purrdf-entail`; the missing work is the SHACL Rules *front-end* (parse `sh:rule`/`sh:TripleRule`/`sh:SPARQLRule`, evaluate node expressions / CONSTRUCT per focus node under conditions and ordering, and drive the fixpoint) plus corpus + docs. This complements the entailment work rather than duplicating it.

paudley:
Closes #64

## Summary
Implement the missing half of SHACL Advanced Features — **SHACL Rules** — bringing `purrdf-shapes` to **100% SHACL-AF coverage** with **zero xfails**.

## Changes
- **Rules engine** (`sh:rule` / `sh:TripleRule` / `sh:SPARQLRule`, with `sh:condition`, `sh:order`, `sh:deactivated`): iterative fixpoint that materializes a new frozen `Arc<RdfDataset>` (base ⊎ inferred). A `Producer` seam keeps rule-kind dispatch out of the loop; cross-round dedup uses owned triples (TermIds aren't stable across the per-round rebuild); value-preserving rules converge with no cap, a fresh-term-minting rule that exceeds an input-derived deterministic bound **hard-fails** naming the offending rule/term. Reuses the node-expression evaluator, target selection, `conforms`, and SPARQL CONSTRUCT + `$this` pre-binding. Public `apply_rules` / `entail_dataset`.
- **Cartesian-product multi-valued function-call arguments** — closes a node-expression completeness gap the DASH corpus surfaced (matches TopBraid/DASH/pySHACL; single-valued is the 1-element case, byte-identical for existing validation).
- **Conformance**: vendored the 7 pySHACL DASH rules tests + 8 first-party fixtures (fixpoint chains, non-monotonic condition, blank-minting, named-graph, …) + 2 error cases; new `rules_conformance.rs` harness compares inferred graphs by RDFC-1.0 isomorphism; byte-frozen. New **SHACL Rules** scoreboard row (17/17 GREEN, 0 xfails); removed the "out of scope" line from CONFORMANCE.md with an AF completeness sign-off.

## Verification
`make check` + `make wasm` + `make conformance` all green; SHACL Rules row 17/17; 360 shapes tests; determinism (byte-identity + isomorphism) and hard-fail (malformed/diverging) covered.

## Summary by CodeRabbit

* **New Features**
  * Added SHACL Rules support, including rule-based inference and canonical N-Triples materialization.
  * Exposed new entailment APIs across Rust, Python, WebAssembly, and C interfaces.
* **Bug Fixes**
  * Improved handling of multi-valued expression arguments and deterministic rule execution.
  * Added safer behavior for malformed inputs and non-terminating rule cases.
* **Tests**
  * Added broad conformance coverage for rule-driven inference, fixpoints, deactivation, ordering, and blank-node stability.
* **Documentation**
  * Updated release notes and SHACL/Rules documentation to reflect the new capabilities.


paudley:
# Implementation Plan for Issue #64 — 100% SHACL-AF coverage (SHACL Rules)

## Issue Summary
Bring `purrdf-shapes` to **100% SHACL Advanced Features coverage** by implementing
the one missing half — **SHACL Rules** (triple derivation) — and closing every
remaining "out of scope" AF item. SHACL-AF *validation* (`sh:expression`, custom
constraint components, `sh:SPARQLFunction`, `sh:SPARQLTargetType`, node
expressions, pre-binding) is already implemented (126/126, 0 xfails). Rules are
absent and documented as out of scope. This ships `sh:rule` / `sh:TripleRule` /
`sh:SPARQLRule` with `sh:condition`, `sh:order`, `sh:deactivated`, iterative
fixpoint application, a vendored DASH rules conformance corpus, first-party
fixtures, and the docs/scoreboard wiring — leaving zero SHACL-AF items unscoped.

Adversarial, no deferrals/follow-ups, no `#NN` refs in repo files/comments, "own
it" (fix any AF gap found, not just Rules; if the node-expression audit finds a
skipped normative kind, close it here).

Greenfield/SOTA/no-compromises; Rust-first; subsume-extend-enhance (reuse the
node-expression evaluator, target selection, conformance check, SPARQL
CONSTRUCT + pre-binding, and the entail-style deterministic materialization — do
not rebuild); low/no optionality, **hard-fail** on unsupported/malformed
constructs and on non-terminating rule sets; MAXIMAL performance (id-native,
immutable-IR materialization); wasm-clean core; standard `sh:` vocab only (no
fabricated IRIs); deterministic output.

## Architecture decision (locked)
- **SHACL Rules live in `purrdf-shapes`** (not `entail`): rules need shapes'
  node-expression evaluator, target selection, and `conforms` check; `entail`
  is lower-level and must not depend on `shapes`. Rules reuse the SPARQL-eval
  crate (already a shapes dep) for `sh:SPARQLRule`.
- **Output is a new frozen `Arc<RdfDataset>`** (base ⊎ inferred triples), the
  first shapes feature to return a data graph rather than a `ValidationReport`,
  mirroring `entail`'s `materialize_*` output shape (union base, sort interned
  triples for determinism, freeze).
- **Fixpoint** is a naive round-based closure (re-run all rules over the growing
  graph until a round adds nothing), guarded for termination (see Task 4).

## Completeness Contract
Every issue requirement maps to a task. Nothing deferred or stubbed.

|---|---|
| `sh:rule` attaches rules; run over the shape's target focus nodes (all target kinds) | 1, 4 |
| `sh:TripleRule` — `sh:subject`/`sh:predicate`/`sh:object` node expressions → triples | 1, 3 |
| `sh:SPARQLRule` — `sh:construct` CONSTRUCT with `$this`/`$shapesGraph`/`$currentShape` pre-binding | 1, 2 |
| `sh:condition` — rule fires only if focus conforms to every condition shape | 1, 4 |
| `sh:order` deterministic ordering; `sh:deactivated` skips a rule | 1, 4 |
| Iterative application to a fixpoint (provable termination; hard-fail on fresh-term divergence) | 4 |
| `Producer` seam (rule-kind dispatch out of the loop); cross-round-stable dedup key | 4 |
| Materialize a new frozen `Arc<RdfDataset>` (immutable-IR, reuse entail-style output) | 4 |
| Deterministic inferred-triple emission order (byte-identity incl. blank-minting rules) | 4, 5 |
| Non-monotonic `sh:condition` (fire-at-check-time, no retraction) fixture | 5 |
| Named-graph target = flattened default graph; rules ∘ entailment composition | 4, 5 |
| Public API to run rules and obtain the entailed dataset (documented, umbrella-re-exported) | 4 |
| Subject/predicate legality (literal in subject/predicate → hard-fail) | 3 |
| Node-expression completeness audit; confirm no other AF surface skipped | 3, 6 |
| Vendor the DASH rules corpus (byte-frozen, SHA-verified, gated) | 5 |
| First-party frozen rule fixtures (fixpoint chains, conditions, ordering, deactivated, cycles, malformed) | 5 |
| New inferred-graph conformance harness (`rules_conformance.rs`, isomorphism compare) | 5 |
| `docs/CONFORMANCE.md`: remove "out of scope", add a Rules scoreboard row, regen matrix + ratchet baseline | 6 |
| DASH rules corpus passes with 0 ledgered xfails (or documented spec-ambiguity ledger) | 5, 6 |
| `make check` + `make wasm` + `make conformance` green | 7 |
| Worktree/branch setup; commit+push per task; PR | 0, 1–7, 8 |

Adopted (folded into the tasks): a formal in-shapes `Producer` seam (Task 4), a
provable termination contract (Task 4), a cross-round-stable dedup key (Task 4),
dual blank-node determinism checks (Tasks 4/5), and non-monotonic-condition,
named-graph, and rules∘entailment-composition fixtures/tests (Tasks 4/5).

Declined (with compelling technical reasons — not "out of scope"):
- **A unified cross-crate `entail::materialize_to_fixpoint(base, producer)`
  framework subsuming RDFS/OWL-RL/RIF/OWL-DL *and* SHACL rules.** REJECTED. The
  four `entail` regimes are **id-native semi-naive** fixpoints over interned
  `[u32;3]` triples (rdfs uses rule-specific incremental indices; rif uses
  compiled Horn triple-patterns) — their performance *is* those id-level
  structures. SHACL rules are **graph-native**: firing needs `conforms` + node
  expressions + SPARQL `CONSTRUCT`, all of which require a materialized
  `Arc<RdfDataset>` each round, so a SHACL rule body cannot participate in a
  semi-naive `[u32;3]` delta-join at all. The only genuinely shared code is a
  ~15-line union-base+`sort_unstable`+freeze emit idiom; sharing it across crates
  would force `shapes → entail` coupling, a `pub` interner, and re-sorting by
  `TermValue` instead of interned-id — which would **change the frozen byte
  output of three shipping regimes** and break their determinism vectors. Forcing
  SHACL through a universal seam buys nothing (SHACL is naive, not semi-naive) and
  risks destabilizing 3 green regimes for 1 new one. The right structure is **two
  families** — id-native pattern rules in `entail` (future SWRL reuses the rif
  chase) and graph-native producers in `shapes` (future SPIN reuses the naive
  driver) — which the plan already builds. (The entail-internal emit duplication
  between `rdfs.rs` and `rif/eval.rs` is a separate DRY cleanup, out of #64's
  scope and not entangled with it.)
- **Semi-naive / delta evaluation for SHACL rules** (compute only the affected
  focus set per round). REJECTED: soundly computing the affected set through
  `conforms` + arbitrary node expressions is complex and error-prone; the corpus
  is small; correctness/determinism dominate; naive re-evaluation to fixpoint is
  the correct default (reason from code, not speedups).
- **`sh:order` stratification to cut rounds.** REJECTED: for monotone
  triple-adding rules the global fixpoint is order-independent; `sh:order` is
  honored for per-round determinism/spec-compliance (which the plan does), not as
  an optimization.
- **Incremental/streaming rule application.** REJECTED: premature; the immutable
  frozen IR re-freezes each round regardless; cost is bounded by the termination
  guard and negligible for the corpus.
- **Reusing `entail`'s `pub(crate)` `Interner` for SHACL fact dedup.** REJECTED: a
  local `FastSet<[RdfTerm;3]>` stays in-crate and avoids a new `shapes → entail`
  dependency and a `pub` interner; the shared surface is too thin to justify the
  coupling.

```bash
BRANCH="paudley/64-shacl-rules"
WORKTREE="$REPO_ROOT/.worktrees/64-shacl-rules"
EXCL="$REPO_ROOT/.git/info/exclude"; grep -qxF '/.worktrees/' "$EXCL" 2>/dev/null || echo '/.worktrees/' >> "$EXCL"
cd "$WORKTREE" && git branch --show-current   # MUST show the feature branch
make check                                     # green baseline before work
gh issue comment 64 --body-file <this-plan-file-path>
```
**⚠️ Do NOT proceed until the worktree exists, on the feature branch, baseline green.**

## Task 1: Rule vocabulary, model, and parsing
Files: `crates/shapes/src/model.rs` (sh: constants), `crates/shapes/src/shapes.rs`
(the `Shape` struct + `parse_shape_inner`), `crates/shapes/src/shapes/parser/node_expr.rs`
(`parse_rules`, mirroring `parse_constraints`).
- Add `sh::` constants: `RULE`, `TRIPLE_RULE`, `SPARQL_RULE`, `SUBJECT`, `OBJECT`,
  `CONSTRUCT`, `CONDITION`, `ORDER` (`PREDICATE` already exists).
- Add a `Rule` model: `enum Rule { Triple { subject: NodeExpr, predicate: NodeExpr,
  object: NodeExpr }, Sparql { construct: String } }` plus common fields
  `conditions: Vec<Term>` (condition shape refs), `order: Option<OrderKey>`,
  `deactivated: bool`. Reuse `NodeExpr` (expression.rs:63) and `parse_node_expr`
  (node_expr.rs:619) for the three TripleRule expressions.
- Add `pub rules: Vec<Rule>` to `Shape` (shapes.rs:313); populate in
  `parse_shape_inner` (shapes.rs:848) via `self.parse_rules(id)?`; fix the other
  `Shape` literal sites the compiler flags (e.g. `constraints.rs:312 ps_as_shape`).
- `parse_rules(id)` reads `objects_of(id, sh::RULE)`, dispatches on `rdf:type`
  (`sh:TripleRule`/`sh:SPARQLRule`), reads the members, `sh:condition` (shape
  refs), `sh:order`, `sh:deactivated`. Hard-fail (`Err`) on a malformed/unknown
  rule node — no silent skip.
- Unit tests: parse each rule kind, ordering, deactivated, conditions; malformed → `Err`.

### After Task 1: commit → push → wiring-verifier → post to #64.

## Task 2: `sh:SPARQLRule` execution (CONSTRUCT + pre-binding)
Files: `crates/shapes/src/sparql.rs`, `crates/shapes/src/prebinding.rs`.
- Add `run_construct_with_shacl_prebinding(dataset, construct, substitutions,
  shapes_graph_iri, current_shape) -> Result<Arc<RdfDataset>, String>` — copy
  `run_select_with_shacl_prebinding` (sparql.rs:461) but return the
  `SparqlResult::Graph(g) => Ok(g)` arm (currently rejected at sparql.rs:450/494/531).
  CONSTRUCT already yields a frozen `Arc<RdfDataset>` of inferred triples via the
  engine (`eval_construct` → `Outcome::Graph`).
- Add a CONSTRUCT-aware static pre-binding check (a `check_construct` analog of
  `check_select`, prebinding.rs:38) rejecting `AS $this` / illegal pre-binding at
  load time (hard-fail).
- Unit tests: `sh:SPARQLRule` with `sh:construct` binds `$this` and derives the
  expected triples; illegal pre-binding → `Err`.

### After Task 2: commit → push → wiring-verifier → post to #64.

## Task 3: `sh:TripleRule` execution + node-expression completeness
`crates/shapes/src/` (a new `rules.rs` module, or extend `engine.rs`).
- For a focus node, evaluate `subject`/`predicate`/`object` node expressions via
  `eval_node_expr(store, focus, expr, guard)` (expression.rs:321); the cartesian
  product of the three result sets yields inferred triples.
- **Legality (hard-fail):** a literal (or, for predicate, a non-IRI) in
  subject/predicate position is an `Err` (no-swallowed-errors); an IRI/blank
  subject and IRI predicate are required.
- **Completeness:** confirm every W3C-normative node-expression kind works in
  subject/predicate/object position (they share the evaluator). Add a test per
  kind used as a rule expression.
- Unit tests: TripleRule over `( focus )` with path/constant/function-call
  expressions; multi-valued expressions → product; illegal subject/predicate → `Err`.

### After Task 3: commit → push → wiring-verifier → post to #64.

## Task 4: Rule engine driver, fixpoint, deterministic materialization, public API
`crates/shapes/src/engine.rs` (+ the `rules.rs` module), `crates/shapes/src/lib.rs`,
umbrella `crates/purrdf/src/`.
- Make `resolve_focus_nodes` (engine.rs:136) `pub(crate)`; reuse it for rule
  targets. Reuse `conforms` (constraints.rs:247) for `sh:condition`.
- **`Producer` seam (keeps rule-kind dispatch out of the loop — SRP; SPIN plugs
  in here later):** a producer is
  `Fn(&RdfDataset /*current*/) -> Result<Vec<[RdfTerm;3]>, String>`; TripleRule
  evaluation (Task 3) and SPARQLRule CONSTRUCT (Task 2) are the two producers.
  The round driver is generic over `&[Producer]`. This is an in-`shapes`, native
  seam — NOT the rejected cross-crate `entail` framework.
- **Driver (one round):** for each non-deactivated shape, resolve focus nodes;
  for each focus node that conforms to every `sh:condition` shape **at firing
  time**; for each rule in `sh:order` order (stable tie-break; skip deactivated):
  run its producer; collect derived triples.
- **Cross-round-stable dedup key (correctness — the per-round dataset is rebuilt
  so `TermId`s are NOT stable across rounds):** the accumulated source of truth is
  a `FastSet<[RdfTerm;3]>` (owned `RdfTerm`/`TermValue` triples), NOT interned ids;
  "is this triple new?" is decided against that set; the per-round frozen
  `Arc<RdfDataset>` is a projection of it (built via the `build_sparql_dataset`
  pattern, engine.rs:373, `push_dataset(current)` + pushed derived quads).
- **Fixpoint + provable termination (hard-fail, deterministic):** value-preserving
  rules range over a **fixed finite Herbrand universe** (terms in base ∪ the
  rules graph) — monotone triple addition over a fixed universe terminates with
  **no cap** (just "stop when a round adds nothing"). The *only* divergence mode
  is a rule that mints a **fresh term** each round (`BNODE()`/fresh-IRI
  expression). Detect precisely: track the term universe; if a round introduces a
  term absent from base ∪ rules AND the round count exceeds a fixed, input-derived
  deterministic bound, return `Err("SHACL rules did not reach a fixpoint …",
  naming the offending rule/term)` — never loop, never silently truncate. The
  bound is a pure function of input so output stays reproducible.
- **Named-graph target:** rules read from and write to the **flattened default
  graph** (`project_dataset` flattens; validation already operates there), for
  consistency with existing SHACL semantics — state this in the code + a fixture.
- **Composition (free from the design — document, don't build a combined API):**
  `apply_rules`/`entail_dataset` return `Arc<RdfDataset>` exactly like
  `entail::materialize`, so RDFS-close ∘ rule-materialize composes; add one
  composition test + a docs note (no combined pipeline API — that would be
  optionality).
- **Determinism:** emit inferred triples in a stable sorted order (mirror
  `entail`'s `sort_unstable()`, entail/src/rif/eval.rs:108) so the materialized
  dataset is byte-deterministic; blank-node labels minted by a rule/CONSTRUCT
  must be allocated deterministically from the sorted per-round dataset (verify
  the SPARQL engine's `BNODE()`/template-blank allocation is not a random/global
  counter). Tested in Tasks 4 & 5 (byte-identity on a **blank-minting** fixture,
  not only value-preserving rules).
- **Public API:** `pub fn apply_rules(data: &ShaclData, shapes: &Shapes) ->
  Result<Arc<RdfDataset>, String>` and an IR convenience `pub fn
  entail_dataset(data: &RdfDataset, shapes: &Shapes) -> Result<Arc<RdfDataset>,
  String>` (mirror `validate_dataset`, engine.rs:309). Re-export from `lib.rs`
  and the `purrdf` umbrella facade.
- Unit tests: single TripleRule; single SPARQLRule; a rule chain requiring ≥2
  fixpoint rounds; `sh:condition` gating (fires/doesn't); `sh:order` effect;
  `sh:deactivated` skip; a self-feeding-but-bounded rule reaches fixpoint; a
  diverging rule → `Err`; determinism (byte-identical output across runs and
  under isomorphic blank relabeling).

### After Task 4: commit → push → wiring-verifier → post to #64.

## Task 5: Conformance corpus + inferred-graph harness
Files: `vectors/shacl/af/rules/**` (vendored + first-party), new
`crates/shapes/tests/rules_conformance.rs`, `scripts/conformance-frozen/vectors-shacl.sha256`.
- **Vendor the DASH rules tests** from the same pinned pySHACL commit already
  used for AF (`5b46638…`, `test/resources/dash_tests/rules/`) into
  `vectors/shacl/af/rules/`, with a `README.md` documenting provenance/license
  (README is auto-excluded from freezing). Rules tests assert an **inferred
  graph**, so they do NOT use `sht:Validate` — convert to a fixture form pairing
  (data graph + shapes/rules graph → **expected inferred graph**), mirroring the
  first-party corpus three-file shape (`data.ttl` / `shapes.ttl` /
  `expected-inferred.nt`) or a small manifest with a first-party rule-test type.
- **First-party fixtures** for edge cases the vendored suite misses: multi-round
  fixpoint chains, `sh:condition` gating, `sh:order`, `sh:deactivated`, cyclic
  data, malformed subject/predicate (expect a documented error), SPARQLRule +
  TripleRule interaction, plus:
  - **Non-monotonic condition** — a rule gated on condition `C` (e.g. `sh:maxCount`
    / `sh:not`) whose own output would make `C` stop conforming in the next round;
    assert the SHACL model: condition checked **at firing time**, monotonic
    accumulation, **no retraction** (no oscillation). Document the chosen semantics.
  - **Blank-minting determinism** — a rule/CONSTRUCT that mints a blank node;
    assert **byte-identical** output across two runs AND stability under
    isomorphic relabeling of the input (exercises the `.goals` byte-determinism
    contract that isomorphism comparison alone would mask).
  - **Named-graph input** — data in a named graph; confirm rules see the flattened
    default graph and write there.
  - **Composition** — `entail::materialize` (RDFS) ∘ `apply_rules` yields the
    expected combined closure.
- **Harness `rules_conformance.rs`:** discover fixtures, run `apply_rules`, and
  compare the produced inferred graph to the expected graph by **RDF isomorphism**
  (reuse `purrdf::canonicalize` / canonical N-Quads so blank labels don't cause
  false failures — the correct spec-conformance check). Assert an exact case count
  and a ledger whose target is **0** (any ledgered xfail must be a documented
  *upstream spec ambiguity*, never a papered-over bug). Keep the corpus check on
  isomorphism AND the Task-4 byte-identity assertions (the two are complementary:
  isomorphism for spec conformance, byte-identity for the determinism contract).
- Keep the vendored-corpus gating **feature-free** — reuse the existing frozen-SHA
  mechanism only; never introduce a `#[cfg(feature)]` for a test corpus.
- Freeze: `python3 scripts/check-corpus-frozen.py --update` (same `vectors/shacl`
  root — no `GUARDED_ROOTS` edit).

### After Task 5: commit → push → wiring-verifier → post to #64.

## Task 6: Scoreboard/docs wiring + AF completeness sign-off
Files: `scripts/conformance-matrix.py`, `scripts/conformance-baseline.json`,
`docs/CONFORMANCE.md`, `AGENTS.md`.
- Add `_suite_shacl_rules()` scraper (runs `cargo test -p purrdf-shapes --test
  rules_conformance`, scrapes a `RULES: passed N total N` scoreboard line) and
  register it in `native_suites()`; add a `"SHACL Rules"` key to
  `conformance-baseline.json["suites"]` with `ledgered` = the xfail budget
  (target 0).
- `docs/CONFORMANCE.md`: **remove** the "SHACL Rules (inference / triple
  derivation) remain out of scope." line; add the Rules row; regenerate the
  matrix block via `python3 scripts/conformance-matrix.py --write-doc`.
- **AF completeness audit sign-off:** in the docs, state that every SHACL-AF
  component (validation + rules) is now implemented — verify nothing else is
  marked out of scope (constraint components, functions, target types, node
  expressions, expression constraints already done; Rules now done).

### After Task 6: commit → push → wiring-verifier → post to #64.

## Task 7: Full gate
```bash
make check      # fmt + clippy pedantic/nursery + build + tests + hygiene
make wasm       # every release crate wasm32-clean (purrdf-shapes included)
make conformance   # matrix GREEN incl. the new SHACL Rules row; no doc drift
```
Fix anything red (incl. conformance-doc regeneration). Commit if changes.

### After Task 7: commit (if changes) → push → wiring-verifier → post to #64.

## Task 8: Create Pull Request (MANDATORY FINAL STEP)
Mechanical deferral scan over the diff, then `gh pr create --title
"feat(shapes): SHACL Rules — 100% SHACL-AF coverage" --body "Closes #64 …"`;
post the plan + the two reflection answers to the PR and #64.
Report: "Stage 1 complete. PR #<num> created for issue #64. Ready for /stage2 64."

## Execution Rules
1. Task 0 mandatory; never work on `main`.
2. Commit after EACH task (github-commit-agent); push after each
   (github-push-agent); wiring-verifier after each.
3. Task 8 (PR) mandatory last.
4. Order: 1 (model/parse) → 2 (SPARQLRule) → 3 (TripleRule) → 4 (driver/fixpoint/
   API) → 5 (corpus/harness) → 6 (scoreboard/docs) → 7 (gate) → 8 (PR).

## Verification
- `apply_rules`/`entail_dataset` derive correct triples for `sh:TripleRule` and
  `sh:SPARQLRule`; conditions/order/deactivated honored; multi-round fixpoint
  converges; diverging rule set hard-fails; output byte-deterministic and
  isomorphism-stable.
- Vendored DASH rules corpus + first-party fixtures pass with 0 ledgered xfails
  (or every xfail is a documented upstream spec ambiguity).
- `docs/CONFORMANCE.md` no longer says Rules are out of scope; the matrix shows a
  green SHACL Rules row; `make check` + `make wasm` + `make conformance` green.

paudley:
### Reflection

**Least confident, and why**
- **CONSTRUCT blank-node determinism.** The native SPARQL engine mints template blanks (`_:c{n}`) from a per-evaluation counter that resets each call, so two focus nodes would both mint `_:c1` and conflate. I relabel every minted blank with a deterministic per-focus prefix to keep co-reference within one CONSTRUCT while giving distinct foci distinct blanks, and it's byte-stable across runs (tested). It's the right fix, but the blank co-reference semantics across a single multi-triple CONSTRUCT template are subtle enough that an exotic template could surprise me.
- **The fixpoint termination bound.** Value-preserving rules provably terminate (monotone growth over a fixed Herbrand universe, no cap). The only divergence mode — fresh-term minting — is caught by a bound (`|universe| + rule_count + 1`). That bound is sound and deterministic, but it's a *guard*, not a proof for pathological rule interactions; a legitimate very-large closure is bounded by the universe size, so it won't false-trip, but I'd want a stress fixture before claiming it's provably tight.
- **The cartesian-product call-arg semantics is a spec-interpretation choice.** SHACL-AF §5 doesn't define multi-valued function-call arguments; I matched the reference implementations (TopBraid/DASH/pySHACL) rather than ledger an xfail, to hit true 100%. If the WG later specifies otherwise, that's the one place this PR took an interpretive stance.

- The **performance profile of the naive fixpoint** on large graphs (O(rounds × freeze), each round rebuilds the frozen dataset). Correct and fine for the corpus; unbenchmarked at scale — a criterion bench for a wide rule set would be worth adding if rules go on a hot path.
- Whether the **7 upstream DASH cases fully represent** SHACL Rules corners; I added 8 first-party fixtures for the ones I could foresee (non-monotonic conditions, blank-minting, named-graph, ordering), but conformance breadth is bounded by what DASH ships.

paudley:
# Stage 2 — Issue #64 Gap Remediation + 0.3.1 Debug-Perf Bugfix + 0.3.2 Release

**Issue:** #64 — feat(shapes): 100% SHACL-AF coverage (implement SHACL Rules)
**PR:** #66 · **Branch:** `paudley/64-shacl-rules` · **Worktree:** `.worktrees/64-shacl-rules`

## Context

Issue #64 asked for 100% SHACL-AF coverage by implementing the missing **SHACL
Rules** half (triple inference). PR #66 delivers a functionally complete,
green rules engine (17/17 DASH rules, 0 xfails, clippy-clean). The deferral hunt
came back **clean** — no stubs, TODOs, or "future work" markers, and the XFAIL
ledger is genuinely empty. What remains is a set of narrow gaps plus two items
the user added mid-review:

1. **Gaps** from adversarial analysis: one explicit-requirement partial
   (`$shapesGraph` never pre-bound), an incomplete node-expression audit, a
   repo-goal gap (rule entailment is Rust-only), and unaddressed allocation-
   discipline feedback.
2. **0.3.1 debug-perf bug** (new): the zero-copy lexer + id-native interner run
   ~300× slower under test builds, blowing compute-bound test timeouts. **Root
   cause found:** the workspace has no `[profile.dev]`/`[profile.test]` override
   for its own crates — `[profile.dev.package."*"]` optimizes *dependencies only*
   (Cargo excludes workspace members from the `"*"` glob), so the hot crates
   compile at `opt-level=0` (no inlining) with overflow-checks in every hot loop.
3. **Ship 0.3.2** (new): a fix release bundling SHACL-AF + the debug-perf fix.
   **Critical:** this branch is still at **0.3.0**; 0.3.1 lives only on `main`,
   so `main` must be merged in before the 0.3.2 bump or the changelog skips 0.3.1.

**User decision (confirmed):** rule entailment will be exposed on **Python + wasm
+ C-API**, at full parity with `validate()`.

## Standing constraints (never violated)

`.goals`: greenfield/SOTA, Rust-first/Python-surface, hard-fails/no-optionality,
maximal utility/performance/portability (wasm-clean core).
`.baseline`: adversarial, **NO DEFERRALS**, no `#NN`/PR refs in repo files, own
every bug. Project: no Cargo features, no hand-editing `generated/`/`vectors/`,
deps pinned in root `[workspace.dependencies]` only, deterministic serializers,
no oxigraph/PyO3 near `purrdf-core`, every release crate wasm-clean, PurRDF/purrdf
casing, no minted vocab IRIs, fixtures use `example.org`.

---

## Phase 4A — Gap remediation (each: fix → commit → push → post → wiring-verify)

### G1 [MEDIUM] `$shapesGraph` never pre-bound in `sh:SPARQLRule`
The issue explicitly lists `$shapesGraph` alongside `$this`/`$currentShape`.
`crates/shapes/src/rules.rs:383` builds `ShaclData::new(ds, ds, None)` and
`rules.rs:395` calls `run_construct_with_shacl_prebinding(ds, …, None, Some(&shape.id))`
— the shapes-graph arg is `None`, so `$shapesGraph` is unbound.
**Fix:** thread the shapes-graph dataset/IRI through `entail_dataset` →
`sparql_rule_producer` → `run_construct_with_shacl_prebinding` so `$shapesGraph`
binds consistently (reuse the existing prebinding path in `crates/shapes/src/prebinding.rs`/`sparql.rs`).
Add a fixture whose CONSTRUCT reads `$shapesGraph` to prove the binding.

### G2 [MEDIUM] Node-expression completeness audit is thin
`node_expr_kinds_work_in_rule_position` (`rules.rs:~800`) exercises the object
position only and omits **function-call** (`FnCall::Builtin`/`UserDefined`) and
**filter-shape** node expressions — the very surface the cartesian-product change
touched. The issue asked for a *confirmed* per-kind audit in subject/predicate/object.
**Fix:** extend the test with positive per-kind cases including function-call and
filter-shape, covering subject and object positions (predicate where legal), plus
the subject/predicate legality hard-fails already present.

### G3 [MEDIUM] Term-not-String in the divergence universe
`build_term_universe`/`fresh_term` (`rules.rs:481-499`) allocate a `String` per
term into `FastSet<String>`. Aligns with the codebase's id-native allocation
discipline (reasoned from allocations, not a benchmarked speedup).
**Fix:** store `Term` in `FastSet<Term>` (Term is `Eq+Hash+Clone`); stringify only
when a fresh term is actually reported. Thread the type change through callers.

### G4 [LOW] Cartesian-product loop allocations in `expression.rs`
`expression.rs:~442-465`: pre-allocate the key vector and reuse one cleared
`bindings` vec across product iterations instead of reallocating + re-formatting
keys each pass. Behavior-neutral cleanup.

### G5 [LOW] Determinism corner fixtures
Add (a) a `sh:SPARQLRule` that mints blanks under a **blank** focus node
(`relabel_blanks` tags with `focus.to_string()` — untested for blank foci), and
(b) a fixpoint-depth fixture that approaches the divergence bound without
false-tripping. Both as first-party `vectors/shacl/af/rules/**` fixtures
(byte-frozen, SHA-updated) or in-crate tests.

> **Declined with evidence (NOT a deferral):** Gemini's "semi-naive delta chase"
> (`rules.rs:229`) is a perf *suggestion*, not a correctness bug. SHACL rules need
> `conforms`/CONSTRUCT re-evaluated over the materialized graph each round; the
> author's plan rejected semi-naive as premature for the small corpus. Per repo
> memory ("don't benchmark to optimize") we keep the naive fixpoint and document
> this reasoning in the PR reply. This is a reasoned resolution, not future work.

### G6 [HIGH] Rule entailment on Python + wasm + C-API (PYTHON-SURFACE goal)
`apply_rules`/`entail_dataset` are Rust-only; `validate()` is exposed on all three
FFI surfaces. For a public SHACL-AF release the feature must reach the consumer
surfaces. **Fix:** add an `entail`/`entail_to_*` entry mirroring the existing
`validate` binding on each surface:
- `bindings/python/src/shacl.rs` — `#[pyfunction]` mirroring `validate`
- `crates/rdf-wasm/src/shacl.rs` — wasm-bindgen entry (must stay wasm-clean)
- `crates/rdf-capi/src/shacl.rs` — C-ABI entry mirroring `validate`
Each sub-agent first reads the existing `validate` signature on its surface and
mirrors shape (inputs, error mapping, ownership). Add a smoke test per surface.

---

## Phase 4B — 0.3.1 debug-perf bugfix

**Root cause:** workspace members compile at `opt-level=0` under dev/test because
`[profile.dev.package."*"]` (root `Cargo.toml:169`) matches dependencies only.
The lexer (`crates/sparql-algebra/src/lexer.rs`, zero `#[inline]`, overflow-checked
byte-cursor arithmetic in every hot loop) and the interners (`crates/rdf/src/native_codecs/text_parse.rs`,
`crates/rdf-core/src/ir/builder.rs`) pay full un-inlined call overhead. Zero
`debug_assert!` in these hot paths, so **`opt-level` (inlining) is the dominant
lever** — overflow-checks/debug-assertions can stay ON (safety preserved).

**Fix:** add targeted per-package dev-profile overrides to root `Cargo.toml`,
mirroring the existing blessed idiom, at `opt-level = 2`:
```toml
[profile.dev.package.purrdf-sparql-algebra]
opt-level = 2
[profile.dev.package.purrdf-rdf]
opt-level = 2
[profile.dev.package.purrdf-core]
opt-level = 2
```
(Bump any still-slow crate to 3 if 2 is insufficient — release is already 3.)
This is a build-profile regression fix reasoned from code, not a perf optimization
asserting a speedup, so no criterion bench is required. Also fix the stale
"nextest 60s slow-timeout" comments (`crates/rdf/tests/proptest_roundtrip.rs:226`
et al.) — there is no nextest; the comment is misleading.
**Verify:** time a ~4350-triple parse test before/after (expect seconds → tens of
ms); confirm the full `cargo test --workspace` no longer approaches CI timeouts.

---

## Phase 4C — 0.3.2 release preparation

Order matters (branch is at 0.3.0; 0.3.1 is on `main` only):
1. **Merge `main` into the branch** (brings 0.3.1: versions, CHANGELOG 0.3.1
   entry, and the `purrdf-entail` publishable fix that resolves the
   `check-versions.py` publish-list mismatch). Resolve conflicts.
2. `make bump VERSION=0.3.2` — `scripts/set-version.py` rewrites all locked
   version sites (root `[workspace.package]`, `pyproject.toml`, npm `package.json`,
   internal `{path,version}` pins, capi `metadata.capi.library`, `uv.lock`). Do
   **not** hand-edit.
3. `make changelog` — git-cliff stamps a `## [0.3.2]` section (SHACL-AF features +
   debug-perf fix + surface exposure) under the existing 0.3.1 entry.
4. Commit `chore(release): 0.3.2`.
5. `make check` (fmt, clippy, build, tests, hygiene, `check-versions.py`,
   `check-no-features.py`, `check-corpus-frozen.py`, `check-issue-refs.py`) +
   `make wasm` + `make conformance` — all green.

> The actual publish is `make release-tags VERSION=0.3.2` (guarded to `main` only,
> pushes `rust-v0.3.2`/`py-v0.3.2`/`npm-v0.3.2` → trusted-publishing lanes). That
> runs **after the squash-merge to main in Stage 3**, not on this branch.

---

## Phase 5 — Feedback loop
Re-trigger CodeRabbit (`@coderabbitai review` — it was rate-limited and never
reviewed). End the turn after posting; resume on notification (no sleep-waiting).
Fold any new PR feedback back into Phase 4.

## Phase 6 — Final summary
Post to issue #64 AND PR #66: every gap + resolution + commit hash, the semi-naive
decline rationale, the debug-perf root cause + fix, and 0.3.2 readiness. Confirm
ready for `/stage3`.

---

## Execution rules
- **Every gap** dispatched via Task tool: general-purpose (fix) → github-commit-agent
  (commit, iterate on pre-commit) → github-push-agent (push) → wiring-verifier
  (integration). **Commit after EACH fix — never batch.** Post a per-gap summary to
  issue + PR.
- **NO DEFERRALS:** any fix that stubs/downgrades/re-defers is a failed fix — re-dispatch.
- All work inside `.worktrees/64-shacl-rules`; never operate from `main`.

## Verification (end-to-end)
- `cd .worktrees/64-shacl-rules && make check && make wasm && make conformance` — all green.
- `cargo test -p purrdf-shapes --test rules_conformance` — 17/17, 0 xfails.
- New per-surface entail smoke tests pass (Python: `pytest`/maturin dev build; wasm: `make wasm`; C-API: capi test).
- `$shapesGraph` fixture: CONSTRUCT reading the shapes graph derives expected triples.
- Debug-perf: `cargo test` no longer near timeout; targeted parse test drops from seconds to ~tens of ms.
- `python3 scripts/check-versions.py` passes at 0.3.2; CHANGELOG has 0.3.1 **and** 0.3.2 entries.

paudley:
### ✅ G1 [MEDIUM] — `$shapesGraph` now pre-bound in `sh:SPARQLRule`

**Gap:** SHACL-AF requires `sh:SPARQLRule` CONSTRUCT to be evaluated with `$this`, `$currentShape`, **and** `$shapesGraph` pre-bound. The engine bound the first two but passed `None` for the shapes-graph arg (`rules.rs` `sparql_rule_producer` / `ShaclData::new(ds, ds, None)`), so `$shapesGraph` was never bound and the fixpoint rounds didn't even carry the shapes graph.

**Fix (`crates/shapes/src/rules.rs`):** each fixpoint round's data handle is now built by a `build_round_base` helper (a direct analog of the validation path's `build_sparql_dataset`) that re-homes the shapes dataset into a named graph under the shapes-graph IRI. The IRI is derived exactly as the validate path derives it — explicit caller override → the shapes document's own graph IRI → dropped when the shapes dataset is empty. **No fabricated vocabulary IRI.** The IRI is threaded through `prepare_rule` → `sparql_rule_producer` into both `ShaclData::new(...)` and the 4th arg of `run_construct_with_shacl_prebinding`. The **final** entailed dataset is still materialized from `base` alone, so the shapes graph never leaks into output and `rules_conformance` stays byte-exact.

**Proof:** new test `sparql_rule_prebinds_shapes_graph` — a CONSTRUCT with `FILTER(?g = $shapesGraph)` derives its head only when the binding is present; reverting the binding to `None` makes it fail. `example.org` fixtures only.

**Verified:** `purrdf-shapes` 336 lib + `rules_conformance` 17/17 (0 xfails) green; clippy pedantic/nursery `-D warnings` clean.

**Commit:** `e9f5750`

paudley:
### ✅ G2 [MEDIUM] — complete node-expression audit in rule positions

**Gap:** Issue #64 asked for a *confirmed* audit that every normative node-expression kind works in `sh:subject`/`sh:predicate`/`sh:object` of a `sh:TripleRule`. The prior test covered only the **object** position and omitted **function-call** (`FnCall::Builtin`/`UserDefined`) and **filter-shape** — function-call being exactly the surface the cartesian-product change touched.

**Fix (`crates/shapes/src/rules.rs`, test-only):** replaced the object-only test with four audit tests + shared helpers covering **all 20** node-expression kind/variants:
- `audit_object_position_every_node_expr_kind` — every kind in object position
- `audit_subject_position_iri_yielding_kinds` / `audit_predicate_position_iri_yielding_kinds` — every IRI-yielding kind in subject/predicate
- `audit_literal_only_kinds_hard_fail_in_subject_and_predicate` — `Count`/`Sum`/`Exists`/`Call::Builtin`(literal) legality hard-fails

Now proven in rule heads: `xsd:string` builtin call, a real `sh:SPARQLFunction` user-defined call, and `sh:filterShape`/`sh:nodes`. `Min`/`Max` confirmed to yield IRIs and work in subject/predicate too.

**Engine bug found:** none — every kind already routes correctly through `eval_node_expr`; this closes the *audit* requirement with real coverage rather than assertion.

**Verified:** `purrdf-shapes` 339 lib + rules_conformance 17/17 + w3c_conformance green; clippy `-D warnings` clean. `example.org` fixtures only.

**Commit:** `9465045`

paudley:
### ✅ G3 [perf/alloc] — divergence universe keyed on `Term`, not `String`

**Feedback (gemini, `rules.rs:500`):** `build_term_universe` allocated a `String` per term into `FastSet<String>`, and `fresh_term` stringified every produced term every round. Aligns with the codebase's id-native allocation discipline (reasoned from allocations, not a benchmarked speedup).

**Fix (`crates/shapes/src/rules.rs`):**
- `build_term_universe` → `FastSet<Term>`, inserts the owned `Term` directly (no `.to_string()`).
- `fresh_term` → `Option<&Term>`, membership tested on the `Term`; stringification happens **only** on the actual divergence hard-fail path.
- `Term` is already an established set key here (`FastSet<[Term;3]>` in `base_triples`), so `Eq`/`Hash` suitability is precedented.

**Behavior-neutral:** convergence/divergence and the divergence-bound formula are unchanged; the divergence error message is byte-identical. Guarded by `diverging_fresh_term_rule_errors`; `rules_conformance` stays 17/17.

**Commit:** `71eff0e`

paudley:
### ✅ G4 [perf/alloc] — reuse bindings buffer in function-call cartesian product

**Feedback (gemini, `expression.rs:465`):** each cartesian-product iteration reallocated a fresh `bindings` vector and re-formatted the argument keys.

**Fix (`crates/shapes/src/expression.rs`):** the `format!("a{i}")` argument keys are computed **once** before the product loop; a single `bindings` buffer is reused across combinations, overwriting only the bound `Term` each pass (`eval_scalar_expr` takes `&[(String, Term)]` by reference, so reuse is safe — no `unsafe`, no awkward clones).

**Behavior-neutral:** same combinations, same last-arg-first order, identical mixed-radix decode → byte-identical `(key, term)` pairs. Covered by the function-call/cartesian tests (`builtin_call_multi_valued_arg_is_cartesian_product`, `..._two_multi_valued_args_product`, `..._empty_arg_yields_empty_product`, the G2 audit function-call cases) + `rules_conformance` 17/17.

**Commit:** `baf933c`

paudley:
### ✅ G5 [determinism] — blank-focus minting + multi-round convergence corners

Two previously-untested determinism corners now have real coverage (`crates/shapes/src/rules.rs`, test-only):

- **`blank_focus_blank_minting_is_distinct_and_stable`** — a `sh:SPARQLRule` mints a fresh blank per focus while the **focus itself is a blank node** (`sh:targetObjectsOf`), exercising `relabel_blanks` (which tags with `focus.to_string()`) on a `_:...` tag. Asserts: distinct minted blank per focus (**no conflation**, `assert_ne!`), byte-stability across two independent runs (re-derivation re-mints the identical label → converges), and correct structure.
- **`value_preserving_chain_converges_over_many_rounds_without_false_divergence`** — a transitive-closure chain over an 8-node path needing ~6 rounds. Asserts the exact closure (all 28 `i<j` pairs, full `BTreeSet` compare — catches under/over-derivation) and that the multi-round chain does **not** false-trip the divergence bound `|universe| + rule_count + 1`.

**Bug found:** none — the engine already handles both corners correctly; this closes the coverage gap. `rules_conformance` stays 17/17.

**Commit:** `90d0231`

paudley:
### ✅ G6 [HIGH] — SHACL rule entailment now reachable on Python + wasm + C-API

**Gap (.goals PYTHON-SURFACE):** `validate` was exposed on all three FFI surfaces, but rule entailment (`entail_dataset`) was Rust-only — the headline SHACL-AF deliverable was inaccessible to the primary consumer surfaces.

**Fix — shared plumbing (reuses existing serializer, no reinvented marshalling):**
- `crates/shapes/src/engine.rs` → `entail_graphs(data_nt, shapes_ttl) -> Arc<RdfDataset>` — the text-in twin of `validate_graphs`; parses identically, calls `rules::entail_dataset`, returns the materialized base⊎inferred dataset.
- `crates/validate/src/entail.rs` (new) → `entail_to_ntriples_string` — serializes via the reused canonical `purrdf_rdf::canonical_flat_nquads` (deterministic, blank-canonical). Both wasm and C-ABI route through `purrdf-validate`, exactly as they do for validate.

**Per-surface entry (each mirrors `validate`'s I/O + error contract + registration):**
- **Python:** `#[pyfunction] entail(shapes_ttl, data_nt) -> str`, GIL-released, registered like `validate`; `purrdf.shapes.entail(...)`, `.pyi` stub added. Smoke: `bindings/python/tests/test_entail.py` — **4 passed** after `maturin develop`.
- **wasm:** `#[wasm_bindgen(js_name = shaclEntail)] shacl_entail(...) -> Result<String, JsError>`. Smoke: 2 unit tests **passed**; `cargo check --target wasm32-unknown-unknown -p purrdf-wasm` **clean** (wasm32 target installed & verified).
- **C-ABI:** `purrdf_shacl_entail_to_ntriples(...)` matching `purrdf_shacl_validate_to_sarif`'s `PurrdfBuffer`/`purrdf_buffer_free` ownership contract; symbol verified present in the built cdylib via `nm -D`. Smoke: 2 tests **passed**.

**Note (pre-existing, not introduced here):** the cbindgen-generated `crates/rdf-capi/include/purrdf.h` currently omits *all* shacl functions (including the pre-existing `validate` one) — the committed snapshot predates the shacl module. It regenerates via the release-only `make capi-header` lane and will emit both `entail` and `validate` symbols; not hand-edited (generated output). This is tracked for the release step.

**Verified:** build + clippy `-D warnings` clean on `purrdf-python`/`purrdf-wasm`/`purrdf-capi`; `cargo fmt --all` clean. `example.org` fixtures only.

**Commit:** `5e263b6`

paudley:
### ✅ Debug-perf bugfix — parse-hot crates optimized in dev/test profile

**Bug (0.3.1 regression):** the zero-copy lexer + id-native interners ran ~300× slower under dev/test than release (a ~4350-triple parse: ~10ms release vs ~3s test), blowing compute-bound test timeouts.

**Root cause:** the only dev override, `[profile.dev.package."*"] opt-level = 2`, matches **dependencies only** — Cargo excludes workspace members from the `"*"` glob. So the hot workspace crates compiled at `opt-level=0` (no inlining), and with zero `#[inline]` in the byte-cursor lexer and overflow-checked arithmetic in every hot loop, un-inlined opt-0 was the tax. There are **zero `debug_assert!`** on these hot paths, so inlining is the dominant lever.

**Fix (`Cargo.toml`):** targeted `[profile.dev.package.<crate>] opt-level = 2` for the three parse-hot crates — `purrdf-sparql-algebra` (lexer), `purrdf-rdf` (text-parse interner), `purrdf-core` (ir/builder interner). **Overflow-checks + debug-assertions stay ON** (they catch real bugs; inlining, not the checks, was the problem). No base `[profile.dev]` bump (keeps the edit-compile loop fast), no Cargo feature, release/bench profiles untouched. Also corrected a stale comment in `proptest_roundtrip.rs` that referenced a nonexistent "nextest 60s slow-timeout" (there is no nextest).

**Verified:** `-C opt-level=2` confirmed applied via `cargo build -v`; `proptest_roundtrip` property finishes ~0.03s; `purrdf-rdf` suite ~1.5s, `purrdf-shapes` ~0.4s slowest suite; `check-no-features.py`/`check-versions.py` green; wasm build unaffected.

**Commit:** `1b71566`

paudley:
### ✅ Phase 4C — 0.3.2 release prepared (SHACL-AF + debug-perf fix)

The branch was still at **0.3.0**; 0.3.1 lived only on `main`. Sequenced the release so the changelog doesn't skip 0.3.1:

1. **Merged `main` into the branch** (`715d609`) — brings the 0.3.1 release: the **edition-2024 migration**, the version bump, and main's Rust-level entail/validate re-exports. Two conflicts, both where 0.3.1's "expose entail+validate" met this branch's SHACL-AF FFI entail work — resolved keeping **both** (`validate/src/lib.rs` re-export; `rdf-capi/src/shacl.rs` extern fn, adopting edition-2024 `unsafe(no_mangle)`).
2. **Reconciliation** (`83f228b`) — post-merge fixups to stay green under edition 2024: let-chain form for the divergence-offender check (clippy `collapsible_if`), rustfmt import ordering, dropped a banned issue-number reference from an audit comment, and refreshed the CONFORMANCE.md matrix (`purrdf.compat parity` 365→369 from merged suites).
3. **`chore(release): 0.3.2`** (`856d215`) — `make bump VERSION=0.3.2` (lockstep across crates.io/PyPI/npm + 18 internal pins + capi C-ABI + uv.lock), Cargo.lock refreshed, and a hand-verified `## [0.3.2]` CHANGELOG section stamped above 0.3.1 (git-cliff mis-buckets on a merge-carrying branch; the section matches its grouping exactly).

**Gates — all green:**
- `make check` ✅ (fmt, clippy `-D warnings`, build, full test suite, all hygiene: no-features / licenses / corpus-frozen / generated / issue-refs / versions, rdf-core-hygiene, wasm)
- `make wasm` ✅ (every release crate wasm32-clean)
- `make conformance` ✅ — **VERDICT: GREEN**, SHACL Rules **17/17**, SHACL-AF 100%, TOTAL 2855 pass · 10 ledgered · 0 unledgered
- `check-versions.py` ✅ 0.3.2 coherent, 16 publishable crates

**Debug-perf confirmed cured under the merged profile:** the previously-timing-out compute-bound suites now complete in seconds (`make check` ran the full workspace test suite without approaching CI timeouts).

> Release **tags** (`rust-v0.3.2`/`py-v0.3.2`/`npm-v0.3.2` → trusted-publishing lanes) are cut by `make release-tags` **from `main` after the squash-merge** — not on this branch. The cbindgen C header (`purrdf.h`) regenerates in that same release lane (`make capi-header`); it currently omits all shacl symbols including the pre-existing `validate` one.

Commits: merge `715d609` · reconcile `83f228b` · release `856d215`.

paudley:
@coderabbitai review

paudley:
## 🏁 Stage 2 complete — gap remediation + debug-perf fix + 0.3.2 prepared

All gaps from the adversarial analysis are **fixed** (no deferrals), the added debug-perf bug is fixed, and 0.3.2 is release-ready. Every fix was committed individually, pushed, and integration-verified.

### Gaps addressed
| # | Pri | Gap | Resolution | Commit |
|---|-----|-----|-----------|--------|
| G1 | MED | `$shapesGraph` not pre-bound in `sh:SPARQLRule` | round-base carries the shapes graph under the validate-derived IRI; `$shapesGraph` bound; guarded test | `e9f5750` |
| G2 | MED | node-expression audit only covered object position | full 20-kind audit across subject/predicate/object incl. function-call + filter-shape | `9465045` |
| G3 | MED | divergence universe allocated a `String` per term | keyed on `Term`; stringify only on divergence; behavior-neutral | `71eff0e` |
| G4 | LOW | cartesian-product loop re-allocated per combination | hoisted keys, reused `bindings` buffer; order-preserving | `baf933c` |
| G5 | LOW | blank-focus minting + deep fixpoint untested | two genuine determinism tests (no-conflation/byte-stable; exact closure, no false-divergence) | `90d0231` |
| G6 | HIGH | rule entailment Rust-only (PYTHON-SURFACE goal) | `entail` exposed on **Python + wasm + C-API** at validate parity; inferred-triple smoke test each | `5e263b6` |

### Reviewer feedback
- **gemini** (8 comments): the allocation-discipline items (Term-not-String G3, bindings-reuse G4) are **applied**.
- **gemini HIGH — semi-naive delta chase (`rules.rs`)**: **declined with evidence** (a reasoned resolution, not a deferral). SHACL rules require `conforms`/CONSTRUCT to be re-evaluated over the *materialized* graph each round, so a frontier-only delta chase is unsound for the general case; the corpus is small and value-preserving. Per repo policy (reason from code, benches report-only) we keep the naive fixpoint. The determinism/termination guarantees are covered by the divergence bound and the G5 tests.
- **CodeRabbit**: was rate-limited (never reviewed) — re-triggered `@coderabbitai review`.

### Added scope
- **Debug-perf bug** (`1b71566`): dev/test built workspace crates at `opt-level=0` (`[profile.dev.package."*"]` matches deps only). Targeted `opt-level=2` for the parse-hot crates (`purrdf-sparql-algebra`/`purrdf-rdf`/`purrdf-core`), safety checks kept ON. Compute-bound suites are fast again.
- **0.3.2 release** (`715d609` merge · `83f228b` reconcile · `856d215` release): merged 0.3.1/edition-2024, bumped lockstep, stamped CHANGELOG.

### Verification
`make check` ✅ · `make wasm` ✅ · `make conformance` ✅ **GREEN** (SHACL Rules 17/17, SHACL-AF 100%, 2855 pass / 10 ledgered / 0 unledgered) · `check-versions.py` ✅ 0.3.2.

**Ready for `/stage3`** (squash-merge to `main`, then `make release-tags VERSION=0.3.2` to publish; C header regenerates via `make capi-header` in that lane).

paudley:
feat(shapes): SHACL Rules — 100% SHACL-AF coverage, entail on all surfaces, ship 0.3.2

Implements the missing half of SHACL Advanced Features — SHACL **Rules** (triple
derivation) — bringing purrdf-shapes to 100% SHACL-AF coverage, exposes rule
entailment on every consumer surface, fixes a 0.3.1 debug/test build-profile
performance regression, and prepares the 0.3.2 release.

## What this branch accomplishes

SHACL Rules engine (sh:rule over all target focus kinds):
- sh:TripleRule — sh:subject/sh:predicate/sh:object as node expressions,
  cross-product emission, subject/predicate legality hard-fails.
- sh:SPARQLRule — CONSTRUCT with SHACL pre-binding of $this, $currentShape, and
  $shapesGraph; deterministic per-focus blank relabeling.
- sh:condition gating, sh:order, sh:deactivated; iterative fixpoint to closure
  with an input-derived divergence bound (hard-fail, no silent non-termination).
- Materializes a frozen Arc<RdfDataset> (base + inferred), deterministic emission.

Public API + surfaces:
- Rust: apply_rules / entail_dataset re-exported from purrdf-shapes.
- Python + wasm + C-API: an `entail` entry mirroring `validate` on each surface,
  returning the materialized dataset as canonical N-Triples/N-Quads (reuses the
  existing canonical serializer; inferred-triple smoke test on each surface).

Conformance & docs:
- 7 vendored pySHACL DASH rules tests + 8 first-party fixtures + 2 error cases;
  rules_conformance.rs compares inferred graphs by RDFC-1.0 isomorphism; frozen.
- SHACL Rules scoreboard row (17/17 GREEN, 0 xfails); "out of scope" line removed
  from CONFORMANCE.md; matrix + baseline regenerated. SHACL-AF is 100% complete.

Debug-perf regression fix:
- 0.3.1's zero-copy lexer + id-native interners built at opt-level=0 under
  dev/test because [profile.dev.package."*"] matches dependencies only (Cargo
  excludes workspace members). Compute-bound suites ran ~300x slower and blew
  timeouts. Fix: targeted opt-level=2 for the parse-hot crates
  (purrdf-sparql-algebra, purrdf-rdf, purrdf-core); overflow-checks and
  debug-assertions kept ON (inlining, not the checks, was the cost).

Release 0.3.2:
- Lockstep bump across crates.io/PyPI/npm + internal pins + capi C-ABI + uv.lock;
  CHANGELOG stamped with the 0.3.2 section above the merged-in 0.3.1 entry.

## Requirements satisfied (issue #64)
- sh:rule / sh:TripleRule / sh:SPARQLRule with condition/order/deactivated ✓
- $this / $shapesGraph / $currentShape pre-binding ✓
- fixpoint to a closure with a termination guard ✓
- frozen materialized dataset + deterministic emission ✓
- node-expression completeness audit across subject/predicate/object incl.
  function-call (builtin + user-defined) and filter-shape ✓
- DASH rules corpus vendored/frozen/gated + first-party fixtures ✓
- CONFORMANCE.md updated, matrix + ratchet baseline regenerated ✓
- make check / make wasm / make conformance all green; 0 ledgered xfails on Rules ✓
- Additionally (repo PYTHON-SURFACE goal): entail exposed on Python/wasm/C-API ✓

- Rust-first core; FFI surfaces are thin wrappers. No optionality — malformed
  rules, illegal subject/predicate, unresolvable conditions, and divergence all
  hard-fail with named errors. No Cargo features added. No oxigraph/PyO3 near
  purrdf-core. Every release crate stays wasm32-unknown-unknown-clean (make wasm).
  No minted vocabulary IRIs (only standard sh:* terms); fixtures use example.org.
  Deterministic serialization (sort-before-freeze + canonical blank relabeling).
  No GitHub issue/PR references in repo files (check-issue-refs green).

## Deferral Gate — PASS
- Code-diff scan (origin/main...HEAD): CLEAN, no TODO/FIXME/stub/"future work".
- PR body/comment scan: 12 hits, ALL false positives — negations asserting
  completeness ("Nothing deferred or stubbed", "no stubs/TODOs/'future work'
  markers", "not future work"), descriptions of REMOVING the prior
  "out of scope" marker because Rules are now done, a CodeRabbit check name, and
  the documented reasoned decline of the semi-naive delta-chase optimization.
- One reviewer suggestion (gemini: semi-naive delta chase) was DECLINED with
  evidence, not deferred: SHACL rules must re-evaluate conforms/CONSTRUCT over
  the materialized graph each round, so a frontier-only chase is unsound for the
  general case; per repo policy perf work is not speculative. This is a reasoned
  resolution; the naive fixpoint's correctness/termination is covered by the
  divergence bound and determinism tests.

## Conflicts resolved (stage2 merge of main / edition-2024)
- crates/validate/src/lib.rs — kept the new entail_to_ntriples_string re-export
  alongside main's rustfmt-reordered model re-export.
- crates/rdf-capi/src/shacl.rs — kept the new purrdf_shacl_entail_to_ntriples
  extern fn; adopted edition-2024 #[unsafe(no_mangle)] to match the merged
  validate function.
- Post-merge edition-2024 reconciliation: let-chain form for the divergence
  offender check (clippy collapsible_if), rustfmt import ordering, and a
  CONFORMANCE.md matrix refresh (purrdf.compat parity 365 -> 369).

- entail on the FFI surfaces returns the WHOLE materialized dataset (base +
  inferred), unlike validate which returns a report — it reuses the canonical
  flat-nquads serializer for byte-determinism.
- The fixpoint is intentionally naive (re-evaluate all rules over the materialized
  graph each round). If profiling ever justifies a semi-naive chase, it must
  preserve conforms/CONSTRUCT-over-materialized-graph semantics per round.
- The cbindgen C header (crates/rdf-capi/include/purrdf.h) is generated by the
  release-only `make capi-header` lane and currently omits all shacl symbols
  (including the pre-existing validate one); regenerate it there to surface both
  purrdf_shacl_validate_to_sarif and purrdf_shacl_entail_to_ntriples.

Closes #64.

PR #68, opened 2026-07-05
perf(rdf): memoize parser line index — fix quadratic diagnostics scan

paudley:
## Critical hotfix — quadratic parse on diagnostics-heavy documents

`DocParser::loc()` (the token → `(line, column)` resolver for parser diagnostics) takes `&self` and rebuilt the whole-buffer `purrdf_iri::LineIndex` newline scan on **every call**. A document that triggers many diagnostic-adjacent lookups therefore paid an **O(n) scan per lookup — quadratic in source length**, a real cliff on large/pathological inputs.

## Fix
`DocParser::line_index` becomes `std::cell::OnceCell<LineIndex>` instead of `Option<LineIndex>`, so the `&self` paths (`loc()` and the span-recording path) can memoize it. First lookup builds the index once; every subsequent lookup is an `O(log lines)` locate. **Linear, not quadratic.**

## Safety / correctness
- **Behaviour identical** — `LineIndex` is deterministic for a fixed source; memoizing only avoids rebuilding it. Computed positions are unchanged.
- No new dependency; `OnceCell` is `wasm32-unknown-unknown`-clean.
- `cargo clippy -p purrdf-rdf --all-targets -D warnings` clean; `purrdf-rdf` test suite (175+ tests) green.

Ships as the **0.3.3** patch release (release-prep commits follow on this PR).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,16 +4,36 @@ All notable changes to the PurRDF crate suite are recorded here. The suite
 ships one lockstep version across crates.io, PyPI, and npm; pre-1.0, a minor
 bump may carry breaking changes and a patch bump is bugfix-only.
 
+## [0.3.3] - 2026-07-05
+
+### Performance
+
+- **rdf:** Memoize the line index so parser diagnostics stay linear
 ## [0.3.2] - 2026-07-05
+
+### Bug Fixes
+
+- **shapes:** Reconcile SHACL-AF work with the merged 0.3.1 baseline
+
+### Features
+
+- **shapes:** SHACL Rules — 100% SHACL-AF coverage
+## [0.3.1] - 2026-07-05
 
 ### Bug Fixes
 
 - **shapes:** Pre-bind $shapesGraph in sh:SPARQLRule CONSTRUCT execution
 - **build:** Optimize parse-hot workspace crates in dev/test profile to remove ~300x regression
 
+### CI & Build
+
+- **release:** Edition 2024, publish purrdf-entail, expose entail+validate, bump 0.3.1
+- **release:** Edition 2024, publish purrdf-entail, expose entail+validate, bump 0.3.1
+
 ### Documentation
 
 - **conformance:** Add SHACL Rules scoreboard row; SHACL-AF is 100% complete
+- **release:** Changelog for 0.3.1
 
 ### Features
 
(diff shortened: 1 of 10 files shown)


paudley:
perf(rdf): memoize parser line index — fix quadratic diagnostics scan; release 0.3.3

## Summary
Critical patch release. Fixes a quadratic-time cliff in the native text parser's
diagnostics path and ships it as 0.3.3.

## The fix
`DocParser::loc()` resolves a token's `(line, column)` for diagnostics by scanning
the whole source with `purrdf_iri::LineIndex`. Because `loc()` takes `&self`, it
rebuilt the entire newline table on every call — a document that triggers many
diagnostic-adjacent lookups paid an O(n) whole-buffer scan per lookup, i.e.
quadratic in the source length.

`DocParser::line_index` is changed from `Option<LineIndex>` to
`std::cell::OnceCell<LineIndex>` so the `&self` paths (`loc()` and the
span-recording path) memoize it: the first lookup builds the index once, every
subsequent lookup is an O(log lines) locate. Linear, not quadratic.

## Files
- `crates/rdf/src/native_codecs/text_parse.rs` — the OnceCell memoization.
- Release bump: `Cargo.toml`, `Cargo.lock`, `bindings/python/pyproject.toml`,
  `bindings/python/uv.lock`, `crates/rdf-wasm/js/package.json`,
  `crates/{rdf-capi,shapes,slice}/Cargo.toml`, `CHANGELOG.md` — 0.3.3 lockstep.

## Correctness / constraints (.goals) respected
- Behaviour identical — `LineIndex` is deterministic for a fixed source, so
  memoizing only avoids rebuilding it; computed positions are unchanged.
- No new dependency; `OnceCell` is `wasm32-unknown-unknown`-clean (no wasm break).
- No Cargo features added. Determinism preserved. No issue/PR refs in repo files.
- clippy `-D warnings` clean on purrdf-rdf; purrdf-rdf test suite (175+ tests) green.

## Deferral Gate — PASS
- Code-diff scan (origin/main...HEAD): CLEAN.
- PR body/comment scan: CLEAN.

- The `OnceCell` is single-threaded interior mutability (DocParser is never shared
  across threads), which is why `&self` memoization is sound here.
- Perf reasoned from complexity (O(n·k) → O(n + k·log n)), not asserted via a
  benchmark, per repo policy.

PR #98, opened 2026-07-13
feat(shapes): project value vocabularies to enum $defs (projection-only)

### LINKED ISSUE #94: shapes/json_schema: project logic:AbstractIndividualType value vocabularies to enum $defs (projection-only)
### Summary
`purrdf-shapes` `json_schema::compile` currently emits a JSON-Schema `enum` **only** from a SHACL `sh:in` constraint (inline on a property, via `Constraint::In`). Value vocabularies modelled as `logic:AbstractIndividualType` — an `owl:Class` whose members are seeded individuals (e.g. `gmeow:TermStability` = stable/experimental/deprecated) — do **not** compile to an `enum`, because they are OPEN by ontology doctrine ("anchor, not a fence") and therefore carry no `sh:in` on the live-validator shapes.

Consumers that want a **closed-record projection** (JSON Schema, Pydantic v2, etc.) need those vocabularies surfaced as standalone, class-keyed enum `$defs` (e.g. `TermStabilityEnum`) — WITHOUT closing the live SHACL validator's value space.

### Requested behaviour
Given the shape store + backing ontology, optionally emit, for each `logic:AbstractIndividualType` value vocabulary:
- a standalone enum `$def` keyed by the class local name (`{ClassLocal}Enum`), members = the sorted individuals of that class — reachable even when no property references the vocabulary; and
- where a property's `sh:class`/`rdfs:range` is such a vocabulary, a `$ref` to that enum `$def` (preserving cardinality).

This must be **projection-only** (it must NOT add `sh:in` to the validating shape set — the live validator stays open-world) and fully deterministic (sorted members, stable `$def` names).

### Current downstream workaround (to be retired when this lands)
`gmeow-ontology` implements this today as an in-repo enrichment over `json_schema::compile`'s output (`crates/pipeline/src/stages/value_vocab.rs`), applied identically to every `$defs` consumer so the cross-surface agreement gate stays green. This is a bridge: the correct home for the derivation is `purrdf-shapes` itself (SUBSUME/EXTEND/ENHANCE). When this is implemented upstream and gmeow bumps the purrdf pin, the in-repo enrichment should be deleted in favour of the native path.

### When complete
Please reference / reopen `Blackcat-Informatics/gmeow-ontology#1408` so the downstream enrichment can be retired against the native purrdf capability.

paudley:
Closes #94

## Summary
Native, **projection-only** derivation in `purrdf-shapes`: classes designated by a caller-supplied marker (`ValueVocab`) are projected to standalone `{ClassLocal}Enum` `$defs` in both JSON Schema and OpenAPI, with referencing properties (`sh:class` / `rdfs:range`) emitting a `$ref`. The live SHACL validator is **never mutated** — value vocabularies stay open-world ("anchor, not a fence"); no `sh:in` is injected. Subsumes the downstream `gmeow-ontology` `value_vocab.rs` pipeline stage.

## Changes
- **Config seam** — `ValueVocab::new(marker_iri)` (caller-supplied `AbstractIndividualType`, no `Default`) + `ValueVocabProjection { vocab, ontology }`; new `compile_with_value_vocab` entry point (`compile` delegates with `None` → byte-identical to before).
- **Standalone enum `$defs`** — keyed `{ClassLocal}Enum`, members `{"@id":curie}` sorted/deduped, enumerated over the ontology∪shapes union via the id-agnostic `DatasetView` seam. The flat `enum` keyword stays **load-bearing** in every case (openapi-generator / Pydantic v2 emit a native enum), with member symbol names and docs on parallel `x-enum-varnames` / `x-enum-descriptions` arrays (from `rdfs:label`/`rdfs:comment`, canonically-first, load-order-independent).
- **Referencing `$ref`** — `sh:class` and ontology `rdfs:range` references resolve to the enum `$def`, cardinality preserved (scalar vs array); `sh:class` wins over `rdfs:range` with a recorded conflict loss.
- **Fail-closed / honest** — undeclared-namespace marker classes and `{Local}Enum` key collisions hard-fail; blank-node members, empty vocabularies, and class/range conflicts are recorded losses, never silent.
- **Two-surface parity** (JSON Schema `$defs` + OpenAPI `components/schemas`) and full determinism (same-process + cross-load) proven.
- **Acceptance proofs** — projected-instance round-trip through the enum `$ref` (via `instance::project_subject` + boon), and open-validator decoupling (a non-anchor value conforms to the live SHACL validator yet is rejected by the closed enum schema; no `Constraint::In` injected).
- **Owned a pre-existing bug** (per `.baseline`) — `term_enum_value` now delegates to the projector's `project_value`, so `sh:in` enum members exactly match projected values (fixes IRI members emitting bare CURIE strings vs `{"@id":curie}` objects, plus latent numeric/typed-literal/blank-node cases by construction).

## Verification
- `cargo test -p purrdf-shapes` — full suite green (382 lib + conformance + equivalence + never-panic + proptest + doctests).
- `make check` (fmt + pedantic/nursery clippy + build + tests + hygiene + `check-no-features.py`) and `make wasm` — clean.
- No Cargo features; no minted vocabulary IRIs; `wasm32`-clean; byte-deterministic.

## Downstream coordination
Comment posted on `Blackcat-Informatics/gmeow-ontology#1408` — the `value_vocab.rs` enrichment can be retired against a `purrdf` pin bump: https://github.com/Blackcat-Informatics/gmeow-ontology/issues/1408#issuecomment-4960495401

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in compilation mode to project value vocabularies into standalone `{Local}Enum` entries within generated JSON Schema/OpenAPI.
  * Exposed new public types and entry point (`ValueVocab`, `ValueVocabProjection`, `compile_with_value_vocab`) for controlling vocabulary projection.
  * Emitted projected enum metadata consistently (including enum variable names and descriptions).
* **Bug Fixes**
  * Corrected IRI member encoding so `sh:in` values serialize consistently as `{"`@id`": ...}` objects.
  * Enforced deterministic enum naming and clarified precedence for vocabulary resolution, including conflict/loss handling.
* **Tests**
  * Expanded coverage for determinism, collision guards, precedence/loss scenarios, and `sh:in` round-tripping.

### DIFF
--- a/crates/gts/src/compact.rs
+++ b/crates/gts/src/compact.rs
@@ -250,8 +250,8 @@ fn base64url_unpadded(data: &[u8]) -> String {
 }
 
 /// Decode a base64url WITHOUT padding (RFC 4648 §5) string — the inverse of
-/// the private `base64url_unpadded` encoder, used to recover a `stream:cose`
-/// literal's raw COSE_Sign1 bytes (Task 5 signature verification).
+/// `base64url_unpadded`, used to recover a `stream:cose` literal's raw
+/// COSE_Sign1 bytes for signature verification.
 ///
 /// # Errors
 /// HARD-ERRORS on a padding character (`=`) or any byte outside the unpadded
--- a/crates/gts/src/compact.rs
+++ b/crates/gts/src/compact.rs
@@ -250,8 +250,8 @@ fn base64url_unpadded(data: &[u8]) -> String {
 }
 
 /// Decode a base64url WITHOUT padding (RFC 4648 §5) string — the inverse of
-/// [`base64url_unpadded`], used to recover a `stream:cose` literal's raw
-/// COSE_Sign1 bytes (Task 5 signature verification).
+/// the private `base64url_unpadded` encoder, used to recover a `stream:cose`
+/// literal's raw COSE_Sign1 bytes (Task 5 signature verification).
 ///
 /// # Errors
 /// HARD-ERRORS on a padding character (`=`) or any byte outside the unpadded
--- a/crates/rdf/src/gts_certify.rs
+++ b/crates/rdf/src/gts_certify.rs
@@ -777,7 +777,7 @@ fn suppressions_ok(pre: &Graph, post: &Graph) -> Result<bool, CertifyError> {
 /// `pre_bytes`/`post_bytes` are UNTRUSTED (an independent verifier's whole
 /// point is to not trust the claimant), so every digest this function
 /// computes over them goes through the fallible RDFC-1.0 canonicalizer
-/// ([`try_refold_digest`]/[`try_effective_digest`], backing
+/// (the private `try_refold_digest`/`try_effective_digest`, backing
 /// [`try_canonicalize_with`]) rather than the panicking
 /// [`canonicalize_with`]: a pathologically symmetric blank graph that stays
 /// under [`POISON_BLANK_LIMIT`]'s cheap blank-COUNT pre-reject yet exhausts
--- a/crates/rdf/src/gts_certify.rs
+++ b/crates/rdf/src/gts_certify.rs
@@ -777,7 +777,7 @@ fn suppressions_ok(pre: &Graph, post: &Graph) -> Result<bool, CertifyError> {
 /// `pre_bytes`/`post_bytes` are UNTRUSTED (an independent verifier's whole
 /// point is to not trust the claimant), so every digest this function
 /// computes over them goes through the fallible RDFC-1.0 canonicalizer
-/// (the private `try_refold_digest`/`try_effective_digest`, backing
+/// (`try_refold_digest`/`try_effective_digest`, backing
 /// [`try_canonicalize_with`]) rather than the panicking
 /// [`canonicalize_with`]: a pathologically symmetric blank graph that stays
 /// under [`POISON_BLANK_LIMIT`]'s cheap blank-COUNT pre-reject yet exhausts
--- a/crates/shapes/src/instance.rs
+++ b/crates/shapes/src/instance.rs
@@ -158,8 +158,11 @@ fn project_subject_data(data: &RdfDataset, ns: &Namespaces, subject: &Term) -> V
 
 /// Project a single object term into its JSON-LD value form.
 ///
-/// MUST match the value-schema convention in [`crate::json_schema`].
-fn project_value(term: &Term, ns: &Namespaces) -> Value {
+/// MUST match the value-schema convention in [`crate::json_schema`]. It is the
+/// SINGLE source of the value encoding: the schema emitter's `sh:in` enum members
+/// delegate here (`crate::json_schema::term_enum_value`) so a projected value and
+/// its enum member can never drift.
+pub(crate) fn project_value(term: &Term, ns: &Namespaces) -> Value {
     match term {
         Term::NamedNode(n) => json!({ "@id": ns.compact_iri(n.as_str()) }),
         Term::BlankNode(b) => json!({ "@id": format!("_:{}", b.as_str()) }),
--- a/crates/shapes/src/json_schema.rs
+++ b/crates/shapes/src/json_schema.rs
@@ -811,8 +811,7 @@ fn assert_value_vocab_enum_keys_unambiguous(
              class `$def` key — disambiguate before projecting"
         );
         if let Some(prev) = key_to_class.insert(key.clone(), class_iri.clone()) {
-            assert_eq!(
-                &prev, class_iri,
+            panic!(
                 "json_schema: distinct value-vocabulary classes share the enum `$def` key \
                  {key:?} ({prev} vs {class_iri}) — cross-namespace local-name twins; \
                  disambiguate before projecting"
--- a/crates/shapes/src/json_schema.rs
+++ b/crates/shapes/src/json_schema.rs
@@ -3878,6 +3878,26 @@ mod tests {
         );
     }
 
+    #[test]
+    #[should_panic(expected = "collides with a class `$def` key")]
+    fn value_vocab_enum_key_class_def_key_clash_hard_fails() {
+        // A vocab class `logic:Color` enum-keys to `ColorEnum`, which is the SAME
+        // `$def` key a primary-namespace target class `meta:ColorEnum` receives
+        // (local-name keyed) — the enum-key-vs-class-key clash guard must reject
+        // it, distinct from both the twins guard and the undeclared-namespace
+        // guard (both `logic:` and `meta:` are declared namespaces here).
+        compile_vocab(
+            r"
+            logic:Color a logic:AbstractIndividualType .
+            logic:red   a logic:Color .
+
+            meta:ColorEnumShape a sh:NodeShape ;
+                sh:targetClass meta:ColorEnum ;
+                sh:property [ sh:path meta:name ; sh:minCount 1 ] .
+        ",
+        );
+    }
+
     // ── Referencing $ref (Task 3) ────────────────────────────────────────────
 
     /// Compile with a SEPARATE ontology dataset and shapes dataset (both `logic:`
@@ -3983,6 +4003,50 @@ mod tests {
             def(&schema, "Term")["properties"]["meta:stability"],
             json!({ "$ref": "#/$defs/TermStabilityEnum" }),
             "a shapes-side rdfs:range resolves via the ontology∪shapes union scan"
+        );
+    }
+
+    #[test]
+    fn value_vocab_multi_range_picks_largest_class_iri_deterministically() {
+        // When ONE predicate declares rdfs:range over TWO distinct value-vocab
+        // classes, the winner is the class with the lexicographically LARGEST
+        // class IRI (".../logic/Beta" > ".../logic/Alpha"), not load order.
+        let forward = schema_of(&compile_vocab(
+            r"
+            logic:Alpha a logic:AbstractIndividualType .
+            logic:a1    a logic:Alpha .
+            logic:Beta  a logic:AbstractIndividualType .
+            logic:b1    a logic:Beta .
+            meta:choice rdfs:range logic:Alpha, logic:Beta .
+            meta:ThingShape a sh:NodeShape ;
+                sh:targetClass meta:Thing ;
+                sh:property [ sh:path meta:choice ; sh:maxCount 1 ] .
+        ",
+        ));
+        assert_eq!(
+            def(&forward, "Thing")["properties"]["meta:choice"],
+            json!({ "$ref": "#/$defs/BetaEnum" }),
+            "the larger class IRI (logic:Beta) wins the multi-range tiebreak"
+        );
+
+        // Reversed triple order (Beta declared before Alpha) must yield the
+        // IDENTICAL winner — the tiebreak is by value, not by scan order.
+        let reversed = schema_of(&compile_vocab(
+            r"
+            logic:Beta  a logic:AbstractIndividualType .
+            logic:b1    a logic:Beta .
+            logic:Alpha a logic:AbstractIndividualType .
+            logic:a1    a logic:Alpha .
+            meta:choice rdfs:range logic:Beta, logic:Alpha .
+            meta:ThingShape a sh:NodeShape ;
+                sh:targetClass meta:Thing ;
+                sh:property [ sh:path meta:choice ; sh:maxCount 1 ] .
+        ",
+        ));
+        assert_eq!(
+            def(&reversed, "Thing")["properties"]["meta:choice"],
+            json!({ "$ref": "#/$defs/BetaEnum" }),
+            "the winner is order-independent: still logic:Beta when declared first"
         );
     }
 
--- a/crates/shapes/src/json_schema.rs
+++ b/crates/shapes/src/json_schema.rs
@@ -28,6 +28,21 @@
 //! `sh:sparql` / `sh:SPARQLTarget` constraints have no JSON Schema equivalent.
 //! They are never silently skipped: each one is dropped, recorded as a
 //! [`LossRecord`], and annotated with a `$comment` on the affected schema.
+//!
+//! # Value-vocabulary enum projection (opt-in, projection-only)
+//!
+//! [`compile_with_value_vocab`] optionally projects *value vocabularies* — an
+//! `owl:Class` whose members are seeded named individuals, marked by a
+//! caller-supplied [`ValueVocab`] — to standalone `{ClassLocal}Enum` `$defs`. Such
+//! vocabularies are deliberately OPEN in the live SHACL validator ("anchor, not a
+//! fence") and carry no `sh:in`; this derivation surfaces their anchor set as a
+//! closed `enum` for codegen / API consumers WITHOUT ever injecting `sh:in` or
+//! otherwise mutating the validating shape set. Members are `{"@id": curie}`
+//! objects (the projector's encoding, so a projected instance validates); the flat
+//! `enum` keyword stays load-bearing in every case, with member symbol names and
+//! docs carried on the parallel `x-enum-varnames` / `x-enum-descriptions` arrays.
+//! A property whose `sh:class` or ontology `rdfs:range` is such a vocabulary emits
+//! a `$ref` to its enum `$def`, cardinality preserved.
 
 use std::collections::{BTreeMap, BTreeSet};
 
--- a/crates/shapes/src/json_schema.rs
+++ b/crates/shapes/src/json_schema.rs
@@ -29,12 +29,15 @@
 //! They are never silently skipped: each one is dropped, recorded as a
 //! [`LossRecord`], and annotated with a `$comment` on the affected schema.
 
-use std::collections::BTreeSet;
+use std::collections::{BTreeMap, BTreeSet};
 
+use ::purrdf::RdfDataset;
 use serde_json::{Map, Value, json};
 
+use crate::data::{GraphFilter, native_quads};
+use crate::model::{rdf, rdfs};
 use crate::shapes::{Constraint, NodeKindValue, Path, Shape, Shapes, Target};
-use crate::term::Term;
+use crate::term::{NamedNode, Term};
 
 const XSD_NS: &str = "http://www.w3.org/2001/XMLSchema#";
 const RDF_NS: &str = "http://www.w3.org/1999/02/22-rdf-syntax-ns#";
@@ -117,7 +120,7 @@ impl Namespaces {
     /// Returns `Err` when `primary_prefix` resolves in neither `doc_prefixes`
     /// nor the builtins.
     pub fn new(primary_prefix: &str, doc_prefixes: &[(String, String)]) -> Result<Self, String> {
-        let mut merged: std::collections::BTreeMap<String, String> = BUILTIN_PREFIXES
+        let mut merged: BTreeMap<String, String> = BUILTIN_PREFIXES
             .iter()
             .map(|(p, n)| ((*p).to_owned(), (*n).to_owned()))
             .collect();
@@ -239,6 +242,60 @@ pub struct CompiledSchema {
     pub losses: Vec<LossRecord>,
 }
 
+// ── Value-vocabulary projection config ───────────────────────────────────────
+
+/// A caller-supplied marker identifying *value-vocabulary* classes.
+///
+/// A value vocabulary is an `owl:Class` whose members are seeded named
+/// individuals (e.g. `TermStability` = `stable` / `experimental` / `deprecated`).
+/// These classes are deliberately OPEN in the live SHACL validator ("anchor, not
+/// a fence") and so carry no `sh:in`; [`compile_with_value_vocab`] nonetheless
+/// projects each one to a standalone `{ClassLocal}Enum` `$def` for JSON-Schema /
+/// OpenAPI consumers, WITHOUT ever mutating the validating shape set.
+///
+/// `abstract_individual_type` is the IRI of the marker class: a class `V` is a
+/// value vocabulary iff `V rdf:type <abstract_individual_type>`. PurRDF mints no
+/// vocabulary IRIs, so this is caller configuration with NO `Default` — mirroring
+/// [`crate::model::BoxRoleVocab`] and `SliceVocab`. A `None` projection (see
+/// [`compile_with_value_vocab`]) means the feature is INACTIVE, not that a
+/// required dependency is missing; compilation then matches [`compile`] exactly.
+#[derive(Debug, Clone)]
+pub struct ValueVocab {
+    abstract_individual_type: String,
+}
+
+impl ValueVocab {
+    /// Construct a value-vocabulary marker from the marker class IRI.
+    #[must_use]
+    pub fn new(abstract_individual_type: &str) -> Self {
+        Self {
+            abstract_individual_type: abstract_individual_type.to_owned(),
+        }
+    }
+
+    /// The marker class IRI whose instances are value-vocabulary classes.
+    #[must_use]
+    pub fn abstract_individual_type(&self) -> &str {
+        &self.abstract_individual_type
+    }
+}
+
+/// The co-required inputs enabling the value-vocabulary enum projection: the
+/// caller's [`ValueVocab`] marker AND the ontology/data graph holding the
+/// individuals to enumerate.
+///
+/// Bundling the two makes "a marker with no ontology to enumerate" unrepresentable
+/// (fail-closed by type). Enumeration additionally scans the shapes graph, so
+/// range/label axioms declared shapes-side are honored (see
+/// [`compile_with_value_vocab`]).
+#[derive(Debug, Clone, Copy)]
+pub struct ValueVocabProjection<'a> {
+    /// The caller-supplied value-vocabulary marker.
+    pub vocab: &'a ValueVocab,
+    /// The ontology / data graph holding the value-vocabulary individuals.
+    pub ontology: &'a RdfDataset,
+}
+
 // ── Compilation context ──────────────────────────────────────────────────────
 
 /// Accumulates losses while compiling so every emitter helper can record one,
@@ -299,6 +356,35 @@ impl<'ns> Ctx<'ns> {
 /// namespace with no declared prefix, or when two distinct target classes
 /// would share a `$defs` key — see [`Namespaces::def_key`].
 pub fn compile(shapes: &Shapes, ns: &Namespaces) -> CompiledSchema {
+    compile_with_value_vocab(shapes, ns, None)
+}
+
+/// Compile a parsed [`Shapes`] graph, optionally projecting *value vocabularies*
+/// to enum `$defs`.
+///
+/// Identical to [`compile`] when `projection` is `None`. When `Some`, each class
+/// `V` marked as a value vocabulary (`V rdf:type <marker>`, the marker taken from
+/// [`ValueVocab`]) is projected to a standalone `{LocalName(V)}Enum` `$def` whose
+/// members are `V`'s named individuals — enumerated from the projection's ontology
+/// graph UNIONED with the shapes graph — encoded as `{"@id": curie}` objects (the
+/// same encoding the instance projector emits, so a projected instance validates).
+/// The flat JSON-Schema `enum` keyword is always the load-bearing form; member
+/// symbol names and docs ride the parallel `x-enum-varnames` / `x-enum-descriptions`
+/// extension arrays. The projection is DERIVATION-ONLY: it never injects `sh:in`
+/// and never mutates the validating shape set, so the live SHACL validator stays
+/// open-world.
+///
+/// # Panics
+///
+/// Panics (build-time, fail-closed) under the same conditions as [`compile`], and
+/// additionally when a value-vocabulary class is in an undeclared namespace or two
+/// value-vocabulary classes would share a `{LocalName}Enum` `$def` key (or an enum
+/// key collides with a class `$def` key).
+pub fn compile_with_value_vocab(
+    shapes: &Shapes,
+    ns: &Namespaces,
+    projection: Option<&ValueVocabProjection<'_>>,
+) -> CompiledSchema {
     // Keying invariant (Gap D, fail-closed): every primary-namespace `$def`
     // is keyed by the class LOCAL NAME and the `@type` discriminator is
     // `<primary_prefix>:<LocalName>`. That is sound ONLY while every target
@@ -309,6 +395,12 @@ pub fn compile(shapes: &Shapes, ns: &Namespaces) -> CompiledSchema {
     // class from an undeclared namespace or a colliding key HARD-fails the
     // build here instead of silently mis-discriminating or clobbering a `$def`.
     assert_target_class_keys_are_unambiguous(shapes, ns);
+
+    // Value-vocabulary enum `$defs` (projection-only): class IRI → (enum key,
+    // `$def` body). Empty when `projection` is `None`. Enum keys are guarded for
+    // collisions against each other and the class `$def` keys before use.
+    let (vocab_enums, vocab_losses) = value_vocab_enum_defs(shapes, ns, projection);
+    assert_value_vocab_enum_keys_unambiguous(shapes, ns, &vocab_enums);
 
     // PASS 1: compute the set of `def_key`s that WILL receive a `$def`, using
     // the EXACT same iteration AND the EXACT same key function (`ns.def_key`)
@@ -332,8 +424,16 @@ pub fn compile(shapes: &Shapes, ns: &Namespaces) -> CompiledSchema {
             }
         }
     }
+    // The value-vocabulary enum `$defs` are always emitted, so a `$ref` to one
+    // (the referencing-property path) never dangles.
+    for (key, _def) in vocab_enums.values() {
+        emitted_defs.insert(key.clone());
+    }
 
     let mut ctx = Ctx::new(emitted_defs, ns);
+    // Enum enumeration losses (empty vocab, blank-node member) recorded first,
+    // deterministically, ahead of the per-shape losses.
+    ctx.losses.extend(vocab_losses);
 
     // Build $defs: one entry per `sh:targetClass` of every active node shape,
     // keyed by the class local name; the body is the shape compiled as an object
@@ -372,6 +472,14 @@ pub fn compile(shapes: &Shapes, ns: &Namespaces) -> CompiledSchema {
     // branch.
     defs.insert("Node".to_owned(), node_def(&class_names, ns));
 
+    // Value-vocabulary enum `$defs` are inserted AFTER the `class_names` snapshot
+    // (mirroring `Node`) so they are never treated as `@type` class branches, yet
+    // still flow into both surfaces (JSON Schema `$defs` + OpenAPI
+    // `components/schemas`) via the shared `defs` map.
+    for (key, def) in vocab_enums.values() {
+        defs.insert(key.clone(), def.clone());
+    }
+
     let schema = root_schema(&defs, ns);
     let openapi = openapi_doc(&defs);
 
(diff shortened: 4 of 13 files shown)


paudley:
# Implementation Plan for Issue #94

## Context (why this change)

`purrdf-shapes` compiles SHACL shapes into a JSON Schema / OpenAPI projection
(`json_schema::compile`). Today an `enum` is emitted **only** from an inline
SHACL `sh:in` constraint (`Constraint::In`, `crates/shapes/src/json_schema.rs:1091`).

Downstream ontologies model *value vocabularies* — an `owl:Class` whose members
are seeded named individuals (e.g. `TermStability` = stable / experimental /
deprecated) — as **open** classes ("anchor, not a fence"): they deliberately
carry **no** `sh:in`, so the live SHACL validator never closes them. Because they
carry no `sh:in`, they never surface as a JSON-Schema enum, even though
codegen / API consumers want exactly that anchor set as an enum.

The gmeow-ontology repo currently patches this with a downstream pipeline stage
(`crates/pipeline/src/stages/value_vocab.rs`). Per the project's
SUBSUME/EXTEND/ENHANCE doctrine, that derivation belongs upstream in
`purrdf-shapes` as a native, **projection-only** capability that never mutates
the validating shape set. This PR delivers that capability and clears the way to
retire the downstream workaround.

**Intended outcome:** classes designated by a caller-supplied marker are
projected to standalone, class-keyed enum `$defs` (`{ClassLocal}Enum`) in both
the JSON Schema and OpenAPI surfaces, with referencing properties emitting a
`$ref` to them — while the live SHACL validator stays open-world.

- Gap analysis / self-review is adversarial: assume the code is substandard until proven otherwise.
- **NO deferrals / follow-ups / future work** — everything ships in this PR.
- **No GitHub issue/PR numbers or process-flow references in repo code or docs.**
  Cross-repo coordination (criterion 8) happens in GitHub, never in the tree.
- Own every bug you touch; fix it now.

- GREENFIELD-FIRST, SOTA, no compromises.
- RUST-FIRST core; Python is only a surface (no core logic added on the Python side).
- SUBSUME/EXTEND/ENHANCE — this is the canonical home for the derivation.
- LOW/NO OPTIONALITY, HARD FAILS — no Cargo features; fail-closed on real config errors.
- MAXIMAL UTILITY / PERFORMANCE / PORTABILITY (wasm32-clean).

Additional hard constraints from `AGENTS.md` / `CLAUDE.md`:
- **No Cargo features / no `cfg`-gated behavior** (`scripts/check-no-features.py` gates CI).
- **PurRDF mints no vocabulary IRIs** — the `AbstractIndividualType` marker is
  caller-supplied config with **no fabricated default**; feature inactive when absent.
  `rdf:type` (a W3C builtin) may be referenced. Fixtures use `example.org`.
- **Byte determinism** in serializers — sorted members, stable keys, no RNG/iteration-order leakage.
- **wasm32-unknown-unknown clean** (`make wasm`); no new deps.
- Gate before "done": `make check` (fmt + pedantic/nursery clippy + build + tests + hygiene) must be clean.

## Completeness Contract

Every requirement and acceptance criterion from issue #94 maps to a numbered

| R1: each marker vocab class emits a `{ClassLocal}Enum` `$def` of its sorted individuals | Task 2 |
| R2: enum members = sorted individuals of that class | Task 2 |
| R3: enum `$def` emitted even when no property references the vocab (standalone) | Task 2 |
| R4: property whose `sh:class`/`rdfs:range` is a vocab emits a `$ref`, cardinality preserved | Task 3 |
| R5: projection-only — no `sh:in` injected; live SHACL validator stays open-world | Task 4 (executable proof) |
| R6: fully deterministic — sorted members, stable `$def` names, byte-identical reruns | Tasks 2 & 3 (executable proof) |
| R7: opt-in (marker/ontology supplied by caller; feature inactive when absent) | Task 1 |
| R8: coordinate downstream — signal gmeow-ontology so `value_vocab.rs` can retire | Task 7 (GitHub, not repo) |
| AC: enum members match the instance projector's `{"@id":curie}` encoding (round-trip) | Task 4 (executable proof) |
| AC: enum `$defs` appear identically in JSON Schema `$defs` AND OpenAPI `components/schemas` | Task 2 (executable proof) |
| AC: marker detection is caller-supplied config; no fabricated `logic:` default | Task 1 |
| AC: `{ClassLocal}Enum` keys are collision-checked (local-name twins, class-key clashes) | Task 2 |
| E1 (adopted transformation, headline-preserving): the `enum` keyword stays load-bearing in every case; member metadata rides parallel `x-enum-descriptions` (from `rdfs:comment`/`rdfs:label`) aligned to the sorted `enum` | Task 2 (executable proof) |
| E2 (adopted): `x-enum-varnames` (sorted member local names) emitted for codegen symbol names, parallel to `enum` | Task 2 (executable proof) |
| E3 (adopted robustness): marker vocab class in an undeclared namespace hard-fails (`is_known`) | Task 2 |
| E4 (adopted robustness): blank-node members and `sh:class`/`rdfs:range` conflicts are recorded losses, never silent | Tasks 2 & 3 |
| E5 (`.baseline` ownership): fix the pre-existing `sh:in`-over-IRI round-trip bug (`term_enum_value` emits bare CURIE strings vs the projector's `{"@id":curie}`) | Task 5 |

## Design Decisions (grounded in the code)

**Seam — plumb the ontology, don't hardcode.** `compile(shapes, ns)`
(`json_schema.rs:301`) has access only to `Shapes`, which retains
`shapes_dataset` = the **shapes graph**, not the data/ontology graph holding the
individuals (`shapes.rs:372`). Enumerating `?x rdf:type C` therefore requires
passing the ontology dataset in. New public API (additive; `compile` preserved so
all existing tests/doctests are untouched):

```rust
// caller config — mirrors SliceVocab / BoxRoleVocab (no Default, no minted IRI)
pub struct ValueVocab { abstract_individual_type: String }   // the marker class IRI
impl ValueVocab { pub fn new(abstract_individual_type: &str) -> Self; }

// co-required inputs bundled so you cannot supply one without the other
pub struct ValueVocabProjection<'a> { pub vocab: &'a ValueVocab, pub ontology: &'a RdfDataset }

pub fn compile(shapes: &Shapes, ns: &Namespaces) -> CompiledSchema {   // unchanged behavior
    compile_with_value_vocab(shapes, ns, None)
}
pub fn compile_with_value_vocab(
    shapes: &Shapes, ns: &Namespaces, projection: Option<&ValueVocabProjection<'_>>,
) -> CompiledSchema;
```
`None` ⇒ byte-identical to today (feature inactive, per the `BoxRoleVocab`
precedent at `constraints.rs:538`). Bundling into `ValueVocabProjection` makes
"vocab supplied but no ontology" impossible by construction (fail-closed by type).

**Member encoding — the critical correctness point.** Instance IRI values project
as `{"@id": ns.compact_iri(iri)}` objects (`instance.rs:164` `project_value`), NOT
bare CURIE strings. The existing `term_enum_value` (`json_schema.rs:1333`, for
`sh:in`) emits bare strings and the code itself flags this as an encoding mismatch
(`json_schema.rs:865`). Therefore the new enum `$def` members MUST be
`{"@id": curie}` **objects**, so a projected instance validates against the enum.
A dedicated helper (not `term_enum_value`) builds them.

**Concept-scheme projection (transformation — adopted, headline-preserving).** The
deepest formalism this feature is a shadow of is a *controlled vocabulary / code
list* (SKOS-shaped): the marker class is a scheme, its individuals are concepts
carrying an `rdfs:label` (a human symbol) and an `rdfs:comment` (the doc). A bare
`{"@id":curie}` enum discards exactly the metadata the feature exists to serve
(codegen/API consumers — openapi-generator, Pydantic v2 — want a symbol name and a
doc string per member). **Critically, the `enum` keyword itself is the headline
deliverable and MUST remain load-bearing in every case** — mainstream codegen turns a
JSON-Schema `enum` array into a native enum type but does NOT do so for
`oneOf`-of-`const`. So the projection keeps the flat `enum` and attaches metadata via
the conventional **parallel** OpenAPI extension arrays (openapi-generator's own
convention), aligned index-for-index to the sorted `enum`:
```json
{ "enum": [ {"@id":"ex:deprecated"}, {"@id":"ex:experimental"}, {"@id":"ex:stable"} ],
  "x-enum-varnames":     ["deprecated", "experimental", "stable"],
  "x-enum-descriptions": ["…rdfs:comment…", "", "…"],
  "$comment": "projection of an open value vocabulary (anchor, not a fence)" }
```
- `enum` (always, load-bearing): `{"@id":curie}` members, sorted by CURIE, de-duplicated.
- `x-enum-varnames` (always): each member's local name — the identifier-safe codegen symbol.
- `x-enum-descriptions` (only when any member has a doc): `rdfs:comment`, falling back to
  `rdfs:label`, empty string where absent — parallel to `enum`, so codegen still sees an enum.
- empty vocab → `{"enum": [], "x-enum-varnames": []}` + a `LossRecord`.
`rdfs:label` / `rdfs:comment` / `rdfs:range` are W3C builtins (added as IRI consts
alongside the existing `rdfs::SUB_CLASS_OF` in `model.rs`) — referencing them mints
nothing, and the enrichment is maximal-by-default with zero optionality knobs. **The
`enum` keyword is never replaced or dropped**, so the headline "enum $def" is
consumed AS an enum by the issue's named consumers in all cases.

**Enumeration.** Reuse the proven cold-path materializer
`data::native_quads(ds, None, Some(&rdf_type_term), Some(&class_term), GraphFilter::AnyGraph)`
(`data.rs:174`), taking the subject `.0`. It rides the id-agnostic `DatasetView`
seam so it works over both the frozen and paged-u64 backends. Vocab classes =
subjects of `?V rdf:type <marker>`; members = NamedNode subjects of
`?x rdf:type V` (direct typing = "seeded individuals"). All enumeration scans the
**union of the supplied `ontology` and the shapes graph** (`Shapes::shapes_dataset`)
— range/label/individual axioms commonly live shapes-side, and scanning both closes
that silent-gap failure mode; a `BTreeSet` dedup keeps the union deterministic and
idempotent when the caller passes one combined dataset. The scan is localized behind
one private `members_of(datasets, class)` helper (single seam). All results sorted by
compacted CURIE via `BTreeSet<String>` / `sort_canonical` and de-duplicated. A
`rdf:type`-member that is a **blank node** (not a legal enum member) is dropped WITH a
recorded `LossRecord` — never silently. A marker vocab class in an **undeclared
namespace** hard-fails (`ns.is_known`, `json_schema.rs:205`), matching the
target-class treatment in `assert_target_class_keys_are_unambiguous`. The enum key is
`format!("{}Enum", local_name(v))` reusing the existing `pub fn local_name`
(`json_schema.rs:211`) — no new accessor.

**Keys.** Enum `$def` key = `{local_name(V)}Enum` — always colon-free, so it is a
valid OpenAPI `components/schemas` key (`^[a-zA-Z0-9._-]+$`), unlike a non-primary
class `def_key` which is a CURIE. Extend the existing collision guard
`assert_target_class_keys_are_unambiguous` (`json_schema.rs:390`) to also register
enum keys and hard-panic on local-name twins across namespaces and on any clash
with a class `$def` key (fail-closed, matching existing style).

**Two-surface parity is automatic.** `openapi_doc(defs)` clones the same `defs`
map into `components/schemas` (`json_schema.rs:580`); any enum `$def` we insert
appears in both surfaces for free. Enum `$defs` are inserted **after** the
`class_names` snapshot (like `Node`) so they are excluded from the `@type`
discriminator.

**Referencing `$ref` (R4).** `Ctx` gains two maps: class-IRI → enum-key (for
`Constraint::Class(V)`) and predicate-IRI → enum-key (for `Path::Predicate(P)`
whose `rdfs:range` — scanned over the same ontology∪shapes union — is a vocab class,
the primary association for *open* vocabularies that carry no `sh:class`). Matching
properties emit `{"$ref": "#/$defs/{Local}Enum"}` as the scalar schema, then the
existing cardinality wrapper (`json_schema.rs:1194`) preserves single-vs-array. The
enum `$defs` are pre-registered in `emitted_defs` (PASS 1) so no `$ref` dangles.
**Precedence (E4):** an explicit `sh:class` wins over `rdfs:range` — if a property
carries `Constraint::Class(C)` where `C` resolves to a vocab, that enum `$ref` is
used and the range path is skipped for that property; a `sh:class C` where `C` is a
*non-vocab* class also wins (keeps its existing node-ref/`$ref`) and the range path is
suppressed, with a recorded `LossRecord` if the predicate's `rdfs:range` pointed at a
*different* vocab (a genuine class/range conflict). Only a property with no
vocab-resolving `sh:class` takes the `rdfs:range` enum `$ref`.

**Empty vocab edge case.** A marker class with zero individuals emits
`{"enum": []}` plus a recorded `LossRecord` (the existing losses channel) noting
the vocabulary is empty — deterministic and honest rather than silently dropped.

## Enhancements Adopted (from the Enhancement Audit)

Folded into the tasks above and reflected in the Completeness Contract (E1–E5):
concept-scheme member metadata (`rdfs:label`/`comment`) projected **headline-preservingly**
as parallel `x-enum-varnames` / `x-enum-descriptions` arrays that keep the `enum` keyword
load-bearing (the transformation item, corrected so it does not narrow the headline);
`is_known` hard-fail for undeclared-namespace marker classes; blank-node and
`sh:class`/`rdfs:range`-conflict loss records; union scan over ontology∪shapes graph;
`members_of` seam localization; sort-then-pick canonical label/comment selection with a
cross-load determinism test; reuse of the existing `local_name` free function; the "marker
supplied but zero matching classes ⇒ byte-identical to `compile`" regression test; and
owning the pre-existing `sh:in`-over-IRI round-trip bug (E5, Task 5).

- **Python-surface mirror of `compile_with_value_vocab`.** N/A — there is no Python
  binding of the schema compile path in this workspace; `json_schema::compile` is
  reachable only from the Rust facade and in-crate tests/doctests (`profile.rs` uses
  only `Namespaces`, not `compile`). Nothing to mirror.
- **Caller-configurable label/doc predicates (e.g. `skos:prefLabel` / `skos:definition`).**
  DECLINED. `rdfs:label` / `rdfs:comment` / `owl:deprecated` are the universal W3C
  builtins understood by every consumer and every codegen; adding a predicate-config
  knob reintroduces optionality (against `.goals` LOW/NO OPTIONALITY) for a case
  well-formed ontologies already cover (SKOS label properties are conventionally
  `rdfs:label` sub-properties). The universal builtins are the maximal default.
- **`oneOf`-of-`const` member structure (with per-member `title`/`description`/`deprecated`)
  instead of a flat `enum` + parallel `x-enum-*` arrays.** DECLINED. Although `oneOf`-of-`const`
  is more expressive for *validation*, it DELETES the `enum` keyword — and the issue's named
  consumers (openapi-generator, Pydantic v2) generate a native enum type from a JSON-Schema
  `enum` array, NOT from `oneOf`-of-`const`. Emitting `oneOf` would silently narrow the
  headline "enum $def" into something codegen does not treat as an enum. The parallel
  `x-enum-varnames`/`x-enum-descriptions` convention delivers the same metadata while keeping
  the `enum` keyword load-bearing — strictly better for the stated consumer.
- **Per-member `owl:deprecated` projection.** DECLINED. There is no headline-safe flat-`enum`
  channel for a per-member deprecation flag (expressing it requires the `oneOf` form declined
  above, which breaks codegen consumption of the enum). Member names and docs — the metadata
  codegen actually consumes — are delivered via the parallel arrays.
- **Open-enum projection (`anyOf: [{enum}, {type:string/@id}]`) instead of a closed
  enum.** DECLINED. The entire purpose of the projection is to hand codegen/API
  consumers the closed anchor set; the *openness* lives in the validator, which we
  deliberately leave untouched. A closed projection is the correct, intended output —
  an open-enum would defeat the feature.
- **Ordered / lattice-structured vocab members (a "sense lattice" style richer
  structure).** DECLINED. JSON Schema `enum` has no ordering semantics and the
  doctrine here is *openness* ("anchor, not a fence"), not *ordering*. There is no
  ordering signal in the source model to project, so this would fabricate
  structure — a violation of the no-minted-vocabulary spirit. The `ValueVocab`
  marker already delivers the real generalization: *any* class the caller
  designates becomes an enum.
- **`rdfs:subClassOf`-transitive member closure (à la `engine::instances_of_class`).**
  DECLINED. The issue specifies "seeded individuals" = direct `rdf:type`. Value
  vocabularies are flat by construction; transitive closure would add nondeterminism
  surface and pull in members the ontology never seeded. Direct typing is the exact,
  deterministic rule.
- **Criterion bench for `compile_with_value_vocab`.** DECLINED. `crates/shapes` has
  **no** bench harness (`crates/shapes/benches/` does not exist), schema compile is
  an explicitly **cold** path (documented at `data.rs:171`), and project doctrine
  forbids benching-to-optimize (benches are report-only, never assert a speedup).
  Standing up a new bench harness for a cold, non-hot-path feature has no payoff.
- **Making the membership predicate configurable (not just `rdf:type`).** DECLINED.
  `rdf:type` is a W3C builtin (allowed to reference) and is the universal membership
  relation for `owl` individuals; parameterizing it adds config surface with no
  real consumer. The marker class IRI is the one genuine caller-config knob.

1. Resolve top-level repo and set names:
   ```bash
   BRANCH="paudley/94-value-vocab-enum-defs"
   WORKTREE="$REPO_ROOT/.worktrees/94-value-vocab-enum-defs"
   ```
2. Keep `.worktrees/` locally ignored (no tracked-file change):
   ```bash
   ```
   ```bash
   cd "$WORKTREE"
   git branch --show-current   # MUST show the feature branch, NOT main
   ```
4. Prime the build environment in the worktree (worktrees inherit no artifacts):
   `cargo build -p purrdf-shapes` (do NOT gate on a full test run — that is stage2/stage3's job).
   ```bash
   gh issue comment 94 --body-file <this-plan-file-path>
   ```

**⚠️ Do NOT proceed to Task 1 until the worktree exists and you are on the feature branch.**

## Task 1: `ValueVocab` config + additive `compile_with_value_vocab` seam

**Files:** `crates/shapes/src/json_schema.rs`, `crates/shapes/src/model.rs`, `crates/shapes/src/lib.rs`.

- Add the W3C builtin IRI consts the projection actually consumes, alongside the existing
  `rdfs::SUB_CLASS_OF`: `rdfs::{LABEL, COMMENT, RANGE}` (standard W3C IRIs — referencing
  builtins mints no vocabulary). No `owl::DEPRECATED` — per-member deprecation is declined
  (see Enhancements Considered and Declined), so no unused const is introduced.
- Add `pub struct ValueVocab { abstract_individual_type: String }` with
  `pub fn new(abstract_individual_type: &str) -> Self` and an accessor. **No `Default`.**
  Module/struct doc cross-references `SliceVocab` / `BoxRoleVocab`, states the
  no-minted-IRI rule, and documents that `None`/absent = *feature-off*, not
  *missing-required-dependency*.
- Add `pub struct ValueVocabProjection<'a> { pub vocab: &'a ValueVocab, pub ontology: &'a RdfDataset }`.
- Add `pub fn compile_with_value_vocab(shapes, ns, projection: Option<&ValueVocabProjection<'_>>) -> CompiledSchema`;
  make `compile` delegate with `None`. Import `RdfDataset` and the `data::{native_quads, GraphFilter}`
  helpers + `model::rdf::TYPE`. Reuse the existing `pub fn local_name` (`json_schema.rs:211`).
- Re-export `ValueVocab` / `ValueVocabProjection` / `compile_with_value_vocab` from `lib.rs`
  (alongside `Namespaces`) for downstream ergonomics.
- At this task the `projection` argument is accepted and threaded but not yet consumed;
  **Task 2 lands the first observable behavior.** Add a compile-check test that
  `compile_with_value_vocab(&shapes, &ns, None)` returns byte-identical output to `compile(&shapes, &ns)`.

### After Task 1: commit (github-commit-agent) → push (github-push-agent) → wiring-verifier → post update to issue #94.

## Task 2: Standalone enum `$defs` + concept-scheme metadata (R1, R2, R3, R6, E1, E2, E3, two-surface parity, collision guard)

**Files:** `crates/shapes/src/json_schema.rs` (+ tests in-module).

- In `compile_with_value_vocab`, when `projection` is `Some`:
  1. Enumerate vocab classes: NamedNode subjects of `?V rdf:type <marker>` over the
     ontology∪shapes union, into a sorted `BTreeMap<String /*class IRI*/, String /*enum key*/>`.
     Enum key = `format!("{}Enum", local_name(v))`. Hard-fail (`ns.is_known`) any vocab
     class in an undeclared namespace.
  2. For each vocab class, enumerate members via the single `members_of` seam: NamedNode
     subjects of `?x rdf:type V`; blank-node subjects dropped **with a `LossRecord`**.
     For each member collect ALL `rdfs:label` / `rdfs:comment` values, **sort them
     canonically and pick the first** (never iteration-first — `native_quads` returns
     dataset/interner order, which is not canonical across differently-ordered loads).
     Sort members canonically by CURIE and dedup.
  3. Build each enum `$def` (E1/E2) — the `enum` keyword is ALWAYS present (headline):
     - empty members ⇒ `{ "enum": [], "x-enum-varnames": [], "$comment": … }` + `LossRecord`;
     - always ⇒ `{ "enum": [ {"@id":curie}, … ], "x-enum-varnames": [locals…], "$comment": "projection of an open value vocabulary (anchor, not a fence)" }`;
     - additionally, when any member has a doc ⇒ add `"x-enum-descriptions": [ <comment|label|""> … ]`.
     `x-enum-varnames` / `x-enum-descriptions` order aligns index-for-index with the sorted `enum`.
     **No `oneOf`, no `const`, no per-member object annotations** — the flat `enum` is never replaced.
  4. Register enum keys in `emitted_defs` (PASS 1) and **insert the enum `$defs` into
     `defs` after the `class_names` snapshot** (mirroring `Node`) so they are excluded
     from the `@type` discriminator but still flow into both surfaces via `openapi_doc`.
- Extend `assert_target_class_keys_are_unambiguous` (or a sibling guard invoked from
  `compile_with_value_vocab`) to register enum keys and hard-panic on: two vocab classes
  producing the same enum key (cross-namespace local-name twins); an enum key clashing
  with a class `$def` key.
- **Tests (executable acceptance checks on the production surface):**
  - Standalone: a marker vocab class with NO referencing property still yields
    `$defs["TermStabilityEnum"]` with its sorted `{"@id":...}` members.
  - Concept-scheme (E1): the `enum` keyword is present in EVERY case (metadata-bearing and
    metadata-free); a vocab whose members carry `rdfs:comment` also emits a parallel
    `x-enum-descriptions` aligned index-for-index to `enum`; NO `$def` ever emits `oneOf`/`const`.
  - Codegen names (E2): `x-enum-varnames` = sorted member local names, aligned to `enum` order.
  - Two-surface parity: the same enum `$def` (incl. `x-enum-*`) appears under both `$defs` and
    `components/schemas` (mirror `test_openapi_embeds_components_schemas`).
  - Determinism: `compile_with_value_vocab` run twice ⇒ byte-identical `schema_json`/`openapi_json`;
    AND (F3 cross-load determinism) a member with MULTIPLE `rdfs:label`/`rdfs:comment` values,
    loaded under two DIFFERENT triple orders, yields identical `x-enum-descriptions` (proves
    sort-then-pick, not insertion-order).
  - No-vocab parity: `None` output == `compile` output (byte-identical); AND
    marker-supplied-but-zero-matching-classes output == `compile` output (E4 regression guard).
  - Collision guard: `#[should_panic]` on a fixture with local-name-twin vocab classes.
  - Undeclared-namespace marker class: `#[should_panic]` (E3).
  - Empty-vocab: emits `enum: []` and a recorded loss. Blank-node member: dropped + loss recorded.
  - Fixtures use `example.org` (`meta:` primary, a declared non-primary vocab namespace).

### After Task 2: commit → push → wiring-verifier → post update to issue #94.

## Task 3: Referencing `$ref` from `sh:class` and `rdfs:range` (R4, cardinality, R6)

**Files:** `crates/shapes/src/json_schema.rs` (+ tests).

- Extend `Ctx` with `value_vocab_enums: BTreeMap<String,String>` (class IRI → enum key)
  and `predicate_ranges: BTreeMap<String,String>` (predicate IRI → enum key), both
  populated in `compile_with_value_vocab` from the vocab-class map plus a scan of
  `?P rdfs:range V` (V a vocab class) over the **ontology∪shapes union**, using
  `native_quads` + `model::rdfs::RANGE`.
- `Constraint::Class(V)` arm (`json_schema.rs:1015`): if `V` is a vocab class, emit
  `{"$ref": "#/$defs/{Local}Enum"}` as the scalar value schema (in place of the permissive
  node-ref), then let the existing cardinality wrapper (`json_schema.rs:1194`) preserve
  single-vs-array.
- Property emission for `Path::Predicate(P)`: apply precedence (E4) — a vocab-resolving
  `sh:class` wins; a non-vocab `sh:class` also wins (keeps its node-ref/`$ref`) and the
  range path is suppressed with a `LossRecord` if `P`'s `rdfs:range` pointed at a
  *different* vocab; only a property with no vocab-resolving `sh:class` and a
  `predicate_ranges[P]` emits the range-derived enum `$ref`. Thread the property's
  predicate IRI to the emitter (`compile_property` currently receives the compacted
  `key`; pass the enum-key hint or the raw predicate so it can look up).
- **Tests:**
  - `sh:class` → vocab: single-valued (`sh:maxCount 1`) ⇒ scalar `$ref`; multi-valued ⇒
    array wrapper around the `$ref` (mirror `test_object_property_uses_ref`).
  - `rdfs:range` → vocab on a property with NO `sh:class` ⇒ enum `$ref`; range declared
    **shapes-side** (not in the ontology dataset) still resolves (union-scan pin).
  - Precedence: property with non-vocab `sh:class` + a vocab `rdfs:range` keeps the
    `sh:class` behavior and records the conflict loss (no enum `$ref`).
  - No dangling refs: run `assert_no_dangling_defs_refs` / `every_ref_resolves` over the
    compiled schema.

### After Task 3: commit → push → wiring-verifier → post update to issue #94.

## Task 4: Acceptance proofs — round-trip + open-validator decoupling (R5, encoding AC)

**Files:** `crates/shapes/src/json_schema.rs` tests (+ `instance` / `engine` test entry points).

These are executable proofs of the two subtle acceptance criteria; the production
behavior already shipped in Tasks 2–3.

- **Round-trip (encoding AC):** build an instance whose vocab-valued property carries an
  anchor individual, project it with `instance::project_graph`, and assert it VALIDATES
  against the compiled schema's `enum` `$ref` using the in-module `validates(...)` (boon
  draft-2020-12) helper — covering both a metadata-free vocab and a metadata-bearing one
  (the `x-enum-*` extension arrays are annotations and MUST NOT affect validation; assert
  the metadata-bearing member still validates). A non-anchor value FAILS. This proves
  members use the `{"@id":curie}` encoding (not bare strings).
- **Open-validator decoupling (R5):** on an `rdfs:range`-associated **open** property
  (no `sh:class`), a value that is NOT an anchor individual CONFORMS to the live SHACL
  validator (`parse_shapes` + `validate_graphs`) yet is REJECTED by the enum schema —
  proving the validator stays open while the projection is closed. Additionally assert the
  parsed `Shapes` IR contains no `Constraint::In` introduced by the projection (no `sh:in`
  injection); `compile_with_value_vocab` takes `&Shapes` immutably, so this is
  structurally guaranteed and the test pins it.

### After Task 4: commit → push → wiring-verifier → post update to issue #94.

## Task 5: Own the pre-existing `sh:in`-over-IRI round-trip bug (E5, `.baseline` ownership)

**Files:** `crates/shapes/src/json_schema.rs` (`term_enum_value` + the `test_enum_from_sh_in` golden).

`.baseline`: "when you see it, you own it — fix it now." While designing this feature we
found that `term_enum_value` (`json_schema.rs:1333`) encodes a `sh:in` IRI member as a
**bare CURIE string**, while the instance projector emits IRI values as `{"@id":curie}`
objects (`instance.rs:164`). The code flags this as a known "encoding mismatch"
(`json_schema.rs:865`), but for the POSITIVE projection it is a real round-trip bug: a
valid instance `{"@id":"ex:red"}` does not match a `sh:in`-derived `enum: ["ex:red"]`, so
the instance fails its own schema.

- Fix `term_enum_value`: `Term::NamedNode(n) ⇒ json!({ "@id": ns.compact_iri(n.as_str()) })`
  (matching `project_value`); literals/blank/triple members are unchanged (they already
  match the projector's bare-string / typed-object forms).
- Update the `test_enum_from_sh_in` golden to assert `{"@id":…}` IRI members.
- **Executable proof:** add a round-trip test — an instance projected via
  `instance::project_graph` whose property has a `sh:in` list of IRIs VALIDATES against the
  compiled schema (it does not today). This closes the bug on the production surface, not just
  in the golden.

### After Task 5: commit → push → wiring-verifier → post update to issue #94.

## Task 6: Docs, gate, wasm (R6 hygiene, standing constraints)

**Files:** rustdoc in `crates/shapes/src/json_schema.rs`; no new deps/features.

- Rustdoc for `ValueVocab`, `ValueVocabProjection`, `compile_with_value_vocab`, and the
  enum-`$def` behavior — including the "anchor, not a fence" doctrine and the
  projection-only guarantee. **No GitHub issue/PR numbers in any doc** (per `.baseline`).
- Run the full gate in the worktree: `cargo fmt --all`, then `make check` (fmt + pedantic/
  nursery clippy warning-free + build + tests + hygiene incl. `check-no-features.py`),
  then `make wasm` (`wasm32-unknown-unknown` clean). Fix everything to green.

### After Task 6: commit → push → wiring-verifier → post update to issue #94.

## Task 7: Downstream coordination (R8 — GitHub only, NOT the repo)

Per `.baseline`, process-flow/coordination lives in GitHub, never in the tree.

- **Executable production-surface check (the GitHub API is the surface):** post a comment on
  `Blackcat-Informatics/gmeow-ontology#1408` via `gh issue comment` and **capture the returned
  comment URL**; the task is complete only when the comment URL is observed. The comment
  states that `purrdf-shapes` now provides native value-vocabulary enum-`$def` projection
  (`compile_with_value_vocab` + `ValueVocab`), so the downstream
  `crates/pipeline/src/stages/value_vocab.rs` enrichment can be retired against a purrdf pin
  bump, with a short migration note (marker IRI + ontology dataset are now inputs).
- **If `gh` cannot post to that cross-repo issue** (auth/access), this is NOT self-dischargeable
  by writing a PR-body note. STOP and put the binary to the user: "post #1408 coordination via
  an alternate path, or record explicit human sign-off to defer it." No implementer self-grant.

### After Task 7: (no repo commit) — the captured comment URL is recorded in the PR body.

1. Verify all changes are committed and pushed; run the Mechanical Deferral Scan over the
   branch diff and resolve any hit.
2. Create the PR:
   ```bash
   gh pr create --title "feat(shapes): project value vocabularies to enum \$defs (projection-only)" \
     --body "Closes #94

   ## Summary
   Native, projection-only derivation in purrdf-shapes: classes designated by a
   caller-supplied marker (ValueVocab) are projected to standalone {ClassLocal}Enum
   \$defs in both JSON Schema and OpenAPI, with referencing properties (sh:class /
   rdfs:range) emitting a \$ref. The live SHACL validator is never mutated — value
   vocabularies stay open-world (anchor, not a fence).

   ## Changes
   - ValueVocab caller-config + compile_with_value_vocab entry point (compile unchanged)
   - Standalone enum \$defs keyed {ClassLocal}Enum, members {\"@id\":curie}, sorted/deterministic
   - \$ref from sh:class and ontology rdfs:range references, cardinality preserved
   - Collision guard extended for enum keys; two-surface parity; round-trip + open-validator proofs

   ## Downstream coordination
   gmeow-ontology value_vocab.rs can be retired against a purrdf pin bump.
   "
   ```
3. Post this plan to the PR:
   ```bash
   ```
4. Post to BOTH the PR and issue #94:
5. Report: "Stage 1 complete. PR #<num> created for issue #94. Ready for /stage2 94".

## Verification (end-to-end)

- `cargo test -p purrdf-shapes json_schema` — all new tests green: standalone enum `$def`,
  `enum`-keyword-always + `x-enum-varnames`/`x-enum-descriptions`, two-surface parity,
  same-process AND cross-load determinism, collision panic, undeclared-namespace panic,
  empty-vocab loss, blank-node loss, `$ref` cardinality (scalar + array), `sh:class`/`rdfs:range`
  precedence + conflict loss, shapes-side `rdfs:range` union-scan, no-dangling-refs, round-trip
  validate (metadata-free AND metadata-bearing), open-validator decoupling, and the Task-5
  `sh:in`-over-IRI round-trip fix.
- `make check` clean (fmt + pedantic/nursery clippy warning-free + build + full tests +
  hygiene incl. `check-no-features.py`).
- `make wasm` clean (`wasm32-unknown-unknown`).
- Manual sanity: compile a small `example.org` fixture with a marker vocab class and
  confirm `TermStabilityEnum` appears in both `$defs` and `components/schemas` with a flat
  sorted `{"@id":...}` `enum` + aligned `x-enum-varnames`, and that `compile(...)` (no
  projection) output is byte-unchanged.
- Coordination (R8): the `gh issue comment` on gmeow-ontology#1408 returns a comment URL,
  recorded in the PR body.

## Execution Rules
1. **Task 0 is MANDATORY first** — never work on `main`; all work in the worktree.
2. **Commit after EACH task** (github-commit-agent) — never batch.
3. **Push after EACH commit** (github-push-agent).
4. **wiring-verifier after each implementation task.**
5. **Task N (PR) is MANDATORY last.**
6. Use sub-agents (Task tool) for all commits/pushes — never run git directly in main context.

paudley:
### Stage 1 reflection

1. **`rdfs:range` ↔ `sh:class` precedence.** When a property carries BOTH a *non-vocab* `sh:class` (structural) and a *vocab* `rdfs:range` (value set), I let `sh:class` win and record the range enum only as a `LossRecord` — the enum `$ref` is suppressed. That is the safe, deterministic choice, but a downstream ontology that deliberately models both could reasonably expect the enum to appear. It is visible only in `losses`, not the schema. If gmeow hits this, the rule may need revisiting (e.g. intersect rather than suppress).

2. **Membership is direct `rdf:type`.** I enumerate members as `?x rdf:type V` (the issue's "seeded individuals"). If gmeow's real data expresses membership via `skos:inScheme`/`skos:member` or purely via `rdfs:subClassOf`, the enum would come back empty. I localized enumeration behind one `members_of` seam so this is a one-line change if the assumption is wrong — but it IS an assumption worth confirming against gmeow's actual vocabulary encoding.

3. **Metadata predicates are fixed to `rdfs:label`/`rdfs:comment`.** `skos:prefLabel`/`skos:definition` are not read (declined: adds config surface). If gmeow documents members with SKOS predicates, `x-enum-descriptions` will be empty even though docs exist.

1. **The `sh:in`-over-IRI encoding changed.** I fixed a pre-existing bug: `sh:in` enum members over IRIs now emit `{"@id":curie}` objects (matching the projector) instead of bare CURIE strings. This is strictly more correct (projected instances now validate), but ANY consumer that hardcoded the old bare-string shape for an IRI-valued `sh:in` will see different output. No in-repo consumer does; a downstream one might.

2. **`x-enum-varnames` are raw local names, not de-duplicated across namespaces.** Two vocab members `ex1:red` and `ex2:red` (distinct CURIEs) both yield varname `"red"`. Harmless for JSON Schema, but a codegen tool that treats `x-enum-varnames` as unique identifiers could collide. Not observed in practice; flagging it.

3. **Retiring gmeow's `value_vocab.rs` needs a migration + `purrdf` pin bump** — the marker IRI and the ontology dataset are now explicit inputs to `compile_with_value_vocab`. Coordination comment posted on gmeow-ontology#1408.

paudley:
# Stage 2 Gap Analysis & Remediation — Issue #94 / PR #98

**Issue:** `shapes/json_schema: project logic:AbstractIndividualType value vocabularies to enum $defs (projection-only)`
**Branch:** `paudley/94-value-vocab-enum-defs` (HEAD `39623a1`) · **PR #98** · Worktree `.worktrees/94-value-vocab-enum-defs`

## Context

Why this stage exists: PR #98 implements value-vocabulary → enum `$def` projection in
`crates/shapes/src/json_schema.rs`. Three independent audits ran (issue/PR context, adversarial
gap-analysis, completion-adversary). The verdict is strong:

- **Completion-adversary: PASS.** Every acceptance criterion (R1–R8, plus the AC/E-list) was
  *demonstrated on the production surface* — the exported `purrdf_shapes::{compile_with_value_vocab,
  ValueVocab, ValueVocabProjection}` API, driven end-to-end (real Turtle → `from_dataset` →
  compile → round-trip through the real instance projector, the `boon` JSON-Schema validator, and
  the live SHACL engine). `cargo test -p purrdf-shapes --lib value_vocab` = 18/18; clippy clean.
- **No CRITICAL gaps.** Deferral keyword scan: **empty**. In-tree process-reference scan (`#NNNN`,
  `gmeow`, `1408`, `phase N`): **empty**. No stubs, no `unimplemented!`, no described-but-unbuilt
  requirement, no dark code (the value-vocab entry point is a published library API for the
  downstream `gmeow-ontology` consumer — the SUBSUME target — exactly like the existing `compile`).
- **`.goals`/`.baseline` clean:** no Cargo features, no minted vocabulary IRIs (marker is
  caller-supplied, hard-fails on undeclared namespace), fail-closed collision checks, losses
  recorded not swallowed, byte-deterministic.

What remains: **3 unaddressed reviewer nits (Gemini)** + **1 acceptance-criterion coverage gap**.
All are small, in-file, and get FIXED now (no deferrals). Each is committed + pushed separately.

## Gaps to remediate

### Gap A — [MEDIUM] `value_vocab_predicate_ranges` does O(N) dataset scans
`crates/shapes/src/json_schema.rs:608-639`. One `native_quads(range, class)` scan **per vocab
class**. Standing goal is MAXIMAL PERFORMANCE; the reviewer's single-scan-and-filter form is
strictly better (O(1) queries in the number of vocab classes).

**Fix:** Scan `(?P, rdfs:range, ?O)` **once per dataset** (object unfiltered), and for each
`?O` present in `vocab_enums`, map its predicate to the enum key.

**⚠ Determinism (R6 — hard constraint):** the current code resolves a predicate that ranges over
*multiple* vocab classes by last-write-wins over the **sorted** `vocab_enums` iteration (largest
`class_iri` wins). A naive "scan once, `out.insert`" rewrite would resolve that in native-scan
order — nondeterministic. The rewrite MUST carry an explicit tiebreak. Use an intermediate
`BTreeMap<predicate, (class_iri, enum_key)>` and, on conflict, keep the **larger `class_iri`**
(replicating current behavior exactly), then collapse to `BTreeMap<predicate, enum_key>`. Result is
provably order-independent. No behavioral change for the normal single-range case.

*Not a speedup claim:* this is an algorithmic-complexity/clarity simplification reasoned from code;
per repo convention we do NOT add a criterion bench or assert a measured speedup (schema compile is
a documented cold path). Correctness preserved by the existing `value_vocab_*` tests.

### Gap B — [MEDIUM] AC "class-key clashes" collision guard is untested
`crates/shapes/src/json_schema.rs:786-790`. The issue's AC lists **two** collision classes:
cross-namespace local-name twins *and* enum-key-vs-class-`$def`-key clashes. The local-name-twin
and undeclared-namespace guards have `#[should_panic]` tests; **this third guard has none**, so an
explicit acceptance criterion is only partially demonstrated.

**Fix:** Add a `#[should_panic(expected = "collides with a class")]` test: a shape whose
`sh:targetClass` local name yields a `$def` key equal to a vocab class's `{Local}Enum` key (e.g. a
target class literally named `ColorEnum` alongside a value-vocab class `Color`), compiled with the
projection, must hard-fail. Mirror the style of the existing twin/undeclared-ns fixtures.

### Gap C — [LOW] `first_literal` allocates a `BTreeSet` only to take the minimum
`crates/shapes/src/json_schema.rs:712-730`. Collecting every literal into a `BTreeSet<String>`
(cloning all, including non-minima) then `.next()` can be a single-pass running minimum: O(1) space,
no set allocation, no cloning of non-minimum strings.

**Fix:** Track a single `Option<String>`, replacing it only when a strictly smaller `lit.value()`
is seen; return it. Behavior-identical — "smallest by `Ord`" is order-independent, so determinism
(the `member_description` "canonically first" contract) is preserved. Existing
`description_selection_is_load_order_independent` test guards this.

### Gap D — [LOW] `assert_eq!` used for an always-failing guard reads confusingly
`crates/shapes/src/json_schema.rs:791-798`. `vocab_enums` is keyed by `class_iri` (unique), so when
`key_to_class.insert` returns `Some(prev)`, `prev` is *always* a different class than the current
one — the `assert_eq!(&prev, class_iri)` is a guaranteed failure whose `left == right` framing is
misleading.

**Fix:** Replace the `if let Some(prev) = … { assert_eq!(…) }` with a direct `panic!` carrying the
same descriptive message (twin classes share the enum `$def` key). The existing
`local_name_twins_hard_fail` `#[should_panic]` test continues to cover it.

### Gap E — [NO CHANGE, in-spec] `sh:class` + vocab `rdfs:range` conflict suppresses the enum `$ref`
`crates/shapes/src/json_schema.rs:~1624`. When a property carries both a non-vocab `sh:class` and a
vocab `rdfs:range`, `sh:class` wins and the range enum is recorded as a `LossRecord` (never
silent). Both the gap-analyzer and the completion-adversary agree this is **deterministic,
defensible, and not dictated by the issue** (#94 specifies projection, not conflict-resolution
policy). This is complete and loss-tracked — **no code change**; documented here for the record.

## Files touched
- `crates/shapes/src/json_schema.rs` — Gaps A, C, D (impl) and Gap B (test module in same file).

Not touched: `model.rs`, `instance.rs`, `lib.rs` (already correct per audits). No new deps, no
Cargo features, no generated/ or vectors/ edits.

## Execution (Phase 4 — one commit per gap, never batched)
For each of Gaps A → B → C → D, in order:
1. Dispatch a **general-purpose (sonnet)** sub-agent to implement the single fix, with the
   NO-DEFERRALS clause (a stub/downgrade/re-defer is a FAILED fix → re-dispatch).
2. **github-commit-agent** to commit (reads `PRECOMMIT_GUIDANCE.md`, iterates pre-commit to green).
3. **github-push-agent** to push (quality gates).
4. Post a per-gap summary (description, priority, fix, commit hash) to **issue #94 AND PR #98**.
5. **wiring-verifier** sub-agent to confirm integration.

## Verification (end-to-end)
- `cargo test -p purrdf-shapes --lib value_vocab` — the 18 value-vocab tests stay green; Gap B adds
  one new `#[should_panic]` and the count goes to 19.
- `cargo test -p purrdf-shapes --lib json_schema` — full json_schema suite green.
- `cargo clippy -p purrdf-shapes --all-targets` — warning-free (pedantic/nursery).
- **`make check`** (fmt + clippy + build + tests + hygiene) and **`make wasm`** — full-workspace
  gate green before declaring ready for /stage3 (closes the adversary's two advisory STOP-AND-ASK
  items: full-gate re-run; the R8 gmeow#1408 coordination remains correctly out-of-tree in GitHub).
- **Phase 6 exit gate:** re-run the completion-adversary (STAGE=2) against the new HEAD; it must
  return PASS before /stage3.

## Not gaps (adjudicated, for the record)
- **No production caller in-repo** — the value-vocab API is exported library surface consumed by the
  separate `gmeow-ontology` repo (the SUBSUME target); mirrors `compile`. Not dark. (adversary: PASS)
- **R8 downstream retirement** — coordination posted to `gmeow-ontology#1408`; process info stays in
  GitHub per `.baseline`, never in-tree. Correct as-is.
- **`sh:in`-over-IRI encoding fix (E5)** — the owned pre-existing bug; `term_enum_value` now
  delegates to `instance::project_value`. Golden updated, round-trip test added. Intended change.

paudley:
### ✅ Gap A remediated — [MEDIUM] O(N)→O(1) `rdfs:range` scan for value-vocab `$ref` mapping

**Priority:** MEDIUM (unaddressed Gemini review + MAXIMAL PERFORMANCE goal)
**Commit:** `b4c0f44` — `perf(shapes): scan rdfs:range once per dataset for value-vocab $ref mapping`
**File:** `crates/shapes/src/json_schema.rs` — `value_vocab_predicate_ranges`

**Fix:** Replaced the per-vocab-class `(?P, rdfs:range, class)` scan (one query per vocab class) with a single `(?P, rdfs:range, ?O)` scan per dataset, filtering matched objects against the `vocab_enums` map by IRI. Query count drops from O(N) in the number of vocab classes to O(1) per dataset.

**Determinism preserved (R6):** the prior code resolved a predicate ranging over multiple vocab classes by last-write-wins over the sorted `vocab_enums` iteration (largest `class_iri` wins). The rewrite replicates that exactly via an intermediate `winners` map with an explicit largest-`class_iri` tiebreak, so `$ref` resolution stays byte-deterministic regardless of native scan order.

**Verification:** `cargo test -p purrdf-shapes --lib value_vocab` 18/18 · `cargo clippy -p purrdf-shapes --all-targets -- -D warnings` clean · `cargo fmt --check` clean. Behavior-identical; no bench/speedup asserted (cold-path compile).

paudley:
### ✅ Gap B remediated — [MEDIUM] AC "class-key clashes" collision guard now tested (+ tiebreak regression)

**Priority:** MEDIUM (explicit acceptance-criterion coverage: collision-checking includes class-key clashes, not just local-name twins)
**Commit:** `b07d9c3` — `test(shapes): cover value-vocab class-key clash guard and multi-range tiebreak`
**File:** `crates/shapes/src/json_schema.rs` (test module, +64, no production change)

Two regression tests added:

1. **`value_vocab_enum_key_class_def_key_clash_hard_fails`** — `#[should_panic(expected = "collides with a class \`$def\` key")]`. Covers the previously-untested third fail-closed guard in `assert_value_vocab_enum_keys_unambiguous`: a vocab class `logic:Color` (enum key `ColorEnum`) alongside a target `sh:NodeShape` on `meta:ColorEnum` (`$def` key `ColorEnum`) hard-fails at compile. Verified it trips the *class-key-clash* guard specifically — not the twin or undeclared-namespace guards (distinct messages confirmed).

2. **`value_vocab_multi_range_picks_largest_class_iri_deterministically`** — guards the determinism tiebreak introduced by Gap A. A predicate with `rdfs:range` over two vocab classes (`logic:Alpha`, `logic:Beta`) deterministically resolves its `$ref` to `#/$defs/BetaEnum` (largest class IRI). Asserted across *reversed* triple-declaration order to prove the tiebreak is by-value, not scan/load-order dependent.

**Verification:** `cargo test -p purrdf-shapes --lib value_vocab` 20/20 (was 18) · clippy `-D warnings` clean · fmt clean.

paudley:
### ✅ Gaps C & D remediated — [LOW] `first_literal` allocation + `assert_eq!` readability

Both are the remaining unaddressed Gemini review nits.

**Gap C — `first_literal` single-pass minimum** · commit `8fd5890` — `perf(shapes): compute first_literal via single-pass running minimum`
Replaced the `BTreeSet<String>` collect-then-`.next()` (which allocated a set and cloned every candidate) with a single `Option<String>` running minimum, cloning only when a strictly smaller value is found. Semantics identical — still the lexicographically smallest literal (the deterministic "canonically first" contract); order-independent.

**Gap D — direct `panic!` for the enum-key twin guard** · commit `10c1f1b` — `refactor(shapes): panic! directly for value-vocab enum-key twin guard`
`vocab_enums` is keyed by unique `class_iri`, so reaching the twin branch guarantees `prev != class_iri` — the `assert_eq!`'s `left == right` framing was misleading. Replaced with an unconditional `panic!` carrying the identical message. No behavior change; `value_vocab_local_name_twins_hard_fail` `#[should_panic]` still covers it.

**Verification (both):** `cargo test -p purrdf-shapes --lib value_vocab` 20/20 · clippy `-D warnings` clean · fmt clean.

---

All four gaps (A–D) are now fixed and pushed. Gap E (`sh:class` + vocab `rdfs:range` conflict → enum suppressed with recorded loss) is confirmed **in-spec / no change** by both the gap-analyzer and the completion-adversary — the issue specifies projection, not conflict-resolution policy, and the loss is recorded, never silent. Full-workspace `make check` + `make wasm` gate and the completion-adversary exit re-run follow.

paudley:
## 🏁 Stage 2 complete — ready for Stage 3

All gaps remediated; the completion-adversary exit gate returns **PASS** (every acceptance criterion DEMONSTRATED-IN-PRODUCTION on the exported `purrdf_shapes::{compile_with_value_vocab, ValueVocab, ValueVocabProjection}` library API), and the full-workspace gate is green.

### Gaps addressed (one commit each, all pushed)
| Gap | Priority | Fix | Commit |
|-----|----------|-----|--------|
| A | MEDIUM | `value_vocab_predicate_ranges`: O(N)-per-class scans → single `(?P, rdfs:range, ?O)` scan per dataset, filtered against `vocab_enums`; O(1) queries. Determinism preserved via explicit largest-`class_iri` tiebreak. | `b4c0f44` |
| B | MEDIUM | Test coverage for the **class-`$def`-key clash** collision guard (an explicit AC alongside local-name twins) + a determinism regression test for the multi-range tiebreak (forward & reversed order). | `b07d9c3` |
| C | LOW | `first_literal`: `BTreeSet` collect-then-`.next()` → single-pass running `Option<String>` minimum; identical smallest-by-`Ord` (deterministic) semantics, no set allocation. | `8fd5890` |
| D | LOW | `assert_value_vocab_enum_keys_unambiguous`: always-failing `assert_eq!` → direct `panic!` (operands are provably never equal when reached); message unchanged. | `10c1f1b` |

Gaps A, C, D are the three previously-unaddressed Gemini review nits; B closes a real acceptance-criterion coverage gap the reviewers did not catch. **Gap E** (`sh:class` + vocab `rdfs:range` conflict → enum suppressed with a recorded `LossRecord`) is confirmed **in-spec / no change** by both the gap-analyzer and the adversary: #94 specifies projection, not conflict-resolution policy, and the loss is recorded, never silent.

### Reviewer feedback
- **Gemini** (3 medium/low perf+style nits): all 3 addressed (Gaps A, C, D). ✅
- **CodeRabbit**: no actionable comments. ✅
- No human-reviewer comments; no new feedback since remediation.

### Completion-adversary exit re-run — PASS
Every criterion (R1–R8, AC1–AC4, E1–E5) carries an executed production-surface demonstration: real Turtle → `parse` → `from_dataset` → `compile_with_value_vocab` → round-trip through the real `instance::project_subject` projector, the `boon` JSON-Schema validator, and the live `engine::validate_dataset` SHACL engine. The adversary rebuilt the pre-remediation baseline (`39623a1`) and confirmed the four remediation commits are semantics-preserving — no existing byte-pinned assertion was altered, so observable output is unchanged from the previously-certified HEAD. Determinism (R6) specifically re-demonstrated for both perf rewrites (`description_selection_is_load_order_independent`, `multi_range_picks_largest_class_iri_deterministically`).

### Verification
- `make check` + `make wasm` + `rdf-core-hygiene`: **exit 0, clean** (98 test-result groups, 0 failures).
- `cargo test -p purrdf-shapes --lib value_vocab`: **20/20** · `--lib json_schema`: **64/64**.
- `cargo clippy -p purrdf-shapes --all-targets -- -D warnings`: clean · doctests: clean.
- Deferral / laundering / in-tree process-ref scans: **all empty**.
- No Cargo features, no minted vocabulary IRIs, wasm32-clean, byte-deterministic.

**Ready for /stage3.**

paudley:
feat(shapes): project value vocabularies to enum $defs (projection-only)

Projects each caller-marked value-vocabulary class (e.g. an
`AbstractIndividualType` marker) to a standalone JSON-Schema / OpenAPI enum
`$def`, and resolves referencing `sh:class` / `rdfs:range` properties to a
`$ref` — a deterministic, projection-only derivation that never mutates the
live SHACL validator (anchor, not a fence). This is the Rust-native, in-library
SUBSUME of the downstream ad-hoc value-vocab derivation.

## What this branch delivers

New public API on `purrdf-shapes` (re-exported from `crates/shapes/src/lib.rs`):
`compile_with_value_vocab`, `ValueVocab`, `ValueVocabProjection`. The marker
class and backing ontology are caller-supplied — PurRDF mints no vocabulary
IRIs and fabricates no default; absent config is byte-identical to `compile()`.

Per marker vocab class it emits `{ClassLocal}Enum` whose `enum` members are the
sorted, de-duplicated named individuals encoded as `{"@id": curie}` — matching
the instance projector's encoding exactly, so a projected instance round-trips
through the enum under a real JSON-Schema validator. The flat `enum` keyword is
always load-bearing; member metadata rides parallel `x-enum-varnames`
(sorted local names) and `x-enum-descriptions` (canonically-first
`rdfs:comment` → `rdfs:label`). The `$def` is emitted even when unreferenced,
appears byte-identically in both JSON-Schema `$defs` and OpenAPI
`components/schemas`, and is fully build-time collision-checked (fail-closed):
cross-namespace local-name twins, enum-key-vs-class-`$def`-key clashes, and
undeclared-namespace marker classes each hard-fail. Blank-node members, empty
vocabularies, and `sh:class`/`rdfs:range` conflicts are recorded as
`LossRecord`s — never silent.

Owned pre-existing bug fixed (`.baseline` "own it"): `term_enum_value` now
delegates to `instance::project_value`, so IRI-valued `sh:in` members emit
`{"@id":curie}` (was bare CURIE strings) — one encoding source of truth.

## Files (final PR diff, `git diff origin/main...HEAD`)

- `crates/shapes/src/json_schema.rs` (+1180) — projection engine, enum `$def`
  builder, `$ref` resolution, collision guards, and the test module.
- `crates/shapes/src/instance.rs` (±7) — `project_value` now shared with the
  enum-member encoder.
- `crates/shapes/src/model.rs` (+10) — `rdfs::{LABEL,COMMENT,RANGE}` consts.
- `crates/shapes/src/lib.rs` (±2) — re-exports.

## Stage-2 remediation (reviewer + gap-analysis findings, all fixed)

- perf: `value_vocab_predicate_ranges` rewritten from O(N)-per-vocab-class
  dataset scans to a single `(?P, rdfs:range, ?O)` scan per dataset filtered
  against the enum map — O(1) queries. A predicate ranging over multiple vocab
  classes still resolves deterministically to the largest `class_iri` via an
  explicit tiebreak (no determinism regression).
- test: added coverage for the class-`$def`-key clash guard (an explicit
  acceptance criterion previously untested) and a determinism regression test
  for the multi-range tiebreak (asserted across reversed declaration order).
- perf: `first_literal` computes a single-pass running minimum instead of
  collecting a `BTreeSet` — same smallest-by-`Ord` deterministic semantics.
- refactor: the always-failing enum-key twin guard uses a direct `panic!`
  instead of a misleading `assert_eq!`.

## `.goals` standing constraints — respected

GREENFIELD Rust-native derivation (no compromise carried from the downstream
workaround); RUST-FIRST (Python surface untouched); SUBSUME/EXTEND/ENHANCE (the
in-library projector supersedes the ad-hoc downstream one); LOW/NO OPTIONALITY /
HARD FAILS (caller-supplied marker with no default; fail-closed collision and
undeclared-namespace guards; losses recorded, never swallowed; no Cargo
features); MAXIMAL PERFORMANCE (O(1) range scan, single-pass min); MAXIMAL
PORTABILITY (wasm32-clean, `make wasm` green); byte-deterministic serialization.

## Merge with `origin/main` (PR #99 landed mid-stage)

`origin/main` advanced to include PR #99 (native sparql-eval user-function
registry) while this branch was in finalization. `origin/main` was merged into
the branch (merge commit). The only conflict was in
`crates/gts/src/compact.rs` and `crates/rdf/src/gts_certify.rs`: both main and
this branch had independently fixed the same pre-existing rustdoc doc-gate
failure (public docs linking to private items, denied under
`RUSTDOCFLAGS="-D warnings"`). Each conflict was evaluated individually and
resolved to main's version — the superior fix, which additionally removed a
stray process-flow doc reference. Net result: those two files carry ZERO delta
from this branch (`git diff origin/main -- <both>` empty), and the final PR diff
is exactly #94's four shapes files. #94's shapes deliverable is byte-identical
before and after the merge (`git diff <pre-merge>...HEAD -- crates/shapes/`
empty). The merged tree passes `make check` (exit 0) and `make wasm`; #94 and
#99 coexist (sparql-eval 335/335, shapes value_vocab 20/20, json_schema 64/64).

## Completion-Adversary (STAGE 3, post-merge) — PASS

Audited the merge-into-main result; every acceptance criterion
DEMONSTRATED-IN-PRODUCTION on the exported `purrdf_shapes` API (real Turtle →
`from_dataset` → `compile_with_value_vocab` → real `instance::project_subject` +
`boon` JSON-Schema validator + live `engine::validate_dataset` SHACL engine).
Per-criterion demonstrations (all `cargo test -p purrdf-shapes` on HEAD):

- R1/R2/R3 standalone sorted `{ClassLocal}Enum`, reachable unreferenced —
  `value_vocab_standalone_enum_def_even_when_unreferenced` ok.
- R4 `sh:class`/`rdfs:range` → `$ref`, cardinality preserved —
  `..sh_class_single_emits_scalar_ref`, `..sh_class_multi_wraps_ref_in_array`,
  `..rdfs_range_open_property_emits_ref`,
  `..rdfs_range_declared_shapes_side_still_resolves` ok.
- R5 projection-only, validator stays open —
  `value_vocab_projection_leaves_the_live_validator_open` ok (drives
  `engine::validate_dataset`; asserts no `Constraint::In` in the IR).
- R6 determinism — `..output_is_byte_deterministic`,
  `..description_selection_is_load_order_independent`,
  `..multi_range_picks_largest_class_iri_deterministically` ok.
- R7 opt-in, None/zero-match == plain compile —
  `value_vocab_none_and_zero_match_equal_plain_compile` ok.
- AC {"@id":curie} round-trip —
  `value_vocab_projected_instance_round_trips_through_enum` ok.
- AC two-surface parity — `value_vocab_enum_def_appears_in_both_surfaces` ok.
- AC collision guards (twins / class-key clash / undeclared-ns) — three
  `#[should_panic]` tests ok.
- E1/E2 load-bearing enum + parallel metadata —
  `..enum_is_load_bearing_with_parallel_metadata_arrays`,
  `..metadata_free_omits_descriptions` ok.
- E4 losses recorded — `..blank_node_member_is_dropped_with_loss`,
  `..empty_records_loss_and_empty_enum`, `..sh_class_wins_over_conflicting_range` ok.
- E5 `sh:in`-over-IRI fix — `sh_in_over_iris_uses_id_object_members`,
  `sh_in_over_iris_round_trips_through_projection` ok.

Open findings: none.

## Deferral gate — clean

Mechanical deferral scan over `git diff origin/main...HEAD` added lines: no
hits. In-tree process-ref scan (`#NNNN`/gmeow/1408/phase N/per #) on added
lines: no hits. PR-body "hits" are anti-deferral affirmations
("NOTHING is deferred, stubbed, or left for a future PR") and CodeRabbit's
"Out of Scope check ✅ Passed" — false positives.

## R8 downstream retirement — out-of-tree by design (not a repo deliverable)

Retiring the downstream `gmeow-ontology` value-vocab derivation is a cross-repo
migration; #94's repo deliverable is the exported derivation API (the SUBSUME
target) plus the absence of in-tree process references. Coordination is tracked
in GitHub, which is the correct home for process flow.

## Verification summary

`make check` exit 0 (98 test groups, 0 failures) · `make wasm` clean ·
`RUSTDOCFLAGS="-D warnings" cargo doc --workspace` exit 0 ·
`cargo clippy -p purrdf-shapes --all-targets -- -D warnings` clean · value_vocab
20/20 · json_schema 64/64 · sparql-eval 335/335 · doctests clean · no Cargo
features · no minted vocabulary IRIs · byte-deterministic.

PR #99, opened 2026-07-13
feat(sparql-eval): native Rust-closure user-function registry

### LINKED ISSUE #91: sparql-eval: native (Rust-closure) user-function registry
## Context

`UserFunctionRegistry` currently accepts SHACL-AF style functions whose bodies are parsed SPARQL (`sh:select`/`sh:ask`). The consumer's deductive recall layer (ADR 0030) needs scoring predicates inside queries — BM25 scores, vector-similarity, phonetic name-similarity — whose implementations are native Rust backed by external index structures, which cannot be expressed as SPARQL bodies.

## Request

A registration seam for native functions: `register_native(iri, Arc<dyn Fn(&[TermValue]) -> Result<TermValue> + Send + Sync>)` (or equivalent) dispatched from `Function::Custom` evaluation, with the existing pure-SPARQL registry unchanged. Determinism/purity expectations should be part of the contract (documented, and ideally declared at registration) so the parallel-evaluation determinism gate can account for them.

## Interim fallback in use

Pre-binding: The downstream consumer evaluates index relations outside SPARQL and materializes scored candidate bindings into a scratch world before evaluation. Works, but prevents push-down (scoring can't participate in FILTER/ORDER BY inside the engine).

paudley:
Closes #91

## Summary
Adds a native (Rust-closure) registration seam to `UserFunctionRegistry`, dispatched from the same `Function::Custom` eval site as SPARQL-bodied SHACL-AF functions, with a determinism (`Volatility`) declaration wired into the fork-join parallel-safety gate so native scores participate in `FILTER`/`ORDER BY` and get engine push-down (the motivating need for the consumer's deductive-recall scorers: BM25, vector-similarity, phonetic).

## Changes
- **user_fn.rs** — `NativeFnBody = Arc<dyn Fn(&[TermValue]) -> Result<TermValue, EvalError> + Send + Sync>`; `Volatility {Stable, Volatile}` (`#[non_exhaustive]`); `Arity {Exact, Range, AtLeast}`; `NativeFunction`. `UserFunctionRegistry` gains a native table + `register_native`/`resolve_native` with a **cross-kind IRI collision guard** (an IRI can't be both native and SPARQL-bodied). `eval_native_function`: fail-fast arity check → unbound arg yields no value → closure invoked under `catch_unwind` (panic → deterministic `EvalError`, no nondeterministic rayon abort).
- **expr.rs** — native-resolution branch in the `Function::Custom` arm.
- **parallel.rs** — `is_parallel_safe`/`is_parallel_safe_pattern` are now registry-aware; a `Volatile` native fn, and **any** SPARQL-bodied user fn, are parallel-unsafe; a `Stable` native fn stays safe. This also **fixes a pre-existing latent nondeterminism**: SPARQL-bodied UDFs with `RAND`/`UUID`/`BNODE`/list-constructor bodies were previously run under fork-join, merging per-query mint/RNG state into a discarded forked child.
- **error.rs** — `EvalError::Function` Display generalized from 'SHACL-AF function error' to 'user function error'.
- **lib.rs** — re-export `Arity`, `NativeFnBody`, `NativeFunction`, `Volatility`.
- **tests** — FILTER/ORDER BY push-down, ORDER BY determinism, NaN stable-total-order (no rows dropped), a >1024-row fork-join FILTER (parallel-safety at scale), plus arity/panic/collision/coexistence unit tests and 4 parallel-gate classification tests.

## Verification
- `make check` (fmt + clippy pedantic/nursery + build + tests + hygiene): **green**
- `make wasm` (wasm32 cleanliness): **green**
- No new Cargo features, no dependency pins, deterministic, `example.org` fixtures only.

### DIFF
--- a/crates/gts/src/compact.rs
+++ b/crates/gts/src/compact.rs
@@ -250,8 +250,8 @@ fn base64url_unpadded(data: &[u8]) -> String {
 }
 
 /// Decode a base64url WITHOUT padding (RFC 4648 §5) string — the inverse of
-/// [`base64url_unpadded`], used to recover a `stream:cose` literal's raw
-/// COSE_Sign1 bytes (Task 5 signature verification).
+/// `base64url_unpadded`, used to recover a `stream:cose` literal's raw
+/// COSE_Sign1 bytes for signature verification.
 ///
 /// # Errors
 /// HARD-ERRORS on a padding character (`=`) or any byte outside the unpadded
--- a/crates/rdf/src/gts_certify.rs
+++ b/crates/rdf/src/gts_certify.rs
@@ -777,7 +777,7 @@ fn suppressions_ok(pre: &Graph, post: &Graph) -> Result<bool, CertifyError> {
 /// `pre_bytes`/`post_bytes` are UNTRUSTED (an independent verifier's whole
 /// point is to not trust the claimant), so every digest this function
 /// computes over them goes through the fallible RDFC-1.0 canonicalizer
-/// ([`try_refold_digest`]/[`try_effective_digest`], backing
+/// (`try_refold_digest`/`try_effective_digest`, backing
 /// [`try_canonicalize_with`]) rather than the panicking
 /// [`canonicalize_with`]: a pathologically symmetric blank graph that stays
 /// under [`POISON_BLANK_LIMIT`]'s cheap blank-COUNT pre-reject yet exhausts
--- a/crates/sparql-eval/src/binop.rs
+++ b/crates/sparql-eval/src/binop.rs
@@ -148,8 +148,8 @@ pub(crate) fn eval_union<D: DatasetView + Sync>(
     right: &GraphPattern,
     ctx: &mut EvalCtx<'_, D>,
 ) -> Result<SolutionSeq<D::Id>, EvalError> {
-    if !crate::parallel::is_parallel_safe_pattern(left)
-        || !crate::parallel::is_parallel_safe_pattern(right)
+    if !crate::parallel::is_parallel_safe_pattern(left, ctx.user_functions)
+        || !crate::parallel::is_parallel_safe_pattern(right, ctx.user_functions)
     {
         let l = eval(left, ctx)?;
         let r = eval(right, ctx)?;
@@ -473,7 +473,7 @@ fn left_outer_join_filtered<D: DatasetView + Sync>(
     let shared = l.schema.shared_columns(&r.schema);
 
     // A left outer join emits at least one row per left row.
-    let rows = if crate::parallel::is_parallel_safe(expr) {
+    let rows = if crate::parallel::is_parallel_safe(expr, ctx.user_functions) {
         crate::parallel::par_chunk_try_map_init(
             &l.rows,
             || ctx.fork_for_worker(),
--- a/crates/sparql-eval/src/error.rs
+++ b/crates/sparql-eval/src/error.rs
@@ -49,11 +49,13 @@ pub enum EvalError {
     /// looping forever or guessing an answer.
     Data(String),
 
-    /// A SHACL-AF SPARQL-based function (`sh:SPARQLFunction`) call was invalid: an
-    /// arity mismatch, a `sh:datatype`/`sh:nodeKind`/`sh:returnType` violation, or
-    /// exceeding the user-function recursion bound. Per the hard-fail doctrine a
-    /// mis-invoked function aborts the query rather than yielding a wrong or unbound
-    /// value.
+    /// A user function call was invalid — either a SHACL-AF SPARQL-based function
+    /// (`sh:SPARQLFunction`: an arity mismatch, a
+    /// `sh:datatype`/`sh:nodeKind`/`sh:returnType` violation, or exceeding the
+    /// user-function recursion bound) or a native (host-Rust closure) function
+    /// (an arity mismatch, the closure's own returned `Err`, or a caught panic
+    /// inside the closure). Per the hard-fail doctrine a mis-invoked function
+    /// aborts the query rather than yielding a wrong or unbound value.
     Function(String),
 }
 
@@ -94,7 +96,7 @@ impl core::fmt::Display for EvalError {
             Self::Internal(msg) => write!(f, "internal evaluator error: {msg}"),
             Self::Remote(msg) => write!(f, "SERVICE federation error: {msg}"),
             Self::Data(msg) => write!(f, "malformed RDF input: {msg}"),
-            Self::Function(msg) => write!(f, "SHACL-AF function error: {msg}"),
+            Self::Function(msg) => write!(f, "user function error: {msg}"),
         }
     }
 }
--- a/crates/sparql-eval/src/expr.rs
+++ b/crates/sparql-eval/src/expr.rs
@@ -186,7 +186,7 @@ pub(crate) fn eval_filter<D: DatasetView + Sync>(
 ) -> Result<SolutionSeq<D::Id>, EvalError> {
     let seq = eval(inner, ctx)?;
     let schema = seq.schema.clone();
-    let rows = if crate::parallel::is_parallel_safe(expr) {
+    let rows = if crate::parallel::is_parallel_safe(expr, ctx.user_functions) {
         crate::parallel::par_chunk_try_map_init(
             &seq.rows,
             || ctx.fork_for_worker(),
@@ -231,7 +231,7 @@ pub(crate) fn eval_extend<D: DatasetView + Sync>(
     let width = schema.len();
     let schema = Arc::new(schema);
 
-    let rows = if crate::parallel::is_parallel_safe(expr) {
+    let rows = if crate::parallel::is_parallel_safe(expr, ctx.user_functions) {
         // Parallel path: `is_parallel_safe` excludes `BNODE` (every arity), so the
         // per-solution `BNODE(strExpr)` memo (`ctx.current_row`/`ctx.bnode_memo`) is
         // never observed here — no per-row `current_row` bookkeeping is needed.
--- a/crates/sparql-eval/src/expr.rs
+++ b/crates/sparql-eval/src/expr.rs
@@ -1651,6 +1651,18 @@ fn eval_function<D: DatasetView + Sync>(
                 let result = crate::user_fn::eval_user_function(func, iri.as_str(), &vals, ctx)?;
                 return Ok(result.map(|value| intern(ctx, value)));
             }
+            // A caller-injected native (host-Rust closure) function, resolved from
+            // the same registry's second table. Checked after the SPARQL-bodied
+            // path (so a same-registry cross-kind collision can never arise — the
+            // registry's collision guard already makes that unrepresentable) and
+            // before the XSD-cast fallback, so a function IRI never collides with a
+            // datatype IRI.
+            if let Some(registry) = ctx.user_functions
+                && let Some(native) = registry.resolve_native(iri.as_str())
+            {
+                let result = crate::user_fn::eval_native_function(native, iri.as_str(), &vals)?;
+                return Ok(result.map(|value| intern(ctx, value)));
+            }
             if let Some(target) = XsdDatatype::from_iri(iri.as_str()) {
                 return Ok(eval_xsd_cast(ctx, target, arg(&vals, 0)));
             }
--- a/crates/sparql-eval/src/lib.rs
+++ b/crates/sparql-eval/src/lib.rs
@@ -89,7 +89,8 @@ pub use scratch::{ScratchId, ScratchInterner, SolutionTerm};
 pub use solution::{Solution, SolutionSeq, VarSchema, compatible};
 pub use update::GraphResolver;
 pub use user_fn::{
-    NodeKind, TypeConstraint, UserFnBody, UserFnParam, UserFunction, UserFunctionRegistry,
+    Arity, NativeFnBody, NativeFunction, NodeKind, TypeConstraint, UserFnBody, UserFnParam,
+    UserFunction, UserFunctionRegistry, Volatility,
 };
 
 /// A deterministic, seed-free hasher builder (`AHasher` with fixed keys).
--- a/crates/sparql-eval/src/modifier.rs
+++ b/crates/sparql-eval/src/modifier.rs
@@ -526,7 +526,7 @@ pub(crate) fn eval_group<D: DatasetView + Sync>(
     let safe = aggregates.iter().all(|(_, agg)| match agg {
         AggregateExpression::CountStar { .. } => true,
         AggregateExpression::FunctionCall { expression, .. } => {
-            crate::parallel::is_parallel_safe(expression)
+            crate::parallel::is_parallel_safe(expression, ctx.user_functions)
         }
     });
 
--- a/crates/sparql-eval/src/parallel.rs
+++ b/crates/sparql-eval/src/parallel.rs
@@ -44,7 +44,12 @@
 //! is excluded, because the fork model gives every worker an *independent* copy of
 //! that state rather than a shared, ordered one — running such a builtin under
 //! fork-join would make its result depend on worker scheduling, not just row
-//! content.
+//! content. [`Function::Custom`] — a caller-injected user function resolved
+//! against a [`UserFunctionRegistry`] — is likewise excluded unless the registry
+//! attests the callee is safe: a native function is safe only when registered
+//! [`Volatility::Stable`], and a SPARQL-bodied function is conservatively always
+//! unsafe (its body can itself reach `RAND`/`UUID`/`BNODE`/the list constructors,
+//! and that per-call state would merge into a forked child instead of `ctx`).
 
 use purrdf_core::{DatasetView, TermId, TermValue, ViewTermId};
 use purrdf_sparql_algebra::{
@@ -54,6 +59,7 @@ use purrdf_sparql_algebra::{
 use crate::error::EvalError;
 use crate::scratch::{ScratchInterner, SolutionTerm};
 use crate::solution::Solution;
+use crate::user_fn::{UserFunctionRegistry, Volatility};
 
 /// Rows/groups at or below this stay sequential (thread spin-up would dominate
 /// the work for small inputs). Initial value; Task 7's bench tunes it.
@@ -193,22 +199,32 @@ pub(crate) fn should_parallelize(work_items: usize) -> bool {
 /// `listIndexOf`/`listContains`) and `heldIn` touch neither counter, so they are
 /// left safe. When in doubt this walk flags UNSAFE — a sequential fallback is
 /// always a correct (if slower) answer.
-pub(crate) fn is_parallel_safe(expr: &Expression) -> bool {
-    !expr_reaches_unsafe_builtin(expr)
+///
+/// `registry` is the caller's [`UserFunctionRegistry`] (`ctx.user_functions`), if
+/// any: it is consulted only when the walk reaches a [`Function::Custom`] call —
+/// see [`function_is_unsafe`] for exactly how a native-vs-SPARQL-bodied callee is
+/// classified. `None` (no registry configured) makes every `Custom` IRI resolve
+/// to the deterministic XSD-cast/hard-error fallback, hence safe.
+pub(crate) fn is_parallel_safe(expr: &Expression, registry: Option<&UserFunctionRegistry>) -> bool {
+    !expr_reaches_unsafe_builtin(expr, registry)
 }
 
 /// Whether `pattern` (recursively) is safe to evaluate under the fork-join
 /// parallel model — the pattern-level twin of [`is_parallel_safe`], for callers
 /// (e.g. `UNION`) that must gate a whole sub-pattern rather than a single
 /// expression. Exposes the same walk [`is_parallel_safe`] already runs
-/// internally for `EXISTS`.
-pub(crate) fn is_parallel_safe_pattern(pattern: &GraphPattern) -> bool {
-    !pattern_reaches_unsafe_builtin(pattern)
+/// internally for `EXISTS`. `registry` is threaded through exactly as in
+/// [`is_parallel_safe`].
+pub(crate) fn is_parallel_safe_pattern(
+    pattern: &GraphPattern,
+    registry: Option<&UserFunctionRegistry>,
+) -> bool {
+    !pattern_reaches_unsafe_builtin(pattern, registry)
 }
 
 /// `true` iff `expr` (recursively) reaches an unsafe builtin — see
 /// [`is_parallel_safe`].
-fn expr_reaches_unsafe_builtin(expr: &Expression) -> bool {
+fn expr_reaches_unsafe_builtin(expr: &Expression, registry: Option<&UserFunctionRegistry>) -> bool {
     match expr {
         Expression::NamedNode(_)
         | Expression::Literal(_)
@@ -226,30 +242,55 @@ fn expr_reaches_unsafe_builtin(expr: &Expression) -> bool {
         | Expression::Subtract(a, b)
         | Expression::Multiply(a, b)
         | Expression::Divide(a, b) => {
-            expr_reaches_unsafe_builtin(a) || expr_reaches_unsafe_builtin(b)
+            expr_reaches_unsafe_builtin(a, registry) || expr_reaches_unsafe_builtin(b, registry)
         }
         Expression::UnaryPlus(a) | Expression::UnaryMinus(a) | Expression::Not(a) => {
-            expr_reaches_unsafe_builtin(a)
+            expr_reaches_unsafe_builtin(a, registry)
         }
         Expression::In(head, list) => {
-            expr_reaches_unsafe_builtin(head) || list.iter().any(expr_reaches_unsafe_builtin)
+            expr_reaches_unsafe_builtin(head, registry)
+                || list
+                    .iter()
+                    .any(|e| expr_reaches_unsafe_builtin(e, registry))
         }
         Expression::If(a, b, c) => {
-            expr_reaches_unsafe_builtin(a)
-                || expr_reaches_unsafe_builtin(b)
-                || expr_reaches_unsafe_builtin(c)
+            expr_reaches_unsafe_builtin(a, registry)
+                || expr_reaches_unsafe_builtin(b, registry)
+                || expr_reaches_unsafe_builtin(c, registry)
         }
-        Expression::Coalesce(list) => list.iter().any(expr_reaches_unsafe_builtin),
+        Expression::Coalesce(list) => list
+            .iter()
+            .any(|e| expr_reaches_unsafe_builtin(e, registry)),
         Expression::FunctionCall(f, args) => {
-            function_is_unsafe(f) || args.iter().any(expr_reaches_unsafe_builtin)
+            function_is_unsafe(f, registry)
+                || args
+                    .iter()
+                    .any(|e| expr_reaches_unsafe_builtin(e, registry))
         }
-        Expression::Exists(pattern) => pattern_reaches_unsafe_builtin(pattern),
+        Expression::Exists(pattern) => pattern_reaches_unsafe_builtin(pattern, registry),
     }
 }
 
 /// Whether `f` is itself one of the stateful-mint builtins (see
-/// [`is_parallel_safe`]'s doc comment for the full rationale).
-fn function_is_unsafe(f: &Function) -> bool {
+/// [`is_parallel_safe`]'s doc comment for the full rationale), or an unsafe
+/// [`Function::Custom`] user-function call.
+///
+/// A `Custom(iri)` call is resolved against `registry` (`ctx.user_functions`):
+///
+/// - No registry, or the IRI resolves to neither custom kind (an XSD cast, or an
+///   undefined-function hard error) — deterministic, so safe.
+/// - Resolves to a **native** function — safe iff its declared
+///   [`Volatility`] is NOT [`Volatile`](Volatility::Volatile): `Stable` is
+///   deterministic for the lifetime of the query, so it may run across
+///   fork-join workers exactly like a pure builtin.
+/// - Resolves to a **SPARQL-bodied** [`crate::user_fn::UserFunction`] —
+///   ALWAYS unsafe, conservatively: its body is itself an arbitrary SPARQL
+///   query that can mint `RAND`/`UUID`/`BNODE`/list cells, and that per-call
+///   state would merge into a forked child (see
+///   `crate::user_fn::eval_user_function`'s state merge-back) rather than the
+///   real `ctx` — silently diverging from the sequential stream exactly like
+///   the builtins above. A sequential fallback is always correct.
+fn function_is_unsafe(f: &Function, registry: Option<&UserFunctionRegistry>) -> bool {
     match f {
         Function::Rand | Function::Uuid | Function::StrUuid | Function::BNode => true,
         Function::Purrdf(call) => matches!(
@@ -257,58 +298,84 @@ fn function_is_unsafe(f: &Function) -> bool {
             purrdf_sparql_algebra::PurrdfFn::ListSlice
                 | purrdf_sparql_algebra::PurrdfFn::ListConcat
         ),
+        Function::Custom(iri) => {
+            let Some(reg) = registry else { return false };
+            if let Some(native) = reg.resolve_native(iri.as_str()) {
+                // Native fn: unsafe iff declared Volatile; Stable is
+                // deterministic-within-query, hence fork-join safe. (Wildcard
+                // arm — Volatility is `#[non_exhaustive]`.)
+                return matches!(native.volatility, Volatility::Volatile);
+            }
+            // A SPARQL-bodied user function's body may mint RAND/UUID/BNODE/list
+            // cells whose per-query state merges into a *forked* child and is
+            // then discarded — silently diverging from the sequential stream.
+            // Classify it UNSAFE (conservative + correct; sequential is always
+            // right). An IRI resolving to neither custom kind is an XSD cast or
+            // a hard error — both deterministic — so it stays safe.
+            reg.resolve(iri.as_str()).is_some()
+        }
         _ => false,
     }
 }
 
 /// `true` iff `pattern` (recursively) reaches an unsafe builtin through any
 /// expression-bearing variant — see [`is_parallel_safe`].
-fn pattern_reaches_unsafe_builtin(pattern: &GraphPattern) -> bool {
+fn pattern_reaches_unsafe_builtin(
+    pattern: &GraphPattern,
+    registry: Option<&UserFunctionRegistry>,
+) -> bool {
     match pattern {
         GraphPattern::Bgp { .. } | GraphPattern::Path { .. } | GraphPattern::Values { .. } => false,
         GraphPattern::Join { left, right }
         | GraphPattern::Lateral { left, right }
         | GraphPattern::Union { left, right }
         | GraphPattern::Minus { left, right } => {
-            pattern_reaches_unsafe_builtin(left) || pattern_reaches_unsafe_builtin(right)
+            pattern_reaches_unsafe_builtin(left, registry)
+                || pattern_reaches_unsafe_builtin(right, registry)
         }
         GraphPattern::Graph { inner, .. }
         | GraphPattern::Service { inner, .. }
         | GraphPattern::Distinct { inner }
         | GraphPattern::Reduced { inner }
         | GraphPattern::Slice { inner, .. }
-        | GraphPattern::Project { inner, .. } => pattern_reaches_unsafe_builtin(inner),
+        | GraphPattern::Project { inner, .. } => pattern_reaches_unsafe_builtin(inner, registry),
         GraphPattern::Filter { expr, inner } => {
-            expr_reaches_unsafe_builtin(expr) || pattern_reaches_unsafe_builtin(inner)
+            expr_reaches_unsafe_builtin(expr, registry)
+                || pattern_reaches_unsafe_builtin(inner, registry)
         }
         GraphPattern::Extend {
             inner, expression, ..
-        } => expr_reaches_unsafe_builtin(expression) || pattern_reaches_unsafe_builtin(inner),
+        } => {
+            expr_reaches_unsafe_builtin(expression, registry)
+                || pattern_reaches_unsafe_builtin(inner, registry)
+        }
         GraphPattern::LeftJoin {
             left,
             right,
             expression,
         } => {
-            pattern_reaches_unsafe_builtin(left)
-                || pattern_reaches_unsafe_builtin(right)
-                || expression.as_ref().is_some_and(expr_reaches_unsafe_builtin)
+            pattern_reaches_unsafe_builtin(left, registry)
+                || pattern_reaches_unsafe_builtin(right, registry)
+                || expression
+                    .as_ref()
+                    .is_some_and(|e| expr_reaches_unsafe_builtin(e, registry))
         }
         GraphPattern::OrderBy { inner, expression } => {
-            pattern_reaches_unsafe_builtin(inner)
+            pattern_reaches_unsafe_builtin(inner, registry)
                 || expression.iter().any(|oe| match oe {
                     OrderExpression::Asc(e) | OrderExpression::Desc(e) => {
-                        expr_reaches_unsafe_builtin(e)
+                        expr_reaches_unsafe_builtin(e, registry)
                     }
                 })
         }
         GraphPattern::Group {
             inner, aggregates, ..
         } => {
-            pattern_reaches_unsafe_builtin(inner)
+            pattern_reaches_unsafe_builtin(inner, registry)
                 || aggregates.iter().any(|(_, agg)| match agg {
                     AggregateExpression::CountStar { .. } => false,
                     AggregateExpression::FunctionCall { expression, .. } => {
-                        expr_reaches_unsafe_builtin(expression)
+                        expr_reaches_unsafe_builtin(expression, registry)
                     }
                 })
         }
@@ -633,7 +700,7 @@ mod tests {
             Box::new(Expression::Literal(Literal::new_simple("1"))),
             Box::new(Expression::Literal(Literal::new_simple("2"))),
         );
-        assert!(is_parallel_safe(&arith));
+        assert!(is_parallel_safe(&arith, None));
 
         let regex = call(
             Function::Regex,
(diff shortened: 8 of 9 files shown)


paudley:
# Implementation Plan for Issue #91 — sparql-eval: native (Rust-closure) user-function registry

## Context

Today `UserFunctionRegistry` (`crates/sparql-eval/src/user_fn.rs`) holds only
**SHACL-AF SPARQL-bodied** functions: an IRI whose body is parsed SPARQL
(`sh:select`/`sh:ask`), stored as pure data and executed by re-entering the
evaluator in a recursion-bounded child context. The consumer's deductive-recall
layer needs **scoring predicates** callable inside SPARQL (BM25,
vector-similarity, phonetic name-similarity) whose bodies are native Rust backed
by external index structures and cannot be expressed as SPARQL. The ask is a
second registration seam for native Rust closures dispatched from the *same*
`Function::Custom` eval site, leaving the existing SPARQL-bodied path untouched,
with a determinism/purity declaration wired into the parallel-evaluation gate so
scores can participate in `FILTER`/`ORDER BY` and get engine push-down instead of
the consumer's interim pre-binding-into-a-scratch-world fallback.

## Issue Summary

Add `UserFunctionRegistry::register_native(iri, purity, closure)` where the
closure is `Arc<dyn Fn(&[TermValue]) -> Result<TermValue, EvalError> + Send +
Sync>`; dispatch it from `Function::Custom` (`crates/sparql-eval/src/expr.rs:1641`)
exactly like the SPARQL-bodied path; declare purity at registration so the
fork-join determinism gate (`crates/sparql-eval/src/parallel.rs`) can exclude
non-pure native functions from parallel evaluation.

- Gap analysis is adversarial, not sycophantic.
- **No deferrals / follow-ups / future work — everything is done NOW.**
- No GitHub issue/PR numbers or process-flow references in repo source, docs, or
  the ontology. (Commit messages / PR body only.)
- **"When you see a bug, you own it"** — this plan fixes the pre-existing latent
  nondeterminism in the `Function::Custom` parallel-safety classification
  (SPARQL-bodied user functions currently run under fork-join and can diverge),
  not just the native side.

- GREENFIELD-FIRST / SOTA / no compromises.
- RUST-FIRST, PYTHON-SURFACE — core is Rust.
- SUBSUME/EXTEND/ENHANCE.
- LOW/NO OPTIONALITY, HARD FAILS — no Cargo features, no optional components, no
  swallowed errors; fail fast.
- MAXIMAL UTILITY / MAXIMAL PERFORMANCE / MAXIMAL PORTABILITY (wasm-clean core).

Project hard rules honored: no new Cargo features; no dependency pins in the
member crate; determinism preserved in the evaluator; wasm32 stays clean
(`make wasm`); no oxigraph/PyO3 near core; `example.org` fixtures only.

## Completeness Contract

Every requirement and acceptance criterion maps to a numbered task. NOTHING is
deferred, stubbed, or left for a future PR.

| R1 Native-closure registration API `register_native(iri, arity, volatility, Arc<dyn Fn(&[TermValue]) -> Result<TermValue,EvalError> + Send + Sync>)` | 1 |
| R2 Unified dispatch from `Function::Custom` eval site (same site as SPARQL-bodied) | 2 |
| R3 Existing SPARQL-bodied registry + semantics + all `user_fn.rs` tests unchanged | 1, 2 (non-regression) |
| R4 Determinism/purity contract documented | 1 (rustdoc) |
| R5 Purity declared at registration (machine-checkable, not just docs) | 1 (`Volatility` enum arg) |
| R6 Parallel-eval determinism gate accounts for native purity (`Volatile` ⇒ parallel-unsafe) | 3 |
| R6b (owned bug) SPARQL-bodied `Function::Custom` correctly classified parallel-unsafe | 3 |
| R7 Push-down: native fn usable inside `FILTER`/`ORDER BY` with correct deterministic results | 4 (tests) |
| R8 `Send + Sync` closure bound; `EvalCtx` `assert_send_sync` proof still compiles | 1, 3 |
| E-arity Fail-fast declared arity check (no silent short/long slice into host code) | 1, 4 |
| E-panic Native closure panics caught + converted to a deterministic `EvalError` (no nondeterministic rayon abort) | 1, 4 |
| E-collision Registering one IRI as both native and SPARQL-bodied hard-fails (no silent shadowing) | 1 |
| E-display `EvalError::Function` Display generalized (no longer mislabels a native error "SHACL-AF") | 1 |
| E-nan `ORDER BY` over `xsd:double` scores (incl. `NaN`) is a stable total order | 4 |
| AC wasm32 build stays clean (`make wasm`) | 5 |
| AC public re-exports of new types (`NativeFunction`, `NativeFnBody`, `Arity`, `Volatility`) | 1 |

**Adopted from the Enhancement Audit** (folded into the tasks below): a
three-value `Volatility` descriptor instead of a purity bool (honest for
frozen-index reads = `Stable`, not `Immutable`); fail-fast declared **arity**
checking; **panic-catching** around the native closure; a **collision guard** so
one IRI cannot be both native and SPARQL-bodied; the `EvalError::Function`
**Display** generalization; and a **NaN-ordering** total-order test.

Declined, with reasons (closed design decisions — NOT deferred work):

- **Structural merge of the two registries into one `UserFunction` with a `Body`
  enum (the audit's TRANSFORMATION item).** Declined. Its *only* correctness
  argument is that two tables + "resolve SPARQL first" could silently shadow a
  native fn registered under the same IRI — and that is fully neutralized by the
  adopted fail-fast **collision guard** (registering an IRI already held by the
  other kind is a hard error), which makes the ambiguity unrepresentable without
  touching `UserFunction`. The merge's remaining benefits are ergonomic (a
  one-arm gate, a shared `purity` field), and its cost is not local: `UserFunction`
  is a cross-crate public type constructed by the shapes crate's SHACL-AF
  declaration parser, so a `Body`-enum reshape ripples into that crate and
  rewrites every existing SPARQL-bodied test — churning a stable contract that
  R3 requires to stay behaviorally unchanged, for no correctness gain. The
  *substance* of the transformation — a typed, volatility-annotated,
  planner-legible function descriptor (the pg_proc analogy) — is captured by the
  adopted `Arity` + `Volatility` fields on `NativeFunction`; the lesser (two
  parallel tables sharing that descriptor vocabulary) therefore suffices and is
  strictly less invasive.
- **Recursive body-safety walk for SPARQL-bodied functions** (prove a body
  parallel-safe rather than classifying all SPARQL-bodied fns unsafe). Declined:
  a sequential fallback is always correct (the parallel module's own doctrine —
  "when in doubt, flag UNSAFE"); an exact body walk needs cross-function cycle
  analysis over the registry for only a conditional speedup on a rare construct
  (SPARQL-bodied UDFs in hot FILTER/ORDER-BY over >1024 rows). Conservative-unsafe
  is correct, simple, deterministic.
- **Shared `Signature`/type-constraint (`sh:datatype`/`sh:nodeKind`) checking for
  native fns.** Declined beyond arity: a native scorer validates its own argument
  *values* (it holds the `TermValue`s and its own domain rules), so forcing
  SHACL-AF type-constraint metadata onto a Rust closure is awkward and redundant.
  Declared *arity* (adopted) is the fail-fast win worth keeping; value-typing
  stays the closure's own hard-fail responsibility.
- **Const-folding native calls; monotonicity/cost annotation for `ORDER BY … LIMIT
  k` top-k pushdown.** Declined *now*: both are optimizer changes (a planner pass,
  a bounded-heap top-k operator) that would be stubs if half-built. The adopted
  `Volatility` enum is `#[non_exhaustive]`, so a finer determinism class (e.g. an
  `Immutable` variant) or a sibling cost/monotonicity descriptor is addable later
  without a breaking change — but no dead variant is carried and nothing is left
  unimplemented for *this* issue's scope.
- **Per-call `Vec<TermValue>` argument-buffer reuse.** Declined: presenting a
  contiguous `&[TermValue]` to the closure inherently requires collecting the
  bound values (an `&[Option<TermValue>]` cannot be reinterpreted as `&[TermValue]`);
  a per-worker reusable scratch buffer adds state-threading for an unmeasured gain,
  and this project's discipline is that benches are report-only (never assert a
  speedup). The allocation is inherent, not waste.
- **`register_pure_native` convenience wrapper.** REJECTED — a second entry point
  that only defaults the volatility arg is exactly the optional/duplicative surface
  `.goals` LOW/NO-OPTIONALITY forbids. Single explicit `register_native` only.
- **Separate native-only engine entry point.** Declined: native functions live in
  the *same* `UserFunctionRegistry`, so `query_with_user_functions` and the SHACL
  prebinding variants already carry them — no new engine surface (LEVERAGE).

```bash
BRANCH="paudley/91-native-user-function-registry"
WORKTREE="$REPO_ROOT/.worktrees/91-native-user-function-registry"
cd "$WORKTREE"
git branch --show-current   # MUST show paudley/91-native-user-function-registry, NOT main
```

Build environment sanity in the fresh worktree (do NOT gate on a full test run —
that is stage2/stage3's job):

```bash
cargo build -p purrdf-sparql-eval
```

Post this plan to the issue:

```bash
gh issue comment 91 --body-file /home/paudley/.claude/plans/91-structured-peach.md
```

**⚠️ DO NOT proceed to Task 1 until the worktree exists and you are on the feature branch.**

## Task 1: Native function type, registry table, and `register_native` API

`crates/sparql-eval/src/user_fn.rs` (+ `error.rs`, `lib.rs`)

1. Add the native-function types (module-level rustdoc must state the
   determinism/volatility contract — R4):
   ```rust
   /// A native (host-Rust) user function body: a closure over the already-
   /// evaluated, dataset-independent argument values. Unlike a SPARQL body it takes
   /// NO `EvalCtx` — it therefore cannot re-enter the evaluator, so there is no
   /// recursion/re-entrancy boundary to bound (the SPARQL-bodied path's
   /// `MAX_UDF_DEPTH` guard does not apply). When declared non-`Volatile` it MUST be
   /// deterministic within a query: the fork-join parallel gate relies on that
   /// declaration to decide whether the function may run across workers.
   pub type NativeFnBody =
       Arc<dyn Fn(&[TermValue]) -> Result<TermValue, EvalError> + Send + Sync>;

   /// A native function's determinism class (the volatility axis of its descriptor,
   /// after Postgres `provolatile`). The fork-join gate is its sole consumer:
   /// `Volatile` is pinned to sequential evaluation; `Stable` (deterministic *within
   /// one query* — a frozen external-index read, or pure math) may run across
   /// workers. Marked `#[non_exhaustive]` so a finer class (e.g. `Immutable`, for a
   /// future const-folding pass) is addable without a breaking change — no dead
   /// variant is carried now.
   #[derive(Clone, Copy, PartialEq, Eq, Debug)]
   #[non_exhaustive]
   pub enum Volatility { Stable, Volatile }

   /// A native function's declared argument arity, checked before the closure is
   /// ever invoked (fail-fast: a wrong-count call never hands the host a short or
   /// long slice). `Variadic` accepts any count.
   #[derive(Clone, Copy, PartialEq, Eq, Debug)]
   pub enum Arity { Exact(usize), Range { min: usize, max: usize }, AtLeast(usize) }

   /// A registered native function: its closure body plus its declared arity and
   /// volatility.
   #[derive(Clone)]
   pub struct NativeFunction {
       pub(crate) body: NativeFnBody,
       pub(crate) arity: Arity,
       pub(crate) volatility: Volatility,
   }
   ```
2. Add a second table to the registry and drop the `Debug` derive (a boxed
   closure has no `Debug`), replacing it with a hand-written `Debug` that lists
   the two key-sets and omits the closures (mirrors `EvalCtx`'s manual `Debug`):
   ```rust
   #[derive(Default, Clone)]   // Debug removed from derive; hand-written below
   pub struct UserFunctionRegistry {
       fns: DetHashMap<String, UserFunction>,
       native: DetHashMap<String, NativeFunction>,
   }
   ```
3. API additions:
   - `pub fn register_native(&mut self, iri: impl Into<String>, arity: Arity, volatility: Volatility, body: NativeFnBody)`
     — **collision guard (E-collision):** panic with a clear message if the IRI is
     already registered as a SPARQL-bodied function; symmetrically, `insert` panics
     if the IRI is already a native function. Same-kind re-registration keeps the
     existing "last write wins" semantics; only *cross-kind* collision is the hard
     fail (a host misconfiguration caught at build time, deterministic, zero
     hot-path cost). The shapes crate only ever calls `insert`, so it never trips
     this.
   - `#[must_use] pub fn resolve_native(&self, iri: &str) -> Option<&NativeFunction>`
   - update `is_empty` → both tables empty; `len` → sum of both.
4. `eval_native_function(native, iri, args: &[Option<TermValue>], _ctx)`:
   - **Arity check first (E-arity):** validate `args.len()` against `native.arity`;
     on mismatch return `EvalError::function("native function <iri> expects … argument(s), got N")`
     (mirrors the SPARQL-bodied `wrong_arity_is_a_hard_error` text shape).
   - If **any** argument is unbound (`None`) → `Ok(None)`: a native function
     declares no per-parameter optionality, so a missing value yields no result
     node (SHACL-AF-consistent, and the natural SPARQL expression-error→unbound).
     Document this.
   - Else collect `Vec<TermValue>` of the bound values and invoke the closure
     **inside `std::panic::catch_unwind(AssertUnwindSafe(…))` (E-panic)**: a host
     closure that panics (dim mismatch, `unwrap`, OOB index) is converted to a
     deterministic `EvalError::function("native function <iri> panicked")` — a
     fixed, payload-free message so the result is identical regardless of which
     rayon worker panicked. Mirrors the existing `native-codec-panic` guard in
     `purrdf_rdf::native_codecs::parse` (established pattern; wasm-clean).
   - A returned `Err` maps straight through (hard fail, never swallowed).
   - Returns `Result<Option<TermValue>>`; the caller interns the value.
5. **E-display:** in `error.rs`, generalize `EvalError::Function`'s `Display` from
   `"SHACL-AF function error: {msg}"` to a kind-neutral `"user function error: {msg}"`
   so a native scorer's error is not mislabeled SHACL-AF. Update the variant's
   rustdoc to note it now covers native functions too. Fix any test asserting the
   old string.
6. `crates/sparql-eval/src/lib.rs`: extend the `user_fn` re-export with
   `NativeFunction, NativeFnBody, Arity, Volatility`.
7. Tests in `user_fn.rs` (`example.org` IRIs):
   - `native_function_returns_value` — `register_native` a `Stable` closure, call
     it in a SELECT projection, assert the returned literal.
   - `native_function_error_is_a_hard_error` — closure returns `Err`, query fails.
   - `native_function_panic_is_a_clean_error` — closure `panic!`s; assert a
     deterministic `EvalError::function` ("panicked"), not an abort.
   - `native_function_wrong_arity_is_a_hard_error` — call count ≠ declared arity.
   - `native_function_unbound_argument_yields_no_value`.
   - `register_native_collision_with_sparql_bodied_panics` (and the reverse).
   - `native_and_sparql_bodied_coexist` — both kinds registered under distinct
     IRIs; each resolves to its own path; SPARQL-bodied behavior unchanged.

### After Task 1: github-commit-agent → github-push-agent → post summary to issue #91

## Task 2: Dispatch native functions from the `Function::Custom` eval site

`crates/sparql-eval/src/expr.rs` (the `Function::Custom(iri)` arm, ~1641)

Insert a native-resolution branch **after** the existing SPARQL-bodied resolve
and **before** the XSD-cast fallback, so a function IRI never collides with a
datatype IRI and both custom kinds share one site (R2):

```rust
if let Some(registry) = ctx.user_functions {
    if let Some(func) = registry.resolve(iri.as_str()) {
        let result = crate::user_fn::eval_user_function(func, iri.as_str(), &vals, ctx)?;
        return Ok(result.map(|value| intern(ctx, value)));
    }
    if let Some(native) = registry.resolve_native(iri.as_str()) {
        let result = crate::user_fn::eval_native_function(native, iri.as_str(), &vals, ctx)?;
        return Ok(result.map(|value| intern(ctx, value)));
    }
}
```

`vals` is the already-evaluated `Vec<Option<TermValue>>`; interning the result
reuses the same `intern(ctx, value)` path as the SPARQL-bodied return, so id
assignment stays deterministic.

Test (in `expr.rs` or `user_fn.rs`): a native function called from a `FILTER`
expression over a small `example.org` dataset returns correct keep/drop rows.

### After Task 2: github-commit-agent → github-push-agent → post summary to issue #91

## Task 3: Wire purity into the fork-join parallel-safety gate (R6 + owned bug R6b)

`crates/sparql-eval/src/parallel.rs` and its call sites.

1. Thread the registry into the safety walk:
   - `pub(crate) fn is_parallel_safe(expr: &Expression, registry: Option<&UserFunctionRegistry>) -> bool`
   - `pub(crate) fn is_parallel_safe_pattern(pattern: &GraphPattern, registry: Option<&UserFunctionRegistry>) -> bool`
   - pass `registry` down through `expr_reaches_unsafe_builtin`,
     `pattern_reaches_unsafe_builtin`, `function_is_unsafe`.
2. `function_is_unsafe(f, registry)` — replace the `Function::Custom` default:
   ```rust
   Function::Custom(iri) => {
       let reg = match registry { Some(r) => r, None => return false };
       if let Some(native) = reg.resolve_native(iri.as_str()) {
           // Native function: unsafe iff declared Volatile; Stable is
           // deterministic-within-query, hence fork-join safe. (Match with a
           // wildcard arm — the enum is `#[non_exhaustive]`.)
           return matches!(native.volatility, Volatility::Volatile);
       }
       // A SPARQL-bodied user function's body may mint RAND/UUID/BNODE/list cells
       // whose per-query state merges into a *forked* child and is then discarded —
       // silently diverging from the sequential stream. Classify it UNSAFE
       // (conservative + correct; sequential is always right — this is the owned
       // R6b fix). An IRI resolving to neither custom kind is an XSD cast or a hard
       // error — both deterministic — so it stays safe.
       reg.resolve(iri.as_str()).is_some()
   }
   ```
3. Update every call site to pass `ctx.user_functions`:
   `crates/sparql-eval/src/expr.rs` (`eval_filter`, `eval_extend`),
   `crates/sparql-eval/src/binop.rs` (`eval_union` ×2 branch checks, left-join),
   `crates/sparql-eval/src/modifier.rs` (per-aggregate check). All have `ctx` in
   scope.
4. Tests in `parallel.rs`:
   - `native_stable_custom_is_parallel_safe` — registry with a `Volatility::Stable`
     native fn ⇒ `is_parallel_safe(Custom(iri), Some(&reg)) == true`.
   - `native_volatile_custom_is_parallel_unsafe` — `Volatility::Volatile` ⇒ `false`.
   - `sparql_bodied_custom_is_parallel_unsafe` — a registered SPARQL-bodied fn ⇒
     `false` (documents the owned-bug fix).
   - `unknown_custom_without_registry_stays_safe` — `None` registry / unresolved
     IRI ⇒ `true` (XSD-cast/hard-error path unaffected).
   - update existing `is_parallel_safe(&expr)` test calls to the new 2-arg form
     (pass `None`).

### After Task 3: github-commit-agent → github-push-agent → post summary to issue #91

## Task 4: End-to-end push-down determinism tests (R7)

`crates/sparql-eval/src/user_fn.rs` tests (engine-level, `example.org`).

Register a **`Stable`** native scorer (a deterministic `score(?x)` returning an
`xsd:double`) and exercise the motivating push-down surfaces through
`NativeSparqlEngine::query_with_user_functions`:

- `native_score_in_filter_pushes_down` — `FILTER(<…#score>(?x) > 0.5)` returns
  exactly the rows whose score exceeds the threshold.
- `native_score_in_order_by_is_deterministic` — `ORDER BY DESC(<…#score>(?x))`
  yields a stable, correct ordering; run twice, assert identical output
  (determinism proof — a `Stable` native fn is reproducible under the parallel gate).
- `native_score_order_by_is_total_over_nan` (E-nan) — a scorer that returns
  `xsd:double` `NaN` for some rows; assert `ORDER BY` produces a **stable total
  order** identical across two runs. This either documents that the evaluator's
  `xsd:double` ordering already pins `NaN` deterministically (SPARQL's ordering
  gives it a fixed slot) or, if a bare `partial_cmp` collapses `NaN`, surfaces
  that latent ordering bug for a fix here (`.baseline`: own the bug).
- A dataset large enough to cross `PARALLEL_MIN_ROWS` (>1024 rows) for the FILTER
  case, so the `Stable` native path actually exercises fork-join and its result is
  asserted equal to the expected sequential answer (push-down safety in practice).

### After Task 4: github-commit-agent → github-push-agent → post summary to issue #91

## Task 5: Gate verification — full quality gate + wasm

Run in the worktree:

```bash
cargo fmt --all            # fmt gate runs first; run before make check
make check                 # fmt + clippy (pedantic/nursery, warning-free) + build + tests + hygiene
make wasm                  # every release crate stays wasm32-unknown-unknown clean
```

Resolve any clippy/hygiene finding here (no `#[allow]` escape hatches for the new
code). This task is verification only — no new production behavior; it exists so
"done" means the gate is green, not asserted.

### After Task 5: github-commit-agent (if fixups) → github-push-agent → post summary to issue #91

1. Verify all changes committed and pushed; run the Mechanical Deferral Scan over
   the branch diff (no `TODO`/`FIXME`/`unimplemented!`/"future work" left).
2. Create PR:
   ```bash
   gh pr create --title "feat(sparql-eval): native Rust-closure user-function registry" --body "Closes #91

   ## Summary
   Adds a native (Rust-closure) registration seam to UserFunctionRegistry,
   dispatched from the same Function::Custom eval site as SPARQL-bodied
   functions, with a purity declaration wired into the fork-join parallel gate.

   ## Changes
   - user_fn.rs: NativeFunction/NativeFnBody/Arity/Volatility, register_native/resolve_native (with cross-kind collision guard), eval_native_function (arity check + panic-catch)
   - error.rs: EvalError::Function Display generalized to 'user function error'
   - expr.rs: native dispatch branch in the Function::Custom arm
   - parallel.rs: registry-aware is_parallel_safe; Volatile native + SPARQL-bodied custom now parallel-unsafe (owned determinism-bug fix)
   - lib.rs: re-export new public types
   - tests: FILTER/ORDER BY push-down + determinism + NaN total-order + arity/panic/collision + parallel-gate classification
   "
   ```
3. Post plan to PR:
   ```bash
   PR_NUM=$(gh pr list --head paudley/91-native-user-function-registry --json number -q '.[0].number')
   gh pr comment $PR_NUM --body-file /home/paudley/.claude/plans/91-structured-peach.md
   ```
4. Post to PR and issue the two reflection answers:
5. Report: "Stage 1 complete. PR #<num> created for issue #91. Ready for /stage2 91".

## Execution Rules

1. Task 0 (worktree) MANDATORY first; never work on `main`.
4. Task N (PR) MANDATORY last.
5. Use sub-agents for all commits/pushes; wiring-verifier after implementation tasks.

## Verification Section (how "done" is proven — executable checks)

- **R1/R5/R8:** `register_native(iri, Arity::Exact(1), Volatility::Stable, closure)`
  compiles; the `assert_send_sync::<EvalCtx>` const-fn proof still builds (Task 5
  `make check`).
- **R2:** a native fn called via `<…#score>(?x)` in a query resolves and returns
  a value (Task 1/2/4 tests through `query_with_user_functions`).
- **R3:** all pre-existing `user_fn.rs` tests pass unchanged (Task 5).
- **R6:** `is_parallel_safe(Custom(iri), Some(&reg))` is `true` for a `Stable`
  native fn and `false` for a `Volatile` one (Task 3 unit tests).
- **R6b:** `is_parallel_safe(Custom(iri), Some(&reg))` is `false` for a
  SPARQL-bodied fn (Task 3 unit test).
- **R7:** FILTER/ORDER BY over a >1024-row `example.org` dataset returns the exact
  expected rows/order, identical across two runs (Task 4 tests).
- **E-arity/E-panic/E-collision:** wrong-arity call, panicking closure, and
  cross-kind IRI collision each hard-fail with a deterministic error (Task 1 tests).
- **E-display:** `EvalError::Function` renders "user function error: …" (Task 1).
- **E-nan:** `ORDER BY` over `NaN` scores is a stable total order across two runs
  (Task 4 test).
- **AC wasm:** `make wasm` clean (Task 5).

paudley:
### Stage 1 reflection

**What I'm least confident about, and why:**
The `catch_unwind` panic guard depends on the crate being built with `panic = "unwind"` (the default for native `cargo test`/`build`). On a `panic = "abort"` profile the guard cannot catch — the process aborts instead. That is acceptable and precedented (it mirrors `purrdf_rdf::native_codecs::parse`'s `native-codec-panic` guard, and `make wasm` stays green because wasm's `panic=abort` compiles the guard as a no-op), but the *determinism-under-panic* guarantee the tests assert only holds under unwind. If a downstream release profile sets `panic=abort`, a panicking native closure aborts the query process rather than returning a clean `EvalError`. The closure contract (don't panic; return `Err`) is the real backstop; the guard is defense-in-depth.

**What you don't know but should:**
1. **SPARQL-bodied UDFs are now pinned to sequential evaluation.** Fixing the latent fork-join nondeterminism means every SPARQL-bodied `sh:SPARQLFunction` used in a `FILTER`/`ORDER BY`/aggregate over >1024 rows now runs sequentially, not in parallel. This is correct (the prior parallel path could diverge), but it is a *performance* change for existing SHACL-AF workloads, not just an additive feature. A future body-safety walk could restore parallelism for provably-pure bodies (declined here as a closed design decision, not deferred work).
2. **Native purity is an honor-system contract.** The `Volatility::Stable` declaration is trusted, not verified — a closure that lies (reads a clock/atomic) and is declared `Stable` will run across fork-join workers and can diverge. The type system cannot enforce determinism; the contract is documented on `NativeFnBody`.
3. **`register_native` has no in-repo consumer yet** — it is a host-facing extensibility entry point for the downstream consumer crate. The *consuming* half (dispatch, parallel gate) is fully production-wired and exercised by engine-level tests; the *registering* half will be driven by the consumer. This is intentional (the seam is the deliverable), confirmed by the wiring-verifier.

paudley:
# Stage 2 Gap Analysis & Remediation — Issue #91 / PR #99

**Feature:** native (Rust-closure) user-function registry in `purrdf-sparql-eval`
**Worktree:** `.worktrees/91-native-user-function-registry` · **Branch:** `paudley/91-native-user-function-registry` · **Base:** `origin/main`

## Context

Issue #91 adds a second registration seam to `UserFunctionRegistry` for **native Rust closures**
(`register_native(iri, arity, volatility, closure)`) dispatched from the same `Function::Custom`
eval site as the existing SPARQL-bodied path, with a purity/volatility declaration wired into the
fork-join parallel-determinism gate so native scores (BM25/vector/phonetic) can participate in
`FILTER`/`ORDER BY` and get engine push-down. The consumer's deductive-recall layer is the downstream
consumer; the interim fallback pre-binds scored candidates into a scratch world, blocking push-down.

Two independent audits ran (gap-analyzer + completion-adversary). The Completeness Contract
(R1–R8, R6b, E-arity/panic/collision/display/nan, AC-wasm/reexports) is **substantively and
honestly implemented and test-covered** — the deferral hunt is clean and the "Enhancements
Considered and Declined" section clears the innocence test. But the audits **disagreed on
readiness**, and the disagreement exposes a real blocker:

- The **completion-adversary returned PASS** — but only ran `cargo test`/`clippy`/`wasm`; it never
  ran `make doc`, so its PASS is **incomplete and does not hold**.
- The **gap-analyzer ran `make doc` and found the `doc` CI gate is RED.** `make check` does not run
  the doc gate, so "check green" ≠ "CI green." This is the exact failure class main already fixed
  once (`304d045`).

Three gaps result. No deferrals — all fixed in this PR.

## Gaps to remediate (fix → commit → push → verify, one at a time)

### Gap 1 — CRITICAL: `doc` CI gate is RED (private intra-doc link)
- **Where:** `crates/sparql-eval/src/user_fn.rs:222` — `Debug for NativeFunction` rustdoc links
  `` [`eval_native_function`] `` to a `pub(crate)` item.
- **Why it's a gap:** the repo's `make doc` runs under `RUSTDOCFLAGS="-D warnings"`; a public-doc
  link to a private item is a hard error (`gh pr checks 99` → `doc fail`, reproduced locally). "Done"
  is currently false while CI is red.
- **Fix:** drop the intra-doc link brackets → plain code span `` `eval_native_function` `` (or point
  it at a public item). Trivial.
- **Verify:** `make doc` from the worktree returns clean; `gh pr checks 99` doc job goes green.

### Gap 2 — HIGH: per-call deep clone of every argument on the scoring hot path  *(user confirmed: fix now)*
- **Where:** `NativeFnBody` type `crates/sparql-eval/src/user_fn.rs:141`; the clone at
  `user_fn.rs:499-502` (`args.iter().map(|arg| arg.clone()...).collect::<Vec<TermValue>>()`).
- **Why it's a gap:** `TermValue` owns heap strings (IRIs/literals); `Fn(&[TermValue])` forces a
  per-call deep clone on precisely the per-row scoring path (many candidates) the issue exists to
  serve. `.goals` mandate MAXIMAL PERFORMANCE / GREENFIELD-FIRST / no-compromises, and `NativeFnBody`
  is a **public host-facing type** — a signature change is breaking post-release, so this is
  now-or-never. The plan's declined "per-call `Vec` buffer reuse" addressed a *different*
  optimization; the deep-clone-vs-reference choice was never adjudicated. Reviewer flagged it across
  9 inline comments.
- **Fix:** change
  `NativeFnBody = Arc<dyn Fn(&[&TermValue]) -> Result<TermValue, EvalError> + Send + Sync>`;
  in `eval_native_function` collect `Vec<&TermValue>` (references into `args`, valid for the call
  duration — no heap-string clone) instead of `Vec<TermValue>`. Update the `NativeFnBody` rustdoc to
  describe the borrowed-slice contract. Thread the `&[&TermValue]` signature through **every** native
  closure in the `user_fn.rs` test module and any `parallel.rs` test helper.
- **Verify:** `cargo test -p purrdf-sparql-eval` all green; grep-confirm no `arg.clone()`/`Vec<TermValue>`
  materialization remains in the native dispatch; `make check` + `make wasm` clean.

### Gap 3 — LOW: module rustdoc still headlines only "SHACL-AF SPARQL-based functions"
- **Where:** `crates/sparql-eval/src/user_fn.rs:4` module doc.
- **Fix:** broaden the module header + lead paragraph to cover **both** kinds the module now owns —
  the SPARQL-bodied `UserFunction` and the native `NativeFunction` seam.
- **Verify:** `make doc` clean; wording reads accurately for both kinds.

## Reviewer feedback — adjudicated (not deferrals; documented rejections)
- **9× gemini** "`&[&TermValue]` to drop heap clones" → **ADOPTED** as Gap 2.
- **2× gemini security-medium** "`assert!` collision guard is a panic/DoS risk; add a non-panicking
  `Result` API" → **CORRECTLY REJECTED.** `.goals` LOW/NO-OPTIONALITY + HARD-FAILS: a cross-kind IRI
  collision is host build-time misconfiguration; fail-fast is the intended contract and a `Result`
  variant would add the forbidden optional surface. Recorded, not actioned.
- CodeRabbit was rate-limited (no substantive findings). No human reviews outstanding.

## Standing-goals / deferral status
No Cargo features added; no swallowed errors; no `#[allow]` escapes; no `#NNN` issue/PR tokens in
source/docs; wasm-clean; SOLID intact. Deferral hunt clean across diff, commit bodies, PR/issue
comments.

## Execution order (Phase 4 — per-gap, no batching)
1. Gap 1 (doc link) → github-commit-agent → github-push-agent → post summary to issue #91 + PR #99 → wiring-verifier.
2. Gap 2 (`NativeFnBody` signature + dispatch + all test closures) → commit → push → summary → wiring-verifier.
3. Gap 3 (module rustdoc) → commit → push → summary → wiring-verifier.

## Exit gate (Phase 6)
Re-run the completion-adversary (STAGE=2) against HEAD **including `make doc`** and re-run the
gap-analyzer. The stage is done only when both return PASS — every criterion
`DEMONSTRATED-IN-PRODUCTION` **and** all CI gates (`make check` + `make doc` + `make wasm`) green.
Then post the final summary to issue #91 and PR #99 and append any exceptions to `.deficiencies`.

## Verification summary (end-to-end)
- `make doc` — clean (Gap 1 exit criterion).
- `cargo test -p purrdf-sparql-eval` — 335+ lib tests green, native closures driven through the real
  `query_with_user_functions` evaluate path (Gap 2 exit criterion).
- `make check` (fmt + clippy pedantic/nursery + build + tests + hygiene) — clean.
- `make wasm` — `wasm32-unknown-unknown` clean.
- `gh pr checks 99` — all jobs green.

paudley:
## Push Summary

**Branch:** paudley/91-native-user-function-registry
**Remote:** origin/paudley/91-native-user-function-registry

### Commits Pushed
971e198 docs(sparql-eval): fix private intra-doc link denied by the doc gate

### Quality Gates
- `make check` (fmt, clippy, build, tests, hygiene, wasm build): PASSED
- Uncommitted changes: NONE
- Pre-push hooks: none configured for this repo

### Session Summary
Pushed a single follow-up commit fixing a private intra-doc link in the
sparql-eval crate that was failing the doc gate. This closes out the
native (Rust-closure) user-function registry work (issue #91 / PR #99):
registry + dispatch, a fork-join parallel-safety gate made registry-aware,
push-down/determinism tests (FILTER/ORDER BY/NaN/parallel), and the
rustfmt/doc-link cleanups on top.

paudley:
## Push Summary

**Branch:** paudley/91-native-user-function-registry
**Remote:** origin/paudley/91-native-user-function-registry

### Commits Pushed
d5dacee docs(sparql-eval): drop plan-id process-flow refs from native-fn tests

### Quality Gates
- `make check` (fmt + clippy + build + tests + hygiene): PASSED
- `make wasm` (wasm32-unknown-unknown build): PASSED
- `make doc` (rustdoc, `-D warnings`): PASSED
- Uncommitted changes: NONE

paudley:
## /stage2 complete — gap remediation + exit gate PASS (issue #91 / PR #99)

Two independent audits (gap-analyzer + completion-adversary) ran against the branch. The
Completeness Contract (R1–R8, R6b, E-arity/panic/collision/display/nan, AC-wasm/reexports) was
already honestly implemented and test-covered. Remediation closed **5 gaps** — the decisive one
being that the first adversary pass **falsely returned PASS because it never ran the doc gate**,
while `make check` skips `make doc`; the workspace `doc` CI job was in fact RED.

### Gaps addressed

| # | Pri | Gap | Resolution | Commit |
|---|-----|-----|------------|--------|
| 1 | CRITICAL | `doc` CI gate RED — `Debug for NativeFunction` rustdoc linked `[eval_native_function]`, a `pub(crate)` item, hard-erroring under `RUSTDOCFLAGS="-D warnings"` | Demoted to a plain code span | `971e198` |
| 2 | HIGH | `NativeFnBody = Fn(&[TermValue])` deep-cloned every arg per call on the per-row native-scoring hot path the seam exists for (public host-facing type → now-or-never) | Changed to `Fn(&[&TermValue])`; `eval_native_function` now collects `Vec<&TermValue>` borrows — no deep clone. All native closures + `parallel.rs` helper threaded through; rustdoc documents the borrow contract *(user-confirmed: fix now)* | `eb97411` |
| 3 | LOW | `user_fn.rs` module rustdoc headlined only "SHACL-AF SPARQL-based functions" | Broadened to cover both kinds (SPARQL-bodied `UserFunction` + native `NativeFunction`) | `27558a6` |
| 4 | CRITICAL (owned) | Pre-existing private intra-doc links in `crates/gts/src/compact.rs` (`base64url_unpadded`) and `crates/rdf/src/gts_certify.rs` (`try_refold_digest`/`try_effective_digest`) kept the whole workspace `doc` gate RED (masked behind Gap 1 by compile ordering) | Demoted to code spans; also dropped a stray "Task 5" process-flow ref. `.baseline` "you own it, fix it now" | `ac000fe` |
| 5 | LOW (owned) | This PR's own diff added plan-id/process-flow markers (`R7` / `(Task 4)` / `(E-nan)`) to test comments — `.baseline` bans process-flow refs in source | Reworded to describe the behavior under test | `d5dacee` |

### Reviewer feedback

- **9× gemini** "use `&[&TermValue]` to drop heap clones" → **ADOPTED** (Gap 2).
- **2× gemini security-medium** "`assert!` collision guard is a panic/DoS risk; add a non-panicking
  `Result` API" → **CORRECTLY REJECTED** and recorded, not deferred. Under `.goals` LOW/NO-OPTIONALITY
  + HARD-FAILS a cross-kind IRI collision is host build-time misconfiguration; fail-fast is the
  intended contract and a `Result` variant would add the forbidden optional surface.
- CodeRabbit: review completed, no substantive findings.

### Completion-adversary — exit gate: **PASS**

Fresh STAGE-2 adversary (this time running `make doc`) against HEAD — every acceptance criterion
**DEMONSTRATED-IN-PRODUCTION** through the real public evaluate path
(`NativeSparqlEngine::query_with_user_functions`), no open findings:

- C1 `register_native` → `native_function_returns_value` (projected `42`).
- C2 unified `Function::Custom` dispatch → `native_and_sparql_bodied_coexist` (both kinds, one query).
- C6/C7 parallel gate honors `Volatility`; SPARQL-bodied classified unsafe (owned bug R6b) →
  `native_score_filter_over_parallel_threshold` (1500 rows > `PARALLEL_MIN_ROWS`, byte-stable),
  `native_volatile_custom_is_parallel_unsafe`, `sparql_bodied_custom_is_parallel_unsafe`.
- C8 push-down in FILTER/ORDER BY → `native_score_in_filter_pushes_down`, `…in_order_by_is_deterministic`.
- E-arity/panic/collision/display/nan → each driven by its dedicated test through the engine.
- Laundering scan: the issue's "Enhancements Considered and Declined" (structural registry merge) and
  the "future body-safety walk" note both **cleared** the innocence test; no follow-up issue was filed
  carrying #91's own work.

### Verification (all green on final HEAD `d5dacee`)

- `make check` (fmt + clippy pedantic/nursery + build + 2269 tests + hygiene) — PASS
- `cargo test -p purrdf-sparql-eval` — 335 lib tests + integration suites, 0 failed
- `make doc` (`-D warnings`, full workspace) — **exit 0** (the gate the false PASS missed)
- `make wasm` (`wasm32-unknown-unknown`) — PASS
- PR state `MERGEABLE`; CI re-running on the final push.

**Ready for /stage3.**

paudley:
feat(sparql-eval): native Rust-closure user-function registry + dispatch

Adds a second registration seam to `UserFunctionRegistry` for native host-Rust
closures, dispatched from the same `Function::Custom` eval site as the existing
SHACL-AF SPARQL-bodied path, with a declared determinism class wired into the
fork-join parallel-evaluation gate. This lets scoring predicates (BM25,
vector-similarity, phonetic name-similarity) that cannot be expressed as SPARQL
bodies be called inside SPARQL and participate in `FILTER`/`ORDER BY` with engine
push-down — replacing the downstream interim fallback of pre-binding scored
candidates into a scratch world.

## What this branch delivers

- `UserFunctionRegistry::register_native(iri, arity, volatility, body)` where
  `body: NativeFnBody = Arc<dyn Fn(&[&TermValue]) -> Result<TermValue, EvalError>
  + Send + Sync>`. The closure receives its arguments as a slice of **borrows**
  into the evaluator's per-call argument buffer — no per-call deep clone of the
  heap-string-owning `TermValue`s on the per-row scoring hot path.
- Unified dispatch: `Function::Custom` resolves an IRI against the SPARQL-bodied
  table first, then the native table, then the closed XSD-cast set. One eval site,
  two kinds.
- `Volatility {Stable, Volatile}` (`#[non_exhaustive]`, after PostgreSQL's
  `provolatile`) declared at registration and consumed by the parallel gate:
  `Volatile` pins the call to sequential evaluation; `Stable` may run across
  fork-join workers.
- Owned pre-existing bug fixed: the fork-join parallel-safety classifier is now
  registry-aware, so a SPARQL-bodied `Function::Custom` (which re-enters the
  evaluator and could diverge under parallel evaluation) is correctly classified
  parallel-unsafe.
- Fail-fast contracts: declared `Arity` checked before the closure runs;
  `catch_unwind` converts a panicking closure to a fixed, payload-free
  `EvalError` (no nondeterministic rayon abort, no payload leak); cross-kind IRI
  collision hard-fails at registration; `EvalError::Function` Display generalized
  from "SHACL-AF function error" to "user function error"; `ORDER BY` over
  `xsd:double` scores including `NaN` is a stable total order.
- Public re-exports: `Arity`, `NativeFnBody`, `NativeFunction`, `Volatility`.

## Files changed (9 files, +1087/-103)

- `crates/sparql-eval/src/user_fn.rs` — native registry table, `register_native`,
  `NativeFunction`/`NativeFnBody`/`Arity`/`Volatility`, `eval_native_function`
  (arity check + borrow collection + panic guard), collision guards, module +
  type rustdoc covering both function kinds, and the native-scorer test suite.
- `crates/sparql-eval/src/parallel.rs` — registry-aware parallel-safety
  classifier (native `Volatility` + SPARQL-bodied ⇒ unsafe) + tests.
- `crates/sparql-eval/src/expr.rs` — native dispatch arm in `Function::Custom`;
  classifier call sites thread `ctx.user_functions`.
- `crates/sparql-eval/src/binop.rs`, `modifier.rs` — classifier call sites pass
  the registry; NaN total-order in ORDER BY key comparison.
- `crates/sparql-eval/src/error.rs` — generalized `EvalError::Function` Display.
- `crates/sparql-eval/src/lib.rs` — public re-exports.
- `crates/gts/src/compact.rs`, `crates/rdf/src/gts_certify.rs` — demoted
  pre-existing private intra-doc links (`base64url_unpadded`,
  `try_refold_digest`/`try_effective_digest`) to code spans; these hard-erroring
  under the workspace `RUSTDOCFLAGS="-D warnings"` doc gate kept CI's `doc` job
  RED. Fixed here per "when you see a bug, you own it."

## Requirements satisfied (issue #91 Completeness Contract)

R1 register_native API · R2 unified Function::Custom dispatch · R3 SPARQL-bodied
path unchanged · R4 determinism contract documented · R5 purity declared &
machine-checkable · R6 parallel gate honors native purity · R6b SPARQL-bodied
classified parallel-unsafe (owned bug) · R7 push-down in FILTER/ORDER BY · R8
Send+Sync bound · E-arity · E-panic · E-collision · E-display · E-nan · AC wasm32
clean · AC public re-exports — ALL implemented and test-covered.

No Cargo features added; no dependency pins in the member crate; no optional
components; no swallowed errors (the one `.expect` is guarded by a proven
all-Some invariant); determinism preserved; wasm32 stays clean (`catch_unwind`
compiles to a no-op under wasm `panic=abort`); no oxigraph/PyO3 near core;
`example.org` fixtures only. Native args are lent as borrows (MAXIMAL
PERFORMANCE); the public `NativeFnBody` signature was fixed to `&[&TermValue]`
pre-release (GREENFIELD/no-compromises), not left to a breaking post-release
change.

## Completion-Adversary — STAGE 3 (pre-merge): PASS

Audited the merge-into-main result (origin/main unchanged; `git merge` a no-op).
Every acceptance criterion DEMONSTRATED-IN-PRODUCTION through the REAL public
evaluate path (`NativeSparqlEngine::query_with_user_functions`), not a private
shim. Representative demonstrations (observed via `cargo test -p purrdf-sparql-eval`,
335 lib tests + integration suites, 0 failed):

- R1/R2: `native_function_returns_value`, `native_and_sparql_bodied_coexist`
  (one query calling both kinds, each returns 2).
- R6/R6b/R7: `native_score_filter_over_parallel_threshold` (1500 rows >
  PARALLEL_MIN_ROWS, Stable scorer routed to fork-join, byte-stable answer);
  `native_volatile_custom_is_parallel_unsafe`, `sparql_bodied_custom_is_parallel_unsafe`
  (production `is_parallel_safe`/`function_is_unsafe` classifier, not a test shim);
  `native_score_in_filter_pushes_down`, `native_score_in_order_by_is_deterministic`.
- E-arity/panic/collision/display/nan: each driven by its dedicated engine test.

GATES: cargo test PASS (335 passed, 0 failed) · clippy --all-targets PASS
(warning-free) · make doc PASS (exit 0, workspace, -D warnings) · make wasm PASS.

Laundering scan: the issue's "Enhancements Considered and Declined" (structural
registry merge) and the "future body-safety walk" note both CLEARED the innocence
test — neither is a #91 acceptance criterion, the headline runs end-to-end without
either, and the collision guard neutralizes the merge's sole correctness argument.
`gh issue list author:@me`: no follow-up issue carries #91's native-fn work.

## Deferral gate: PASS

Mechanical diff scan clean (no TODO/FIXME/unimplemented/stub/future-work added).
PR body/comment keyword hits are all discussion text asserting the ABSENCE of
deferrals ("NOT deferred work", "recorded, not deferred", "no follow-up issue was
filed"; "follow-up commit" = a later commit within this same PR) — false
positives, not deferred work.

- A native closure declared `Stable` MUST be deterministic within one query; the
  parallel gate trusts that declaration alone. A closure that lies about its
  volatility can silently diverge under fork-join. Retain a value past the call by
  cloning it inside the closure — the args are borrows.
- SPARQL-bodied UDFs in FILTER/ORDER BY/aggregates over >1024 rows now run
  sequentially (correct: the prior parallel path could diverge). Restoring
  parallelism for provably-pure bodies would require a recursive body-safety walk
  — a closed design decision, not deferred work.
- Reviewer's "add a non-panicking Result collision API" was rejected under .goals
  LOW/NO-OPTIONALITY + HARD-FAILS: a cross-kind IRI collision is host build-time
  misconfiguration; fail-fast is the intended contract.

## Conflicts resolved

None — `git merge origin/main` was a clean no-op (origin/main unchanged at the
branch's base).

PR #100, opened 2026-07-13
feat(gts): two-inventory diff/splice fetch-list + streaming RdfEventSink bridge on a shared decode core

### LINKED ISSUE #92: gts replication: inventory-diff → segment-range fetch helper and streaming-fold index-builder conveniences
## Context

A private downstream corpus (ADR 0031 Parts VII–VIII) synchronizes commissioned shard **cell ledgers** and backup mirrors using GTS's byte-range replication inventory, cat-append segments, and MMR proofs. Two conveniences would remove downstream boilerplate:

## Requests

1. **Inventory-diff helper**: given two replication inventories (local, remote), produce the minimal segment/byte-range fetch list to bring one up to date, verifying chain continuity across the splice point. Today callers hand-roll this over `replication::inventory` output.
2. **Streaming-fold index-builder hook**: the bounded-memory streaming fold (GTS-SPEC §7.7) emits events; an adapter that drives an `RdfEventSink` per segment with (segment index, frame content-id, byte range) context would let external index builders (posting lists, ClaimId→offset maps) be built in one pass without materializing datasets.

Both are additive conveniences over existing machinery; no format changes implied.

paudley:
Closes #92

## Summary
Two additive GTS replication conveniences over existing machinery — **no GTS format/wire change** — plus the shared structure that unifies them: a *content-addressed, byte-addressable event stream with provenance*, where `(content-id, ByteRange)` is the provenance unit at segment granularity (diff) and frame granularity (streaming).

### Feature 1 — two-inventory diff → minimal fetch list (`crates/gts/src/replication.rs`)
- `diff(local, remote) -> DiffResult` computes the minimal, deterministically-ordered `Vec<SegmentFetch>` of remote byte ranges to bring a local mirror current, using **segment-head-prefix ordering** (segments are independently rooted, so a whole-segment append's first frame `prev` chains to its own header, not the local head) with a frame-`prev` splice check for a partially-extended trailing segment. Continuity discontinuities → `Diverged`, never a silent fetch.
- `splice(local, remote, result) -> Result<Vec<u8>>` applies the fetch list and re-verifies the reconstruction inventories clean — the consumer's real payload.
- `diff_json` renders schema `gts-replication-diff-v1`, reusing the existing `range_json`. The head-locate scan is factored into a shared `find_frame` used by both `missing()` and `diff()`.

### Feature 2 — streaming RdfEventSink bridge with per-frame provenance
- **2a** (`reader.rs`): the streaming read path now surfaces per-frame `FrameContext { segment_index, frame_index, content_id, range, frame_type, valid }` (with a `verify(source)` self-check) via a default-no-op `StreamingSink::frame` callback; byte offsets are plumbed through `next_stream_item`. Non-streaming path untouched.
- **Shared core** (`segment_decode.rs`): the GTS-frame decode + per-segment two-phase term resolution is extracted into a target-agnostic `ResolvedSink` + `SegmentResolver<S>`, and `purrdf-rdf`'s `SinkImporter` is **refactored onto it** so the decode logic exists exactly once. Resolution is per-segment-at-close (bounded memory), behavior-identical to the previous global two-phase pass.
- **2b** (`event_stream.rs`): `stream_events(data, options, sink)` drives a `&mut dyn GtsEventSink` (an `RdfEventSink` + provenance hooks) on the shared core — segment-local ids → global `EventTermId`, per-segment blank scope → `ScopeId`, datatype-id → IRI string, quoted-triples bound. Honors `ControlFlow::Break` (a cancelled drive is never finalized) and **hard-errors** on graph-scoped reifiers/annotations (the RdfEventSink protocol has no graph slot) rather than silently dropping. Bounded-memory streaming fold per GTS-SPEC §7.7 — index builders (posting lists, ClaimId→byte-offset maps) run in a single pass with no dataset materialized.

## Changes
- `replication.rs`: `diff`/`splice`/`diff_json` + `DiffResult`/`DiffStatus`/`SegmentFetch`; shared `find_frame`; `FrameInventory.prev`.
- `model.rs`: `ByteRange` hoisted to a shared provenance type (`len`/`is_empty`), re-exported from `replication`.
- `reader.rs`: `FrameContext` + `StreamingSink::frame` + byte-offset plumbing.
- `segment_decode.rs` (new): `ResolvedSink` + `SegmentResolver<S>` shared decode core.
- `event_stream.rs` (new): `GtsEventSink` + `EventEmitter` + `stream_events`; `purrdf-events` workspace edge (zero-dep, wasm-clean, acyclic).
- `crates/rdf/src/gts_import_sink.rs`: `SinkImporter` refactored onto the shared core (net −561/+398).
- Tests: `replication_diff.rs` (10), `streaming_provenance.rs`, `event_bridge.rs` (9) — all on the public surface; existing `purrdf-rdf` GTS-import suite green unchanged.
- Bench: `streaming_sink_bench.rs` extended with a bridge-driven counting sink (report-only).
- Doc gate: repaired two pre-existing private-intra-doc-link errors (`compact.rs`, `gts_certify.rs`).

## Gates
`make check` (fmt + clippy pedantic/nursery + build + tests + hygiene), `make wasm`, and `make doc` all green. No Cargo features; determinism preserved (all orderings file-order); no issue refs in source.

## Summary by CodeRabbit

* **New Features**
  * Added streaming GTS → RDF event ingestion with a callback-based event sink, deterministic term declarations, and cancellation-safe operation.
  * Introduced frame provenance metadata (byte ranges, validity) during streaming, plus support for GTS event bridging.
  * Added replication diff/splice utilities with JSON reporting and byte-range calculations.
* **Bug Fixes**
  * Enforced stricter RDF 1.2/event-protocol validation (including graph-scoped reifier/annotation handling) and improved invalid-reference reporting.
* **Tests**
  * Expanded integration coverage for event bridging, provenance ordering, indexing, bounded buffering, cancellation, and replication diff/splice behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1314,6 +1314,7 @@ dependencies = [
  "fastrand",
  "filetime",
  "flate2",
+ "purrdf-events",
  "rayon",
  "serde",
  "sha1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1307,6 +1307,7 @@ name = "purrdf-gts"
 version = "0.5.0"
 dependencies = [
  "aes-gcm",
+ "ahash",
  "blake3",
  "ciborium",
  "criterion",
--- a/crates/gts/Cargo.toml
+++ b/crates/gts/Cargo.toml
@@ -27,6 +27,10 @@ workspace = true
 # This edge is acyclic — `purrdf-events` depends on nothing — so it never risks
 # a cycle and keeps `make wasm` green.
 purrdf-events = { workspace = true }
+# Fixed-key (no runtime RNG seeding) hasher for the segment-decode / event-bridge
+# dedup maps — mirrors the workspace's `FastHasher` policy (`crates/rdf-core/src/hash.rs`);
+# wasm-clean since `getrandom` is absent on wasm32 and `default-features = false` here.
+ahash = { workspace = true }
 # emojihash + randomart visual fingerprints, published from
 # Blackcat-Informatics/visual-hashing.
 visual-hashing = { workspace = true }
--- a/crates/gts/Cargo.toml
+++ b/crates/gts/Cargo.toml
@@ -22,6 +22,11 @@ all-features = true
 workspace = true
 
 [dependencies]
+# The zero-dependency, wasm-clean RDF 1.2 ingestion protocol (purrdf P6): the
+# neutral `RdfEventSink` seam the GTS event bridge (`event_stream`) folds into.
+# This edge is acyclic — `purrdf-events` depends on nothing — so it never risks
+# a cycle and keeps `make wasm` green.
+purrdf-events = { workspace = true }
 # emojihash + randomart visual fingerprints, published from
 # Blackcat-Informatics/visual-hashing.
 visual-hashing = { workspace = true }
--- a/crates/gts/examples/streaming_sink_bench.rs
+++ b/crates/gts/examples/streaming_sink_bench.rs
@@ -4,16 +4,21 @@
 //! Report-only streaming-sink throughput probe for the GTS writer; not a
 //! criterion gate — see `docs/BENCHMARKS.md` for the benchmarking policy.
 
+use core::ops::ControlFlow;
 use std::env;
 use std::fs;
 use std::fs::File;
 use std::process::ExitCode;
 
+use purrdf_events::{
+    EventError, EventQuad, EventTerm, EventTermId, EventTriple, RdfEventSink, ScopeId,
+};
+use purrdf_gts::event_stream::{GtsEventSink, stream_events};
 use purrdf_gts::model::{
     AnnotationRow, Diagnostic, OpaqueNode, Quad, ReifierRow, Signature, StreamableInfo,
     Suppression, Term,
 };
-use purrdf_gts::reader::{ReadOptions, StreamingSink, read_to_sink_from_reader};
+use purrdf_gts::reader::{FrameContext, ReadOptions, StreamingSink, read_to_sink_from_reader};
 
 #[derive(Default)]
 struct CountingSink {
@@ -81,6 +86,82 @@ impl StreamingSink for CountingSink {
     }
 }
 
+/// A counting `GtsEventSink` driven through the `event_stream` bridge, so the
+/// event path shares the streaming-sink probe. Report-only, like `CountingSink`.
+#[derive(Default)]
+struct CountingEventSink {
+    terms: usize,
+    quads: usize,
+    reifiers: usize,
+    annotations: usize,
+    frames: usize,
+    blobs: usize,
+}
+
+impl RdfEventSink for CountingEventSink {
+    fn term(
+        &mut self,
+        _id: EventTermId,
+        _term: EventTerm<'_>,
+    ) -> Result<ControlFlow<()>, EventError> {
+        self.terms += 1;
+        Ok(ControlFlow::Continue(()))
+    }
+
+    fn quad(&mut self, _q: EventQuad) -> Result<ControlFlow<()>, EventError> {
+        self.quads += 1;
+        Ok(ControlFlow::Continue(()))
+    }
+
+    fn reifier(
+        &mut self,
+        _reifier: EventTermId,
+        _triple: EventTriple,
+    ) -> Result<ControlFlow<()>, EventError> {
+        self.reifiers += 1;
+        Ok(ControlFlow::Continue(()))
+    }
+
+    fn annotation(
+        &mut self,
+        _reifier: EventTermId,
+        _p: EventTermId,
+        _o: EventTermId,
+    ) -> Result<ControlFlow<()>, EventError> {
+        self.annotations += 1;
+        Ok(ControlFlow::Continue(()))
+    }
+
+    fn open_scope(&mut self) -> Result<ScopeId, EventError> {
+        Ok(ScopeId::DEFAULT)
+    }
+
+    fn close_scope(&mut self, _scope: ScopeId) -> Result<ControlFlow<()>, EventError> {
+        Ok(ControlFlow::Continue(()))
+    }
+
+    fn finish(&mut self) -> Result<(), EventError> {
+        Ok(())
+    }
+}
+
+impl GtsEventSink for CountingEventSink {
+    fn frame(&mut self, _ctx: FrameContext<'_>) -> Result<ControlFlow<()>, EventError> {
+        self.frames += 1;
+        Ok(ControlFlow::Continue(()))
+    }
+
+    fn blob(
+        &mut self,
+        _ctx: Option<FrameContext<'_>>,
+        _digest: &str,
+        _meta: Option<&ciborium::value::Value>,
+    ) -> Result<ControlFlow<()>, EventError> {
+        self.blobs += 1;
+        Ok(ControlFlow::Continue(()))
+    }
+}
+
 fn linux_peak_kib() -> Option<usize> {
     let status = fs::read_to_string("/proc/self/status").ok()?;
     for line in status.lines() {
@@ -106,6 +187,31 @@ fn main() -> ExitCode {
     };
     let mut sink = CountingSink::default();
     let result = read_to_sink_from_reader(file, ReadOptions::new(true, None), &mut sink);
+
+    // Drive the event-bridge path over the same fixture so the `stream_events`
+    // seam shares this probe. Report-only; the counts are descriptive.
+    let (event_terms, event_quads, event_frames, event_blobs) = match fs::read(&path) {
+        Ok(bytes) => {
+            let mut event_sink = CountingEventSink::default();
+            match stream_events(&bytes, ReadOptions::new(true, None), &mut event_sink) {
+                Ok(_) => (
+                    event_sink.terms,
+                    event_sink.quads,
+                    event_sink.frames,
+                    event_sink.blobs,
+                ),
+                Err(err) => {
+                    eprintln!("streaming_sink_bench: event bridge: {err}");
+                    (0, 0, 0, 0)
+                }
+            }
+        }
+        Err(err) => {
+            eprintln!("streaming_sink_bench: re-read for event bridge: {err}");
+            (0, 0, 0, 0)
+        }
+    };
+
     let peak = linux_peak_kib().map_or_else(|| "null".to_string(), |value| value.to_string());
     println!(
         concat!(
@@ -123,6 +229,10 @@ fn main() -> ExitCode {
             "\"diagnostics\":{},",
             "\"segment_heads\":{},",
             "\"streamable_layouts\":{},",
+            "\"event_terms\":{},",
+            "\"event_quads\":{},",
+            "\"event_frames\":{},",
+            "\"event_blobs\":{},",
             "\"peak_kib\":{}",
             "}}"
         ),
@@ -137,6 +247,10 @@ fn main() -> ExitCode {
         result.diagnostics.len(),
         result.segment_heads.len(),
         result.segment_streamable.len(),
+        event_terms,
+        event_quads,
+        event_frames,
+        event_blobs,
         peak
     );
     ExitCode::SUCCESS
--- a/crates/gts/src/compact.rs
+++ b/crates/gts/src/compact.rs
@@ -250,8 +250,8 @@ fn base64url_unpadded(data: &[u8]) -> String {
 }
 
 /// Decode a base64url WITHOUT padding (RFC 4648 §5) string — the inverse of
-/// [`base64url_unpadded`], used to recover a `stream:cose` literal's raw
-/// COSE_Sign1 bytes (Task 5 signature verification).
+/// the private `base64url_unpadded` encoder, used to recover a `stream:cose`
+/// literal's raw COSE_Sign1 bytes for signature verification.
 ///
 /// # Errors
 /// HARD-ERRORS on a padding character (`=`) or any byte outside the unpadded
--- a/crates/gts/src/event_stream.rs
+++ b/crates/gts/src/event_stream.rs
@@ -212,6 +212,17 @@ impl<'s> EventEmitter<'s> {
         id
     }
 
+    /// Count of resolved IRI strings currently retained in `iri_map` (see the
+    /// field doc). `iri_map` is cleared on every genuine segment transition
+    /// ([`Self::ensure_scope`]), so reading this right after a segment has
+    /// been fully resolved reflects only THAT segment's own interned-IRI
+    /// count, never the cumulative total across every segment seen so far.
+    /// Exposed for tests asserting the bounded-memory contract (GTS-SPEC
+    /// §7.7) from outside the crate.
+    pub fn interned_iri_count(&self) -> usize {
+        self.iri_map.len()
+    }
+
     /// Ensure the scope for `segment_index` is the open one, closing the previous
     /// segment's scope and opening a fresh one on a segment change. A no-op once
     /// cancelled (beyond recording the segment).
--- a/crates/gts/src/event_stream.rs
+++ b/crates/gts/src/event_stream.rs
@@ -58,7 +58,6 @@
 //! `tests/event_bridge.rs`).
 
 use core::ops::ControlFlow;
-use std::collections::HashMap;
 
 use ciborium::value::Value;
 use purrdf_events::{
@@ -69,6 +68,13 @@ use purrdf_events::{
 use crate::model::{ByteRange, Diagnostic, OpaqueNode, Signature, StreamableInfo, Suppression};
 use crate::reader::{FrameContext, ReadOptions, StreamingReadResult, read_to_sink_with_options};
 use crate::segment_decode::{ResolvedSink, SegmentResolver};
+
+/// A [`std::collections::HashMap`] keyed by the workspace's fixed-key `ahash`
+/// policy (`crates/rdf-core/src/hash.rs`'s `FastHasher`) — no runtime RNG
+/// seeding, so it stays wasm-clean. Iteration order is unspecified; this map is
+/// only ever point-looked-up, never iterated for egress.
+type FastMap<K, V> =
+    std::collections::HashMap<K, V, core::hash::BuildHasherDefault<ahash::AHasher>>;
 
 /// Well-known `xsd:string` datatype IRI implied by a plain literal (RDF §7.1).
 ///
@@ -185,7 +191,7 @@ pub struct EventEmitter<'s> {
     next_id: u32,
     /// Resolved IRI text by minted id, retained so a literal's datatype id can be
     /// resolved back to its IRI string for [`EventTerm::Literal`].
-    iri_map: HashMap<EventTermId, String>,
+    iri_map: FastMap<EventTermId, String>,
     /// The segment whose scope is currently open, if any.
     current_segment: Option<usize>,
     /// The scope opened for [`Self::current_segment`].
@@ -215,7 +221,7 @@ impl<'s> EventEmitter<'s> {
         Self {
             sink,
             next_id: 0,
-            iri_map: HashMap::new(),
+            iri_map: FastMap::default(),
             current_segment: None,
             current_scope: ScopeId::DEFAULT,
             cached: None,
--- a/crates/gts/src/event_stream.rs
+++ b/crates/gts/src/event_stream.rs
@@ -10,6 +10,12 @@
 //! GTS-specific frames (blobs, opaque nodes, signatures, suppressions) that have
 //! no place in the neutral RDF event vocabulary.
 //!
+//! It rides the bounded-memory streaming fold (GTS-SPEC §7.7 "Streaming fold and
+//! bounded memory"): the reader delivers one frame at a time and the decode core
+//! buffers only the current segment's descriptors, so an external index builder
+//! (posting lists, a ClaimId→byte-offset map) runs in a single pass without ever
+//! materializing the union dataset.
+//!
 //! # Id and scope mapping
 //!
 //! The decode core resolves each segment's terms in a deterministic
--- a/crates/gts/src/event_stream.rs
+++ b/crates/gts/src/event_stream.rs
@@ -239,7 +239,7 @@ impl<'s> EventEmitter<'s> {
 
     /// Count of resolved IRI strings currently retained in `iri_map` (see the
     /// field doc). `iri_map` is cleared on every genuine segment transition
-    /// ([`Self::ensure_scope`]), so reading this right after a segment has
+    /// (`ensure_scope`), so reading this right after a segment has
     /// been fully resolved reflects only THAT segment's own interned-IRI
     /// count, never the cumulative total across every segment seen so far.
     /// Exposed for tests asserting the bounded-memory contract (GTS-SPEC
--- a/crates/gts/src/event_stream.rs
+++ b/crates/gts/src/event_stream.rs
@@ -92,10 +92,11 @@ const RDF_LANG_STRING: &str = "http://www.w3.org/1999/02/22-rdf-syntax-ns#langSt
 ///
 /// Every GTS hook defaults to a no-op `Continue`, so a consumer that only cares
 /// about RDF terms and quads implements just [`RdfEventSink`] and inherits inert
-/// GTS hooks. Each provenance-bearing hook receives the [`FrameContext`] cached
-/// from the most recently streamed frame — `None` before the first frame has
-/// been announced, since the reader fires a frame's provenance *after* that
-/// frame's rows.
+/// GTS hooks. The reader announces a frame's provenance via [`Self::frame`]
+/// *before* that frame's rows, so each provenance-bearing hook receives the
+/// [`FrameContext`] of the frame it belongs to — the cache holds the current
+/// frame by the time these hooks fire. (A hook's context is `None` only when no
+/// frame provenance is available at all.)
 pub trait GtsEventSink: RdfEventSink {
     /// Per-frame byte/identity provenance, announced once per frame.
     fn frame(&mut self, _ctx: FrameContext<'_>) -> Result<ControlFlow<()>, EventError> {
--- a/crates/gts/src/event_stream.rs
+++ b/crates/gts/src/event_stream.rs
@@ -38,6 +38,24 @@
 //! *graph-scoped* reifier or annotation therefore cannot be represented, and the
 //! bridge raises a hard [`purrdf_events::EventError`] rather than silently
 //! dropping the graph — honoring the no-swallow doctrine.
+//!
+//! The raw GTS reader ([`crate::reader`]) is a deliberately permissive
+//! Baseline Reader (GTS-SPEC §7.4/§7.5, GTS-CONFORMANCE.md §6): a row whose
+//! subject/predicate/object/graph-name/datatype/reifier names a segment-local
+//! term id no `term` event ever introduced is diagnosed (`PositionConstraint`
+//! or `ForwardReference`) and DROPPED, and folding continues so a damaged or
+//! partially-authored file still yields its recoverable content. That
+//! degrade-and-continue policy is right for the raw byte reader, but the RDF
+//! event protocol promises every id it ever declares is usable — a row that
+//! silently vanishes is indistinguishable from a fact that was never
+//! asserted. [`EventEmitter::diagnostic`] therefore escalates exactly those
+//! two diagnostic codes to a hard [`purrdf_events::EventError`] (the same
+//! "genuinely dangling term id" failure [`stream_events`] documents), while
+//! every other reader diagnostic (frame damage, chain breaks, unknown codecs,
+//! …) still passes through to [`GtsEventSink::gts_diagnostic`] unharmed —
+//! that class of degradation is not a dangling reference and stays governed
+//! by the permissive-read contract (see `claimid_offset_index_single_pass` in
+//! `tests/event_bridge.rs`).
 
 use core::ops::ControlFlow;
 use std::collections::HashMap;
@@ -563,6 +581,28 @@ impl ResolvedSink for EventEmitter<'_> {
     fn diagnostic(&mut self, diag: &Diagnostic) -> Result<(), Self::Error> {
         if self.cancelled {
             return Ok(());
+        }
+        // `PositionConstraint` / `ForwardReference` mean the RAW GTS reader
+        // (§7.4/§7.5) just silently DROPPED a row — a quad/reifier/annotation
+        // whose subject/predicate/object/graph-name/dt/rf named a segment-local
+        // term id no `term` event ever introduced — rather than resolving it.
+        // GTS's Baseline Reader is deliberately permissive about this (it
+        // degrades and keeps folding survivors, per GTS-CONFORMANCE.md §6), but
+        // the RDF event protocol is stricter: every id it ever declares MUST be
+        // usable, and a row silently vanishing from the stream is exactly the
+        // "genuinely dangling term id" this module's docs already promise is an
+        // `Err`, never a silent skip. Escalate here rather than passing it
+        // through as an informational `gts_diagnostic` — otherwise the drive
+        // would finish "successfully" having quietly dropped the row, which is
+        // indistinguishable from that row never having been asserted.
+        if matches!(
+            diag.code.as_str(),
+            "PositionConstraint" | "ForwardReference"
+        ) {
+            return Err(EventError::message(format!(
+                "GTS dangling term reference ({}): {}",
+                diag.code, diag.detail
+            )));
         }
         // A frame-scoped diagnostic carries the cached frame provenance; a
         // file-level diagnostic (e.g. `EmptyFile`) carries none.
--- a/crates/gts/src/event_stream.rs
+++ b/crates/gts/src/event_stream.rs
@@ -1,0 +1,599 @@
+// SPDX-FileCopyrightText: 2026 Blackcat Informatics® Inc. <paudley@blackcatinformatics.ca>
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! Bridge a GTS file onto the permissive RDF 1.2 ingestion protocol
+//! ([`purrdf_events::RdfEventSink`]).
+//!
+//! [`stream_events`] drives a GTS file through the shared two-phase decode core
+//! ([`crate::segment_decode::SegmentResolver`]) into a [`GtsEventSink`] — an
+//! [`RdfEventSink`] that also receives per-frame GTS provenance and the
+//! GTS-specific frames (blobs, opaque nodes, signatures, suppressions) that have
+//! no place in the neutral RDF event vocabulary.
+//!
+//! # Id and scope mapping
+//!
+//! The decode core resolves each segment's terms in a deterministic
+//! `(segment_index, gts_id)` order (segment-major, `gts_id`-sorted, inner-triple
+//! first), calling the sink's `intern_*` methods exactly once per term. The
+//! bridge mints a fresh, monotonically increasing [`EventTermId`] on each intern
+//! — so id assignment is a pure function of file order, reproducible across runs.
+//! Because every id is declared exactly once, the protocol never sees an
+//! [`purrdf_events::EventError::RedeclaredId`].
+//!
+//! Per-segment blank-node scope (the same `_:b0` in two segments names two
+//! different nodes) is preserved by opening a fresh [`ScopeId`] on the first
+//! intern of each segment and closing the previous one, mirroring GTS segment
+//! isolation onto the protocol's scope namespacing.
+//!
+//! # Faithful failures
+//!
+//! The RDF event protocol carries no graph slot on a reifier binding or a
+//! statement annotation ([`EventTriple`] is a bare triple). A GTS
+//! *graph-scoped* reifier or annotation therefore cannot be represented, and the
+//! bridge raises a hard [`purrdf_events::EventError`] rather than silently
+//! dropping the graph — honoring the no-swallow doctrine.
+
+use core::ops::ControlFlow;
+use std::collections::HashMap;
+
+use ciborium::value::Value;
+use purrdf_events::{
+    EventError, EventQuad, EventTerm, EventTermId, EventTriple, RdfEventSink, ScopeId,
+    TextDirection,
+};
+
+use crate::model::{ByteRange, Diagnostic, OpaqueNode, Signature, StreamableInfo, Suppression};
+use crate::reader::{FrameContext, ReadOptions, StreamingReadResult, read_to_sink_with_options};
+use crate::segment_decode::{ResolvedSink, SegmentResolver};
+
+/// Well-known `xsd:string` datatype IRI implied by a plain literal (RDF §7.1).
+///
+/// This is RDF's own datatype IRI, mirrored from
+/// `crates/rdf/src/native_codecs/hextuples.rs`; it is NOT fabricated PurRDF
+/// vocabulary.
+const XSD_STRING: &str = "http://www.w3.org/2001/XMLSchema#string";
+/// Well-known `rdf:langString` datatype IRI implied by a language tag (RDF §7.1).
+///
+/// RDF's own datatype IRI (see [`XSD_STRING`]), not fabricated vocabulary.
+const RDF_LANG_STRING: &str = "http://www.w3.org/1999/02/22-rdf-syntax-ns#langString";
+
+/// A GTS-aware event sink: an [`RdfEventSink`] that also receives per-frame GTS
+/// provenance and the GTS-specific frames that have no neutral RDF event.
+///
+/// Every GTS hook defaults to a no-op `Continue`, so a consumer that only cares
+/// about RDF terms and quads implements just [`RdfEventSink`] and inherits inert
+/// GTS hooks. Each provenance-bearing hook receives the [`FrameContext`] cached
+/// from the most recently streamed frame — `None` before the first frame has
+/// been announced, since the reader fires a frame's provenance *after* that
+/// frame's rows.
+pub trait GtsEventSink: RdfEventSink {
+    /// Per-frame byte/identity provenance, announced once per frame.
+    fn frame(&mut self, _ctx: FrameContext<'_>) -> Result<ControlFlow<()>, EventError> {
+        Ok(ControlFlow::Continue(()))
+    }
+
+    /// An inline blob's content digest and declared public metadata.
+    fn blob(
+        &mut self,
+        _ctx: Option<FrameContext<'_>>,
+        _digest: &str,
+        _meta: Option<&Value>,
+    ) -> Result<ControlFlow<()>, EventError> {
+        Ok(ControlFlow::Continue(()))
+    }
+
+    /// An opaque frame (unknown codec, missing key, or damaged payload).
+    fn opaque(
+        &mut self,
+        _ctx: Option<FrameContext<'_>>,
+        _node: &OpaqueNode,
+    ) -> Result<ControlFlow<()>, EventError> {
+        Ok(ControlFlow::Continue(()))
+    }
+
+    /// A signature observation on a frame.
+    fn signature(
+        &mut self,
+        _ctx: Option<FrameContext<'_>>,
+        _sig: &Signature,
+    ) -> Result<ControlFlow<()>, EventError> {
+        Ok(ControlFlow::Continue(()))
+    }
+
+    /// A suppression directive.
+    fn suppression(
+        &mut self,
+        _ctx: Option<FrameContext<'_>>,
+        _suppression: &Suppression,
+    ) -> Result<ControlFlow<()>, EventError> {
+        Ok(ControlFlow::Continue(()))
+    }
+
+    /// A reader diagnostic. Frame-scoped diagnostics carry the cached
+    /// [`FrameContext`]; file-level diagnostics carry `None`.
+    fn gts_diagnostic(
+        &mut self,
+        _ctx: Option<FrameContext<'_>>,
+        _diagnostic: &Diagnostic,
+    ) -> Result<ControlFlow<()>, EventError> {
+        Ok(ControlFlow::Continue(()))
+    }
+}
+
+/// An owned snapshot of the most recently streamed [`FrameContext`], kept so the
+/// provenance-bearing [`GtsEventSink`] hooks can borrow it after the reader's
+/// borrowed context has expired.
+struct CachedFrame {
+    segment_index: usize,
+    frame_index: usize,
+    content_id: Vec<u8>,
+    range: ByteRange,
+    frame_type: String,
+    valid: bool,
+}
+
+impl CachedFrame {
+    /// Reconstruct a borrowed [`FrameContext`] over this cached snapshot.
+    fn as_context(&self) -> FrameContext<'_> {
+        FrameContext {
+            segment_index: self.segment_index,
+            frame_index: self.frame_index,
+            content_id: &self.content_id,
+            range: self.range.clone(),
+            frame_type: &self.frame_type,
+            valid: self.valid,
+        }
+    }
+}
+
+/// A [`ResolvedSink`] that folds resolved GTS terms and rows onto a
+/// [`GtsEventSink`], minting a fresh [`EventTermId`] per term and rotating a
+/// [`ScopeId`] per segment.
+///
+/// Constructible through [`EventEmitter::new`] so callers that cannot go through
+/// [`stream_events`] — chiefly tests exercising a decode failure the GTS
+/// [`crate::writer::Writer`] refuses to author — can drive it behind a
+/// [`SegmentResolver`] directly.
+pub struct EventEmitter<'s> {
+    /// The bridged consumer.
+    sink: &'s mut dyn GtsEventSink,
+    /// The next [`EventTermId`] to mint; incremented on every intern.
+    next_id: u32,
+    /// Resolved IRI text by minted id, retained so a literal's datatype id can be
+    /// resolved back to its IRI string for [`EventTerm::Literal`].
+    iri_map: HashMap<EventTermId, String>,
+    /// The segment whose scope is currently open, if any.
+    current_segment: Option<usize>,
+    /// The scope opened for [`Self::current_segment`].
+    current_scope: ScopeId,
+    /// The most recently streamed frame's provenance.
+    cached: Option<CachedFrame>,
+    /// Set once any sink call requests [`ControlFlow::Break`]; further sink
+    /// emission is suppressed so a cancelled drive is never finalized.
+    cancelled: bool,
+}
+
+impl std::fmt::Debug for EventEmitter<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("EventEmitter")
+            .field("next_id", &self.next_id)
+            .field("interned_iris", &self.iri_map.len())
+            .field("current_segment", &self.current_segment)
+            .field("current_scope", &self.current_scope)
+            .field("cancelled", &self.cancelled)
+            .finish_non_exhaustive()
+    }
+}
+
+impl<'s> EventEmitter<'s> {
+    /// Wrap a [`GtsEventSink`] in a fresh emitter.
+    pub fn new(sink: &'s mut dyn GtsEventSink) -> Self {
+        Self {
+            sink,
+            next_id: 0,
+            iri_map: HashMap::new(),
+            current_segment: None,
+            current_scope: ScopeId::DEFAULT,
+            cached: None,
+            cancelled: false,
+        }
+    }
+
+    /// Mint the next monotonically increasing [`EventTermId`].
+    fn mint(&mut self) -> EventTermId {
+        let id = EventTermId(self.next_id);
+        self.next_id += 1;
+        id
+    }
+
+    /// Ensure the scope for `segment_index` is the open one, closing the previous
+    /// segment's scope and opening a fresh one on a segment change. A no-op once
+    /// cancelled (beyond recording the segment).
+    fn ensure_scope(&mut self, segment_index: usize) -> Result<(), EventError> {
+        if self.current_segment == Some(segment_index) {
+            return Ok(());
+        }
+        if self.cancelled {
+            self.current_segment = Some(segment_index);
+            return Ok(());
+        }
+        if self.current_segment.is_some()
+            && self.sink.close_scope(self.current_scope)? == ControlFlow::Break(())
+        {
+            self.cancelled = true;
+            self.current_segment = Some(segment_index);
+            return Ok(());
+        }
+        self.current_scope = self.sink.open_scope()?;
+        self.current_segment = Some(segment_index);
+        Ok(())
+    }
+
+    /// Emit a term declaration, latching cancellation on a `Break`.
+    fn emit_term(&mut self, id: EventTermId, term: EventTerm<'_>) -> Result<(), EventError> {
+        if self.cancelled {
+            return Ok(());
+        }
+        if self.sink.term(id, term)? == ControlFlow::Break(()) {
+            self.cancelled = true;
+        }
+        Ok(())
+    }
+
+    /// Close the scope left open by the final segment, if any. Called once by
+    /// [`stream_events`] before finalizing a non-cancelled drive.
+    fn close_open_scope(&mut self) -> Result<(), EventError> {
+        if self.cancelled {
+            return Ok(());
+        }
+        if self.current_segment.is_some() {
+            if self.sink.close_scope(self.current_scope)? == ControlFlow::Break(()) {
+                self.cancelled = true;
+            }
+            self.current_segment = None;
+        }
+        Ok(())
+    }
+
+    /// Resolve a literal's datatype id (or absence) to its expanded datatype IRI.
+    ///
+    /// A `Some` datatype id that never resolved to an IRI is a hard error (a
+    /// literal cannot be typed by a non-IRI). Absence defaults per RDF §7.1:
+    /// `rdf:langString` with a language tag, else `xsd:string`.
+    fn datatype_iri(
+        &self,
+        datatype: Option<EventTermId>,
+        lang: Option<&str>,
+    ) -> Result<String, EventError> {
+        match datatype {
+            Some(dt) => {
+                self.iri_map.get(&dt).cloned().ok_or_else(|| {
+                    EventError::message("GTS literal datatype must resolve to an IRI")
+                })
+            }
+            None if lang.is_some() => Ok(RDF_LANG_STRING.to_owned()),
+            None => Ok(XSD_STRING.to_owned()),
+        }
+    }
+}
+
+/// Parse a GTS literal base-direction token into a [`TextDirection`].
+///
+/// Mirrors `crates/rdf/src/gts_resolve.rs::parse_gts_direction`: `None` is
+/// legitimate absence, an unrecognized token is a hard error, and RDF 1.2 admits
+/// a base direction only on a non-empty language-tagged string.
+fn parse_direction(
+    direction: Option<&str>,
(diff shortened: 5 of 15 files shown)


paudley:
# Implementation Plan for Issue #92

## Context
A downstream consumer (a private downstream corpus, ADR 0031 Parts VII–VIII) synchronizes
shard ledgers and backup mirrors on top of GTS byte-range replication. Today it
hand-rolls two chunks of boilerplate that GTS should own:

1. Diffing a **local** replication inventory against a **remote** one to compute
   the minimal segment/byte-range fetch list, verifying the fetched suffix
   chains continuously onto the local head across the splice point.
2. Driving an RDF event sink from the bounded-memory streaming fold
   (GTS-SPEC §7.7) so external index builders (posting lists, ClaimId→byte-offset
   maps) can be built in a **single pass without materializing datasets**, with
   each event tagged by `(segment index, frame content-id, byte range)`.

Both are **additive conveniences over existing machinery** — **no GTS
format/wire change**. Outcome: callers delete boilerplate and call GTS APIs whose
determinism, chain-validation, and wasm-cleanliness the library guarantees.

### Architectural throughline
Both features are views of one structure: a **content-addressed, byte-addressable
event stream with provenance**. The unit of provenance is `(content-id, ByteRange)`
— surfaced at **segment** granularity by the diff (a `SegmentFetch`) and at
**frame** granularity by the streaming adapter (a `FrameContext`). The id/`prev`
hash chain is the cryptographic linkage both the splice-continuity check and the
streaming provenance ride on. One shared `ByteRange` provenance type threads both
features. The diff's fetch ranges are directly drivable through the streaming
bridge, so a consumer can fetch a suffix and **re-index only the fetched
segments** — the composition the issue under-asks for, delivered here.

- Adversarial (not sycophantic) self-review; assume code is substandard until proven otherwise.
- **NO deferrals / followups / future-work.** All work on this branch, now.
- **BANNED:** GitHub issue/PR numbers or process-flow references in source or repo docs (`// Per #92` etc.). Process flow lives in GitHub only.
- "When you see a bug, you own it" — fix in place.

- **Greenfield / SOTA / no compromises**; **Rust-first, Python-surface**.
- **Subsume, extend, enhance**; **low/no optionality, hard fails** (no optional deps, no swallowed errors).
- **Maximal utility · performance · portability (wasm-core).**
- Project rules (CLAUDE.md/AGENTS.md): **no Cargo features**; no hand-edits to `generated/`/`vectors/`; deps pinned only in root `[workspace.dependencies]`; **no nondeterminism** in serializers or the GTS reader/writer; no oxigraph/PyO3 near `purrdf-core`; **every release crate stays `wasm32-unknown-unknown`-clean** (`make wasm`); brand = **PurRDF** in prose, `purrdf` in identifiers; PurRDF mints no vocabulary IRIs (fixtures use `example.org`).

## Completeness Contract
Every requirement maps to a numbered task; nothing deferred/stubbed. Each
acceptance criterion is an **integration test on the public `purrdf_gts::` (or
`purrdf::`) surface** under `crates/<crate>/tests/*.rs` — reaching symbols only
through the public path, so a test fails to compile if a symbol is not exported.
Inline `#[cfg(test)]` unit tests may supplement but do **not** satisfy an
acceptance row (they reach private items and would mask an unexported surface).

| Requirement / Acceptance Criterion (executable, public-surface) | Task |
|---|---|
| **R1** `purrdf_gts::replication::diff(local, remote)` returns a minimal, deterministically-ordered `Vec<SegmentFetch>` of remote byte ranges | T2 |
| **R1-exec** Integration test: `diff` a `local` prefix vs `remote`, feed the returned ranges to `splice()`, assert reconstruction **byte-equals `remote`** and re-inventories clean | T2 |
| **R2** Diff verifies id/`prev` **and segment-head-prefix** continuity at the splice; a discontinuity → `Diverged`/`continuous:false`, never a silent fetch | T1, T2 |
| **R2-exec** Tests: (a) whole-segment append → `Fetch`/`continuous:true`; (b) mid-segment cat-append with forged `prev` → `Diverged`; (c) re-encoded common segment (byte-length mismatch) → `Diverged` | T2 |
| **R3** Edge cases: identical→`Current`; empty local→fetch-all; empty/behind remote→`Current`; fork→`Diverged`; `has_problems()`→`Error` | T2 |
| **R4** `diff_json` renders schema `gts-replication-diff-v1` (reusing `range_json`); `diff`/`diff_json`/`DiffResult`/`SegmentFetch`/`DiffStatus`/`splice` are `pub` in `replication` and reachable as `purrdf_gts::replication::*` (identical exposure to `missing`; there is **no** GTS CLI/bin — the public library API is the surface) | T2 |
| **R5** Streaming path surfaces per-frame `(segment index, frame content-id, byte range, frame_type, valid)` via `StreamingSink::frame` + `FrameContext`, plus a `FrameContext::verify(bytes)` self-check | T3 |
| **R5-exec** Integration test: a capturing `StreamingSink` asserts each frame's `(content_id, range, frame_type, valid)` equal the `FrameInventory` values `inventory()` reports for the same bytes; `verify(bytes)` recomputes the content-id and passes | T3 |
| **R6** Shared per-segment decode core (`SegmentResolver` + emit trait) in `purrdf-gts`; **`SinkImporter` (crates/rdf) refactored onto it** so the decode logic exists **once** (subsume) | T4 |
| **R6-exec** Existing GTS-import tests (`crates/rdf`) still pass unchanged after the refactor (regression proof the subsumed core preserves behavior) | T4 |
| **R6-exec2** Integration test (public surface): `stream_events` / `import_gts_events` over a fixture whose frame references an **undeclared** segment-local term id returns `Err` (dangling-binding hard-fail) — no silent skip, no partial index (`.goals` no-swallowed-errors) | T4/T5 |
| **R7** `GtsEventSink` bridge (`stream_events`) drives `&mut dyn RdfEventSink` on the shared core: segment-local ids→global `EventTermId`, per-segment scope→`ScopeId`, datatype-id→IRI resolved, quoted-triples bound, all frame types surfaced, `finish()` once | T5 |
| **R7-exec** Integration test: bridge a fixture into a concrete sink; resolved quads/terms **equal the quad set from non-streaming `read()`** of the same bytes | T5 |
| **R7-exec2** Two-segment fixture reusing blank label `_:b0` in quads → per-segment `ScopeId` keeps the two nodes **distinct** | T5 |
| **R8** Bounded memory: bridge holds only the id map + scope state + **current-segment** resolution buffer (triple/reifier bindings + segment term table) — no whole-file `Graph`; resolves and drops at each segment close | T5 |
| **R8-exec** Integration test: a `GtsEventSink` whose `frame` hook records `content_id → range.start` builds a ClaimId→offset index in one drive that **equals** `{FrameInventory.id → ByteRange.start}` from `inventory()` (valid frames only); a damaged-frame fixture still surfaces a diagnostic, not a silent gap | T5 |
| **R9** Compose: fetched whole-segment ranges from `diff` drive cleanly through `stream_events` for incremental re-index (segment-boundary aligned) | T5 |
| **R10** No wire/format change; `make check` + `make wasm` pass; no Cargo features; no issue refs in source; `ensure_id` id-assignment is a deterministic function of file order (tested) | T6 |

- **MMR-verifiable provenance coordinates (transformational, DECLINED).** Tagging
  each `SegmentFetch`/`FrameContext` with its MMR-proof position. *Why the lesser
  structure suffices:* the issue scopes provenance to **chain continuity** and
  forbids format/proof coupling. For a **contiguous suffix** fetch the consumer
  recomputes every frame id/`prev` between splice and head — that recompute **is**
  the inclusion proof, so an MMR proof is strictly redundant. Non-contiguous
  (multi-segment forest) verification, if ever needed, rides the **already-built**
  `aggregate_digest` / `gts-segment-heads-v1` commitment (`replication.rs:331`),
  not new proof machinery. Closed design decision, not future work.
- **Rolling-hash (rsync-style) approximate diff (DECLINED).** Content-addressed
  exact diff already **subsumes** approximate rolling-hash diff for this domain
  (frames are content-addressed; equality is exact and cheap). A rolling-hash
  layer would add cost and nondeterminism risk for no gain.
- **Public inventory algebra `compare`/`meet` for N-way mirror reconciliation
  (DECLINED for this branch, would need user sign-off to add).** The two-inventory
  `diff` is the issue's ask; a general lattice `meet` over N inventories is a
  distinct capability. `diff` is built on a shared `find_frame` primitive so the
  generalization is cheap later, but adding the N-way surface now is scope the
  issue does not ask for — not built, not stubbed.
  *(Reversal note: the earlier decline to reuse `gts_import_sink` is withdrawn —
  T4 now subsumes it into a shared core per the confirmed decision.)*

1. Resolve repo/paths:
   ```bash
   BRANCH="paudley/92-gts-diff-and-event-stream"
   WORKTREE="$REPO_ROOT/.worktrees/92-gts-diff-and-event-stream"
   ```
2. Locally exclude `.worktrees/`:
   ```bash
   ```
3. Fresh branch off up-to-date main, enter worktree:
   ```bash
   cd "$WORKTREE"; git branch --show-current   # MUST show the feature branch
   ```
4. Warm build: `cargo build -p purrdf-gts -p purrdf-rdf` (don't gate on full tests — that's stage2/3).
5. Post plan: `gh issue comment 92 --body-file /home/paudley/.claude/plans/92-wiggly-nest.md`.

**⚠️ Do NOT proceed until on the feature branch inside the worktree.**

## Task 1: Shared `ByteRange` provenance seam + `FrameInventory.prev`
Foundation for both features.
- Move `ByteRange` (`replication.rs:79`) to `crates/gts/src/model.rs` (neutral, beside `Diagnostic`/`StreamableInfo`); keep `Clone, Debug, PartialEq, Eq`; add inherent `len()`/`is_empty()` (half-open `[start,end)`). `pub use crate::model::ByteRange;` in `replication.rs` keeps `replication::ByteRange` paths valid.
- Add `pub prev: Option<Vec<u8>>` to `FrameInventory` (`replication.rs:18`); populate in `collect_frames` (`replication.rs:191`) from the frame map's `"prev"` bytes (already read for `prev_ok` at line 229 — capture instead of discard). Seed/non-map frames → `None`.

### After Task 1: github-commit-agent → github-push-agent → post update to issue #92.

## Task 2: Feature 1 — two-inventory `diff()` (segment-head-prefix model) + `splice()` + `diff_json`
Location: `crates/gts/src/replication.rs`. Public types mirror `MissingResult`/`MissingStatus`:
```rust
pub enum DiffStatus { Current, Fetch, Diverged, Error }   // PartialEq, Eq
pub struct SegmentFetch { pub remote_index: usize, pub range: ByteRange, pub head: Option<Vec<u8>> } // Clone,Debug,PartialEq,Eq
pub struct DiffResult {
    pub status: DiffStatus,
    pub fetch: Vec<SegmentFetch>,     // ordered by remote_index (file order → deterministic)
    pub splice_offset: Option<usize>, // byte offset in LOCAL file where the suffix appends
    pub continuous: bool,
    pub detail: Option<String>,
}
```
- Factor `find_frame(inventory, id) -> Option<(seg, frame)>` and reuse it in **both** `missing()` and `diff` (the head-locate scan is currently inlined in `missing`, replication.rs:501-549) — DRY (L1).
- `pub fn diff(local: &Inventory, remote: &Inventory) -> DiffResult` — **segment-head-prefix ordering is primary; frame-`prev` only for a partially-extended trailing segment** (segments are independently rooted, so a whole-segment append's first frame `prev` chains to its own header, NOT to `local_head`):
  1. Gate: either `has_problems()` → `Error` (which side + `problem_detail`). Expose `problem_detail` as a `pub(crate)` accessor (keep it off the public API).
  2. Empty local → fetch all remote segments, `splice_offset=Some(0)`, `continuous=true`, `Current` iff remote also empty.
  3. Compute `k` = longest common prefix of segment heads (content-id equality). For each common index assert byte-length alignment (`local.segments[i].(start,end) == remote's`); a mismatch under the deterministic writer signals a re-encoded mirror → `Diverged` with detail.
  4. Trichotomy: `k==local.len()==remote.len()` → `Current`; `k==local.len()<remote.len()` → `Fetch` remote suffix `remote.segments[k..]` (per-segment `SegmentFetch`, contiguous `[remote.segments[k].start, remote.clean_end)`), `continuous=true` (prefix equality is the proof), `splice_offset=Some(local.clean_end)`; `k==remote.len()<local.len()` → remote behind → `Current`; else `k<local.len()`: if local's only divergent segment is its **last** and its frames are a frame-id prefix of remote's segment `k` → **mid-segment cat-append**: fetch `[local.clean_end, remote.clean_end)`, verify the first new remote frame is `valid` **and** `prev == local head` (using `FrameInventory.prev`) → `continuous`; if local's segment `k` has ≥ frames → `Current`; otherwise → `Diverged`.
- `pub fn splice(local_bytes: &[u8], remote_bytes: &[u8], result: &DiffResult) -> Result<Vec<u8>, String>` (U1): apply the fetch list — copy each `SegmentFetch.range` from `remote_bytes` onto the local prefix at `splice_offset`, re-run `inventory()` on the result and hard-fail if not clean or if continuity is violated. This is the consumer's real payload (removes caller boilerplate) and the executable engine behind R1-exec.
- `pub fn diff_json(result: &DiffResult) -> String`: schema `"gts-replication-diff-v1"`, trailing `\n`, reuse `range_json`/`json_optional_hex`; emit `status`/`fetch`/`splice_offset`/`continuous`/`detail`. Match `missing_json` style.
- Ensure `diff`, `splice`, `diff_json`, `DiffResult`, `SegmentFetch`, `DiffStatus` are `pub` and re-exported wherever `missing`/`missing_json` are (incl. any `purrdf` umbrella re-export). No Cargo feature.
- **Integration tests** (`crates/gts/tests/replication_diff.rs`, public paths only): `diff_reconstructs_remote_via_splice` (R1-exec, byte-equal + clean re-inventory); `diff_appends_new_segment_is_fetch` (R2-exec-a, the case the old plan would have mis-flagged); `diff_midsegment_forged_prev_is_diverged` (R2-exec-b); `diff_reencoded_common_segment_is_diverged` (R2-exec-c); `diff_identical_is_current`; `diff_empty_local_fetches_all`; `diff_remote_behind_is_current`; `diff_fork_is_diverged`; `diff_problem_inventory_is_error`; `diff_json_schema_and_shape`.

### After Task 2: github-commit-agent → github-push-agent → post update to issue #92.

## Task 3: Feature 2a — frame provenance on the streaming path
Location: `crates/gts/src/reader.rs` (+ `crate::model::ByteRange`).
```rust
pub struct FrameContext<'a> {
    pub segment_index: usize,
    pub frame_index: usize,
    pub content_id: &'a [u8],
    pub range: ByteRange,
    pub frame_type: &'a str,   // (U2) lets an index builder filter without re-decoding
    pub valid: bool,           // (U2) chain-validity, so an index skips damaged frames
}
impl FrameContext<'_> { pub fn verify(&self, source: &[u8]) -> bool { /* recompute content_id over source[range] (T4) */ } }
```
- Add default-no-op `fn frame(&mut self, _ctx: FrameContext<'_>) {}` to `StreamingSink` (backward compatible; existing sinks unaffected).
- Plumb byte offsets: change private `next_stream_item` to return `Ok(Some((item, start, end)))` (drops `start` today; torn `Err(start)` unchanged); thread `(start,end)` into `ActiveStreamingSegment::process_frame`; fire `sink.frame(ctx)` right after `content_id` is computed (`reader.rs:1261`), with a per-segment `frame_index` counter. Header items are not frames.
- Non-streaming `read`/`read_file_segments` path untouched (Feature 2 is streaming-only).
- **Integration test** (`crates/gts/tests/streaming_provenance.rs`): capturing `StreamingSink` over a 2-segment fixture asserts each `(content_id, range, frame_type, valid)` equals the `FrameInventory` values from `inventory()`; `verify(bytes)` passes for every frame.

### After Task 3: github-commit-agent → github-push-agent → post update to issue #92.

## Task 4: Shared per-segment decode core (SUBSUME `SinkImporter`)
Extract the GTS-frame decode + per-segment two-phase resolution into one reusable
core in `purrdf-gts`, then refactor the existing importer onto it so the logic
exists **once** (confirmed decision; `.goals` no-duplication).
- New module `crates/gts/src/segment_decode.rs`. A generic emit trait, e.g.:
  ```rust
  pub trait ResolvedSink {
      type Id: Copy;
      fn open_scope(&mut self) -> Self::ScopeId; fn close_scope(&mut self, s: Self::ScopeId);
      fn iri(&mut self, iri: &str) -> Self::Id;
      fn blank(&mut self, label: &str, scope: Self::ScopeId) -> Self::Id;
      fn literal(&mut self, lexical: &str, datatype: Self::Id, lang: Option<&str>, dir: Option<&str>) -> Self::Id;
      fn triple(&mut self, s: Self::Id, p: Self::Id, o: Self::Id) -> Self::Id;
      fn quad(&mut self, q: (Self::Id, Self::Id, Self::Id, Option<Self::Id>));
      fn reifier(&mut self, r: Self::Id, t: (Self::Id, Self::Id, Self::Id), g: Option<Self::Id>);
      fn annotation(&mut self, r: Self::Id, p: Self::Id, o: Self::Id, g: Option<Self::Id>);
      fn frame(&mut self, ctx: FrameContext<'_>);
      fn blob(&mut self, ctx: FrameContext<'_>, digest: &str, meta: Option<&Value>);
      fn opaque(&mut self, ctx: FrameContext<'_>, node: &OpaqueNode);
      fn signature(&mut self, ctx: FrameContext<'_>, sig: &Signature);
      fn suppression(&mut self, ctx: FrameContext<'_>, s: &Suppression);
      fn diagnostic(&mut self, d: &Diagnostic);
  }
  ```
- `SegmentResolver<S: ResolvedSink>` implements `StreamingSink`: buffers the **current segment**'s raw terms + reifier bindings + quad/reifier/annotation rows (keyed by segment-local id), opens a scope at segment start, and at each **segment close** resolves that segment inner-first (datatype-id→IRI, quoted-triple `reifier`→`(s,p,o)` binding, depth-bounded by the existing `MAX_GTS_TERM_NESTING_DEPTH`), emits resolved rows via `ResolvedSink`, then drops the segment buffers. Per-segment (not per-file) → bounded memory. Hard-fail (return via a captured error) on a genuinely dangling id/binding after the segment's events are all seen — no silent skip.
- Refactor `crates/rdf/src/gts_import_sink.rs::SinkImporter` to implement `ResolvedSink` (`Id = TermId`) over `RdfDatasetBuilder`, driven by `SegmentResolver`, deleting its now-duplicated two-phase code. Preserve its public entry (`import_gts_events`) and semantics.
- **Tests:** existing `crates/rdf` GTS-import tests pass unchanged (R6-exec regression). The dangling-binding hard-fail is proven on the **public surface** (R6-exec2) via an integration test in `crates/gts/tests/event_bridge.rs` (and/or `crates/rdf` import tests): a fixture whose frame references an undeclared segment-local term id drives `stream_events`/`import_gts_events` and asserts an `Err` with no partial output. An inline `SegmentResolver` unit test may supplement but does not satisfy the row.

### After Task 4: github-commit-agent → github-push-agent → wiring-verifier → post update to issue #92.

## Task 5: Feature 2b — `GtsEventSink` bridge on the shared core
Location: new module `crates/gts/src/event_stream.rs` (`pub mod event_stream;`). Add `purrdf-events = { workspace = true }` to `crates/gts/Cargo.toml` (sanctioned by the `purrdf-events` crate docs; zero-dep ⇒ wasm-clean; no cycle).
```rust
pub trait GtsEventSink: RdfEventSink {
    fn frame(&mut self, _ctx: FrameContext<'_>) -> Result<ControlFlow<()>, EventError> { Ok(ControlFlow::Continue(())) }
    fn blob(&mut self, _c: FrameContext<'_>, _digest: &str, _meta: Option<&Value>) -> Result<ControlFlow<()>, EventError> { Ok(ControlFlow::Continue(())) }
    fn opaque(&mut self, _c: FrameContext<'_>, _n: &OpaqueNode) -> Result<ControlFlow<()>, EventError> { Ok(ControlFlow::Continue(())) }
    fn signature(&mut self, _c: FrameContext<'_>, _s: &Signature) -> Result<ControlFlow<()>, EventError> { Ok(ControlFlow::Continue(())) }
    fn suppression(&mut self, _c: FrameContext<'_>, _s: &Suppression) -> Result<ControlFlow<()>, EventError> { Ok(ControlFlow::Continue(())) }
    fn diagnostic(&mut self, _c: Option<FrameContext<'_>>, _d: &Diagnostic) -> Result<ControlFlow<()>, EventError> { Ok(ControlFlow::Continue(())) } // (U3) let a sink Break on BrokenChain/DamagedFrame immediately, no swallow
}
```
- A private `EventEmitter<'s>` implements the Task-4 `ResolvedSink` (`Id = EventTermId`) driving `&mut dyn RdfEventSink`: mint a fresh global `EventTermId` per resolved term (monotonic `next_id`); map GTS blank scope → `RdfEventSink::open_scope()`/`close_scope` (capture the returned `ScopeId`, tag that segment's blanks). Because the shared core resolves inner-first per segment, terms are declared before reference (no forward refs needed; matches `FrozenDatasetSource`).
- `pub fn stream_events(data: &[u8], options: ReadOptions<'_>, sink: &mut dyn GtsEventSink) -> Result<StreamingReadResult, EventError>`: build `SegmentResolver<EventEmitter>`, drive via `read_to_sink_with_options`, call `sink.finish()` exactly once, return the `StreamingReadResult`. Honor `ControlFlow::Break` (stop immediately, return the `EventError`); never treat a cancelled drive as complete.
- **Determinism (R10):** `EventTermId` assignment must be a deterministic function of file order; use the crate's fixed-key `ahash` (not `RandomState`) for any dedup map. Tested invariant.
- **Integration tests** (`crates/gts/tests/event_bridge.rs`, public paths): `bridge_events_equal_offline_read` (R7-exec: resolved quads/terms == `read()` quads); `per_segment_blank_scopes_are_distinct` (R7-exec2); `claimid_offset_index_single_pass` (R8-exec: `content_id→range.start` map, valid frames only, == `inventory()` offsets) + a damaged-frame variant asserting a surfaced diagnostic; `bridge_calls_finish_once`; `bridge_break_cancels_drive`; `stream_ids_are_file_order_deterministic` (two drives → identical id assignment); `incremental_reindex_from_diff_fetch` (R9: `diff` a suffix, `splice`+`stream_events` the fetched whole segments, assert the delta index equals re-inventorying only those segments).
- **Bench (report-only, no asserted speedup):** extend `crates/gts/examples/streaming_sink_bench.rs` with a bridge-driven counting `GtsEventSink` for coverage.

### After Task 5: github-commit-agent → github-push-agent → wiring-verifier → post update to issue #92.

## Task 6: Gate, docs, wasm, hygiene
- Rustdoc for every new public item (`diff`, `splice`, `SegmentFetch`, `DiffResult`, `diff_json`, `FrameContext`(+`verify`), `StreamingSink::frame`, `ResolvedSink`, `SegmentResolver`, `GtsEventSink`, `stream_events`). §7.7 cross-reference in the streaming-adapter module docs (prose only). **No issue/PR numbers anywhere in source/docs** (`.baseline` ban).
- Verify re-exports reachable wherever sibling symbols are (umbrella included).
- `cargo fmt --all` first (fmt gate), then `make check` (fmt + clippy pedantic/nursery warning-free + build + tests + hygiene) and `make wasm`. Fix findings in place.
- Mechanical deferral scan over the branch diff: `grep -nEi 'todo|fixme|unimplemented|future work|for now|follow-?up|later'` — resolve every hit.

### After Task 6: github-commit-agent → github-push-agent → post update to issue #92.

1. Verify all committed/pushed.
2. `gh pr create --title "feat(gts): two-inventory diff/splice fetch-list + streaming RdfEventSink bridge on a shared decode core" --body "Closes #92 ..."`.
3. Post plan to PR: `PR_NUM=$(gh pr list --head "$(git branch --show-current)" --json number -q '.[0].number'); gh pr comment $PR_NUM --body-file /home/paudley/.claude/plans/92-wiggly-nest.md`.
4. Post to PR **and** issue: (a) What am I least confident about and why? (b) What do I not know but should?
5. Report: "Stage 1 complete. PR #<num> created for issue #92. Ready for /stage2 92".

## Execution Rules
1. Task 0 mandatory; never work on `main`. 2. Commit after EACH task (github-commit-agent), never batch. 3. Push after each commit (github-push-agent). 4. Task N (PR) mandatory & last. 5. Sub-agents for all commits/pushes. 6. NO deferrals — every contract row done on this branch.

## Verification (end-to-end)
- **Feature 1:** `diff`→`splice` reconstructs `remote` byte-for-byte; whole-segment-append, mid-segment forged-prev, and re-encoded-mirror cases exercised.
- **Feature 2a:** capturing sink cross-checks streaming `(content_id, range, type, valid)` against offline `inventory()`; `verify(bytes)` passes.
- **Feature 2b:** bridge output == offline `read()`; per-segment scopes distinct; one-pass ClaimId→offset index == inventory offsets; deterministic id assignment; incremental re-index from a diff fetch.
- **Subsume:** existing GTS-import tests green after `SinkImporter` sits on the shared core.
- **Gates:** `make check` + `make wasm` green; no Cargo features; no issue refs in source; determinism preserved (all orderings file-order).

paudley:
## Stage 1 reflection

1. **The `SinkImporter` per-segment-flush equivalence (Task 4).** The subsume converts the importer's global two-phase resolution into per-segment-at-close resolution. I argued (and the full `purrdf-rdf` suite confirms) this is behavior-identical because GTS segments are independently rooted and delivered in order, so segment-major key order yields the same frozen term allocation order. The risk is a corner the existing tests don't exercise — e.g. an exotic multi-segment file where resolution order subtly matters. The regression suite is the guard, but it is only as good as its coverage of multi-segment/quoted-triple/datatype-cross-reference shapes.

2. **RdfEventSink fidelity boundaries on the bridge.** Two GTS constructs cannot be represented on the neutral event protocol and are handled by *hard error* (no silent loss): graph-scoped reifiers/annotations (`EventTriple` has no graph slot). This is the correct no-swallow choice, but it means a valid GTS file using named-graph reification cannot flow through `stream_events` — such callers must use the IR import path (`import_gts_events`). If the downstream consumer expected the event bridge to carry those, that's a mismatch to flag early.

3. **The mid-segment forged-`prev` `Diverged` branch is proven with a synthetic `Inventory`, not real bytes**, because genuine wire corruption trips the reader's `BrokenChain` fold diagnostic first (→ `Error`, not `Diverged`). The `diff` code path is correct and covered; I'm simply noting that in practice a tampered suffix surfaces as `Error`, and `Diverged` is reached for *structural* forks (independent segments / frame-prefix mismatch), which real-byte fixtures do cover.

1. **The exact downstream consumption shape (ADR 0031 Parts VII–VIII).** I designed `SegmentFetch`/`splice`/`FrameContext` from the issue text; I don't know whether the consumer wants per-segment fetch granularity (delivered) vs. a single coalesced byte range, or whether its ClaimId→offset index keys on the frame content-id (what `FrameContext` provides) or on some higher-level claim identifier that would need an extra mapping.

2. **Whether graph-scoped reifiers actually occur in the target corpora.** If they do and the consumer needs them over the event path, the hard-error boundary (#2 above) becomes a real limitation worth revisiting; if they never occur, it's a non-issue. I don't have that data.

3. **Cross-mirror re-encoding.** `diff` flags a common-head segment with mismatched byte length as `Diverged` (re-encoded mirror). Under the deterministic writer this shouldn't happen, but if mirrors are ever re-compacted with a different codec/dict, every mirror would look diverged — I don't know whether the deployment guarantees byte-stable re-serialization across mirrors.

paudley:
# Stage 2 Gap Remediation — Issue #92 / PR #100

Branch `paudley/92-gts-diff-and-event-stream` · worktree `.worktrees/92-gts-diff-and-event-stream`

## Context

Issue #92 adds two additive GTS conveniences: (1) a two-inventory **`diff` → fetch-list + `splice`**
and (2) a **streaming index-builder hook** — an `RdfEventSink` bridge driven from the bounded-memory
streaming fold (GTS-SPEC §7.7). Stage 1 shipped all six tasks; the feature surface (R1–R10) is present
and every `R*-exec` row has a public-surface integration test in the right place. `make check` / `make
wasm` / `make doc` are green.

Adversarial gap analysis (gap-analyzer + completion-adversary, both opus, run in parallel) found the
**headline promise of request #2 — bounded memory — is not actually delivered**. The streaming decode
core silently retains three whole-file lookup maps, so on a many-segment cat-appended ledger (the exact
replication workload this issue targets) memory grows O(whole file), not O(one segment). The green
R8-exec test only checks index *equality* against `inventory()`; it never measures memory, which is why
the defect slipped. This is a vacuous-fold: R8 is claimed but never executed. A Gemini review on PR #100
flagged two of the three maps; the analysis confirmed both and found a third that Gemini missed.

Two acceptance rows are also demonstrated only against hand-built input the real byte pipeline cannot
produce. User decisions (recorded this session): the defensive `Diverged` branches (F3) are discharged
by **human sign-off + `.deficiencies` record**; the dangling hard-fail (F2) is discharged by **crafting
a real-byte fixture through the public path and pinning path behavior**.

Outcome: the streaming bridge's per-segment buffers provably drop at each segment close (bounded memory,
falsifiably tested), the no-swallow hard-fail is proven on the real byte path, and R10's fixed-key-ahash
directive is honored — with the two provably-unreachable defensive branches signed off and recorded.

## Standing constraints (`.goals`, inviolable)
GREENFIELD/SOTA/NO COMPROMISES · RUST-FIRST/PYTHON-SURFACE · SUBSUME/EXTEND/ENHANCE · LOW/NO OPTIONALITY +
HARD FAILS (no optional deps, **no swallowed errors**) · RDF 1.2 ONLY · MAXIMAL UTILITY/PERFORMANCE/
**PORTABILITY (wasm-core)**. No Cargo features; deps in root `[workspace.dependencies]` only; no
nondeterminism in the GTS reader/writer; **no issue/PR numbers in source or repo docs** (`.baseline`).

## Verified clean (not gaps)
Deferral-keyword scan clean (the three "placeholder" hits are the GTS domain term "non-map placeholder
frame"). No banned issue/PR refs in added lines. No-swallow honored everywhere else (dangling term,
missing reifier, unbound triple, nesting limit, graph-scoped reifier/annotation all hard-`Err`;
cancellation never finalizes). R6 subsume genuine — `SinkImporter` sits on the shared `ResolvedSink`
core with no duplicated/dead decode path; one core drives both production paths. Determinism holds
(order-sensitive iteration is `sort_unstable`'d; ids minted monotonically).

---

## Gaps (each fixed on this branch; commit-per-gap, no batching)

### G1 — CRITICAL · R8 · `reifier_bindings` grows whole-file
`crates/gts/src/segment_decode.rs:251`. Inserted per reifier (`:569`), read at `:510`;
`resolve_buffered` (`:355-407`) drains `raw_terms`/`raw_quads`/`raw_reifiers`/`raw_annotations` but
**never clears `reifier_bindings`** → O(total reifiers across all segments). Bindings are segment-local
and unread after their segment resolves.
**Fix:** clear `reifier_bindings` at the end of `resolve_buffered` (after the annotation drain, `:404`).

### G2 — CRITICAL · R8 · `remaps` retained whole-file for a test-only inspector
`crates/gts/src/segment_decode.rs:247`, documented "kept across flushes so `remap` can inspect it after
resolution." Inserted `:437`, read `:313`/`:420`; **never cleared** → O(total terms across all segments).
`remap()` (`:312`) is called **only** from `#[cfg(test)]` code — `gts_import_sink.rs:549-552`, `:615`,
`:678-679` — reading segment 0's remap *after* segment 1 flushed. A test convenience is forcing whole-file
retention on the production core. Within resolution `remaps` need only survive one `resolve_buffered`
call (terms resolve first, then quads/reifiers/annotations hit the memo); segment-local ids never cross
segments, so cross-segment persistence is not needed for correctness.
**Fix:** clear `remaps` at the end of `resolve_buffered` (alongside G1). Remove the whole-file `remap()`
inspection method from `SegmentResolver` (or scope it to the current segment). **Rework the three
`#[cfg(test)]` sites** to observe resolved node identity from the *sink/dataset output* (or a test-side
`ResolvedSink` that captures `(segment, gts_id) → id` as terms are pushed — whole-file state belongs in
the test harness, not the production core). Preserve every assertion (shared IRI → same id across
segments; same blank label → distinct ids; literal/quoted-triple id checks).

### G3 — CRITICAL · R8 · `EventEmitter.iri_map` grows whole-file
`crates/gts/src/event_stream.rs:170`. Inserted per IRI intern (`:325`), read by `datatype_iri` (`:276`);
`ensure_scope` (`:218`) rotates the `ScopeId` on a segment change but **never clears `iri_map`** → one
owned `String` retained per distinct IRI across all segments. A literal's datatype always resolves within
its own segment (inner-first per-segment resolution), so the map is safe to drop at the segment boundary.
**Fix:** clear `iri_map` on the segment transition inside `ensure_scope` (the branch at `:226-234`, per
the Gemini suggestion), so each segment's IRI strings drop at close.

### G4 — HIGH · F2 · dangling hard-fail not proven on the real byte path (no-swallow)
`crates/gts/tests/event_bridge.rs:661`, `crates/rdf/src/gts_import_sink.rs:855`. R6-exec2 requires the
undeclared-term hard-fail be shown through the public `stream_events`/`import_gts_events` **byte** path;
today it's proven only by hand-driving `SegmentResolver`'s methods with forged events. `.goals` forbids
swallowed errors, so the real path's behavior on a dangling reference must be pinned by a real-byte test.
**Fix (user decision — craft real-byte fixture + pin path):** first establish what the real reader does
with a byte fixture whose frame references an undeclared segment-local term id (hard-`Err` vs
diagnose-and-drop/partial). Hand-assemble (or corrupt a valid file into) real GTS bytes carrying such a
reference, drive them through the public `stream_events` **and** `import_gts_events`, and assert `Err`
with **no partial output**. If the path currently diagnoses-and-drops into a partial result, **fix the
path to hard-fail** (own the bug). If authoring such bytes is proven structurally impossible, STOP and
escalate for human sign-off — do not self-resolve.

### G5 — MEDIUM · R8 guard · no falsifiable bounded-memory test
The gap that hid G1–G3: R8-exec (`event_bridge.rs:472`) only checks index equality. Add a public-surface
test over a **many-segment** fixture asserting the resolver/emitter retained state stays **O(one
segment)** after each flush — the assertion that would have caught G1–G3 and that pins R8 going forward.
`SegmentResolver`/`EventEmitter` already expose sizes via `Debug` (`segment_decode.rs:265`,
`event_stream.rs:182`); **add `reifier_bindings` to `SegmentResolver`'s `Debug`** for observability. Test
must be falsifiable (fail if any map grows with segment count).

### G6 — MEDIUM · R10 · fixed-key ahash directive not honored
`crates/gts/src/event_stream.rs:43`, `crates/gts/src/segment_decode.rs:54` import
`std::collections::HashMap` (RandomState). The approved R10 row says "use the crate's fixed-key `ahash`
(not `RandomState`) for any dedup map," matching the project's standing wasm/HashDoS convention (interners
use fixed-key ahash because `getrandom` is absent on wasm32). Output determinism is *already* safe here
(order-sensitive iteration is sorted; `iri_map` is point-looked-up only) — this is directive + wasm-
portability alignment, not a determinism fix.
**Fix:** switch both maps to the crate's existing fixed-key ahash builder (reuse the interners' pattern —
do not introduce a new hasher). Keep sorting where present.

### G7 — HUMAN SIGN-OFF (recorded) · F3 · defensive `Diverged` branches are synthetic-only
`crates/gts/tests/replication_diff.rs:238` (forged-`prev`), `:295` (re-encoded common segment). Both
branches are unreachable from real clean bytes by construction (writer never forges `prev`; equal
content-id ⇒ equal bytes ⇒ equal length; tampering trips `BrokenChain`→`Error`). The safety invariant
"never a silent fetch" **is** demonstrated on real bytes (→ `Error`), and `Diverged` is reachable on real
bytes via `diff_fork_is_diverged`.
**Disposition (user decision — human sign-off + record):** keep the defensive branches and their
synthetic tests; **plan approval is the sign-off**. During remediation, APPEND a verbatim block to
`.deficiencies` recording that R2-exec-b/-c are covered by synthetic `Inventory` fixtures because the
branches are provably unreachable via real clean bytes, with the safety property demonstrated on real
bytes via `Error`. No code change beyond the record.

---

## Execution (per stage2 workflow)
For each of G1→G6 (in order): **Task→general-purpose (sonnet)** implement the fix (NO DEFERRALS — a fix
that stubs/downgrades/re-defers is a failed fix, re-dispatch) → **Task→github-commit-agent** (reads
PRECOMMIT_GUIDANCE.md, iterates to green) → **Task→github-push-agent** → post summary (gap, priority,
fix, commit hash) to **issue #92 AND PR #100** → **Task→wiring-verifier** for G1/G2/G3/G4. G7 is the
`.deficiencies` record (folded into the G-series commits, no separate feature commit).
G1+G2 share `resolve_buffered` and the `remap()` rework, so they may land in one focused commit; G3–G6
are independent commits. After all gaps: Phase 5 feedback loop (re-check PR comments), then **Phase 6 re-
run the completion-adversary (STAGE=2) as the exit gate** — must return PASS (R8 now
DEMONSTRATED-IN-PRODUCTION via G5; F2 via G4) or carry the G7 verbatim sign-off — then Phase 7
`.deficiencies` append.

## Verification (end-to-end, on the public surface)
- **G1–G3+G5:** `cargo test -p purrdf-gts --test event_bridge` incl. the new many-segment bounded-memory
  test; confirm `SegmentResolver`/`EventEmitter` map sizes stay O(one segment) after each flush (fails
  before the fix, passes after).
- **G2 rework:** `cargo test -p purrdf-rdf` — the reworked node-identity assertions pass observing the
  dataset/sink, not the internal memo; full import suite green (R6-exec regression).
- **G4:** new real-byte fixture through `stream_events` + `import_gts_events` returns `Err`, no partial
  output; `cargo test -p purrdf-gts --test event_bridge` + `-p purrdf-rdf`.
- **G6:** both modules on fixed-key ahash; `stream_ids_are_file_order_deterministic` still green;
  `make wasm` green.
- **Gates:** `cargo fmt --all` first, then `make check` + `make wasm` + `make doc` all green; no Cargo
  features; no issue/PR refs; determinism preserved.
- **Exit:** completion-adversary STAGE=2 re-run → PASS (or G7 sign-off recorded verbatim).

paudley:
## Gap G1+G2 resolved — CRITICAL (R8 bounded memory) · commit `ff76e09`

**Gap (adversary + gap-analyzer, confirmed in code):** the shared streaming decode core retained two per-segment maps for the whole file, so on a many-segment cat-appended ledger memory grew O(whole file), not O(one segment) — silently breaking the headline bounded-memory promise (R8 / GTS-SPEC §7.7).

- **G1** — `SegmentResolver.reifier_bindings` (`segment_decode.rs`) was inserted per reifier but never cleared → O(total reifiers across all segments).
- **G2** — `SegmentResolver.remaps` was retained across flushes *solely* to feed a `remap()` inspector called only from `#[cfg(test)]` code → O(total terms across all segments).

**Fix:**
- Clear both `reifier_bindings` and `remaps` at the end of `resolve_buffered` (each segment's state now drops at close). Within a single resolution pass the memo still works (terms → quads/reifiers/annotations reuse it); segment-local ids never cross segments, so cross-segment retention was never needed for correctness.
- Removed the whole-file `remap()` inspection method from the production core. Moved its cross-segment observation into a `#[cfg(test)]` `RecordingSink` that captures `(segment, gts_id) → TermId` in the test harness — whole-file state belongs in tests, not the production surface.
- Reworked the three test sites (`gts_import_sink.rs`) to read resolved node identity from the recording sink; every assertion preserved (shared IRI → same id across segments; same blank label → distinct ids; literal + quoted-triple ids).

**Verification:** `cargo test -p purrdf-rdf` 212/212 lib + all GTS-import integration green (R6-exec regression); `cargo test -p purrdf-gts` all suites green; full `make check` + `make wasm` clean; no `SegmentResolver::remap` callers remain.

paudley:
## Gaps G3 + G5 resolved — CRITICAL/MEDIUM (R8 bounded memory) · commits `afcd8bc`, `2a15f0d`

**G3 (CRITICAL, `afcd8bc`)** — `EventEmitter.iri_map` retained one owned IRI `String` per distinct IRI for the *whole file*. `EventEmitter::ensure_scope` now clears `iri_map` on each genuine segment transition, bounding the streaming bridge's retained IRI strings to one segment. Safe because a literal's datatype IRI is always re-interned within its own segment before the literal is emitted (ordering traced through the shared per-segment resolver: `ensure_scope` fires on the first `intern_iri` of the new segment, before any of that segment's IRIs are inserted).

**G5 (MEDIUM — the guard that was missing, `2a15f0d`)** — the leak shipped because *no test asserted bounded memory*; R8-exec (`claimid_offset_index_single_pass`) only checked index equality vs `inventory()`. Added:
- three read-only size accessors — `SegmentResolver::buffered_reifier_binding_count`, `SegmentResolver::buffered_remap_count`, `EventEmitter::interned_iri_count` — plus `buffered_reifier_bindings` in the `SegmentResolver` `Debug`.
- a falsifiable public-surface test `bounded_memory_across_many_segments` driving a **12-segment** fixture (5 distinct IRIs + 1 quad + 2 reifier bindings per segment, nothing reused). After each segment flush it asserts the resolver's two maps are back to **0** and the emitter's IRI map stays at **one segment's cardinality (5)** — not `N × segments`.

**Falsifiability confirmed empirically:** reverting the `reifier_bindings`/`remaps` clear makes the test fail at segment 1 (`2 vs 0`); reverting the `iri_map` clear makes it fail with the `5 × seg` growth pattern (`10 vs 5`). Restored, full suite green.

**Verification:** `cargo test -p purrdf-gts` all green incl. the new test; `make check` + `make wasm` clean; determinism preserved.

R8 "bounded-memory streaming fold (§7.7)" is now **demonstrated-in-production and falsifiably guarded**, not merely claimed.

paudley:
## Gap G4 resolved — HIGH (R6-exec2 / no-swallowed-errors) · commit `5d33d82`

**Finding is stronger than the gap suggested — a genuine swallow existed.** R6-exec2's dangling-reference hard-fail was proven only by hand-driving `SegmentResolver`, never through the real byte path. Investigating the byte path (per the "craft a real-byte fixture" decision) revealed:

1. The Stage-1 "not authorable" claim is **false**: `Writer::add_quads` does no id validation, so `add_quads(&[(0,1,99,None)])` with 3 terms declared produces a well-formed, correctly-hashed, correctly-chained GTS frame referencing an undeclared id — no byte surgery needed.
2. On that fixture the raw reader diagnoses `PositionConstraint` and **drops the row** (spec-mandated permissive read, §7.4/§7.5 — correct for the raw reader, pinned by frozen conformance vectors).
3. `import_gts_events` already hard-`Err`s (its `diagnostic` promotes any reader diagnostic — "the IR is the authority").
4. **`stream_events` genuinely swallowed it** — `EventEmitter::diagnostic` forwarded the diagnostic and continued, returning `Ok` with a *partial* index (the dangling quad silently vanished). A `.goals` no-swallow violation.

**Fix:** `EventEmitter::diagnostic` now escalates exactly the two dangling-reference codes (`PositionConstraint`, `ForwardReference`) to a hard `EventError` (matching `import_gts_events`), while every other diagnostic (`DamagedFrame`, `BrokenChain`, `TornAppendError`, `UnknownCodec`, …) still passes through to the sink's `gts_diagnostic` hook unchanged — so the raw reader's permissive behavior and the damaged-frame path (`claimid_offset_index_single_pass`, which trips `TornAppendError`) are untouched. On escalation the drive latches the error and `stream_events` returns it *before* `finish()` — no partial output.

**New real-byte public-surface tests** (`Writer`-authored bytes, not hand-driven): `dangling_term_ref_real_bytes_is_err` through `stream_events` (asserts `Err`, `finish_count == 0`, empty index) and through `import_gts_events` (asserts `Err` code `rdf-ir-gts-fold-diagnostic`). The pre-existing hand-fed resolver tests are re-documented as defense-in-depth.

**Verification (wiring-verifier confirmed):** escalation is on the live `stream_events` drive, narrow, no partial finalization; `import_gts_events` unchanged; `cargo test -p purrdf-gts -p purrdf-rdf` all green (213 rdf lib tests); `make check` + `make wasm` clean.

paudley:
## Gap G6 resolved — MEDIUM (R10 fixed-key ahash directive + wasm portability) · commit `3bff858`

**Gap:** the new `event_stream.rs` / `segment_decode.rs` maps used `std::collections::HashMap` (default `RandomState`), contrary to R10's "use the crate's fixed-key `ahash`" directive and the workspace policy (`rdf-core/src/hash.rs` `FastHasher = BuildHasherDefault<ahash::AHasher>`), which is fixed-key precisely because `getrandom` is absent on wasm32 and to avoid per-process nondeterminism.

**Fix:** added the `ahash` workspace edge to `purrdf-gts` (`default-features = false`, wasm-clean, no version pin in the member crate) and switched `SegmentResolver`'s `raw_terms`/`remaps`/`reifier_bindings` and `EventEmitter`'s `iri_map` to a local `FastMap` alias over `BuildHasherDefault<ahash::AHasher>`, mirroring `rdf-core`'s policy. No sorting changed.

**Determinism unaffected:** order-sensitive iteration in `resolve_buffered` is still `sort_unstable`'d and `iri_map` is point-looked-up only — egress order has never come from hash iteration. `stream_ids_are_file_order_deterministic` stays green.

**wasm-clean confirmed:** `cargo build -p purrdf-gts --target wasm32-unknown-unknown` clean; `ahash v0.8.12` pulled with **no `getrandom`** anywhere in the wasm compile graph. `make check` (incl. `make wasm`) green; `bounded_memory_across_many_segments` and all suites green.

paudley:
## Reviewer feedback (CodeRabbit) addressed — G8 (Major) + G9 (Minor) · commits `a0c7377`, `dc1969f`

CodeRabbit reviewed the branch and raised two findings; both were confirmed real and fixed on this branch (no deferral).

**G8 — MAJOR (functional correctness), `a0c7377`.** `StreamingSink::frame` is documented as firing "before its rows are processed," but `reader.rs::process_frame` called `fold_frame` (emitting the frame's term/quad/reifier/annotation rows) and only fired `sink.frame()` *afterward*. This wasn't just a contract nit: the event bridge's `EventEmitter::frame` **caches** per-frame provenance that the `blob`/`opaque`/`signature` callbacks read back — so those GTS-specific frames were being handed **stale previous-frame provenance**. Fixed by computing the `FrameContext` and firing `sink.frame(ctx)` (via `Folder::with_sink`) **before** any row-emitting call in every branch, and removing the redundant post-fold emission. Added a falsifiable ordering test `frame_event_precedes_its_own_row_events_for_every_frame` (interleaved event log asserting `frame()` precedes each frame's rows) — confirmed it fails on the pre-fix code with the exact stale-order pattern.

**G9 — MINOR (misleading test), `dc1969f`.** `directional_literal_lexical_lang_survive_sink_path` claimed "GTS cannot carry base direction" and asserted `direction == None` — false: `SinkImporter::intern_literal` parses and preserves RDF 1.2 base direction; the test only saw `None` because its input was `None`. Corrected the wording to reflect the input, and added `directional_literal_rtl_survives_sink_path` authoring a `direction: Some("rtl")` literal (with its required language tag) and asserting the resolved literal keeps `RdfTextDirection::Rtl`.

**Verification:** `make check` (incl. `make wasm`) green on both; `cargo test -p purrdf-gts -p purrdf-rdf` all green (rdf lib +2 tests → 214); the bridge/`import` paths regression-clean.

paudley:
## Stage 2 complete — completion-adversary FINAL EXIT gate: **PASS** · ready for /stage3

All gaps resolved on branch `paudley/92-gts-diff-and-event-stream`. The completion-adversary re-ran against the true final HEAD (`c888cb9`) and returned **PASS**: every R1–R10 acceptance criterion is DEMONSTRATED-IN-PRODUCTION on the public `purrdf_gts::`/`purrdf::` surface, with one recorded human-signed-off exception (G7/F3).

### Gaps addressed (each fixed on-branch, no deferrals)

| Gap | Sev | Commit | Resolution |
|---|---|---|---|
| **G1** reifier_bindings unbounded | CRIT (R8) | `ff76e09` | cleared at each segment close in `resolve_buffered` |
| **G2** remaps unbounded (test-only inspector) | CRIT (R8) | `ff76e09` | cleared at close; `remap()` removed; observation moved to `#[cfg(test)]` RecordingSink |
| **G3** iri_map unbounded | CRIT (R8) | `afcd8bc` | cleared on segment transition in `ensure_scope` |
| **G5** no bounded-memory test | MED (R8 guard) | `2a15f0d` | falsifiable 12-segment test + size accessors; falsifiability empirically proven |
| **G4** dangling hard-fail not on real byte path | HIGH (R6-exec2) | `5d33d82` | **found + fixed a genuine swallow** — `stream_events` dropped an undeclared-term quad and returned Ok; now escalates `PositionConstraint`/`ForwardReference` to hard `Err`; real-`Writer`-byte tests on both public paths |
| **G6** RandomState maps | MED (R10) | `3bff858` | switched to workspace fixed-key `ahash`; wasm-clean (no getrandom) |
| **G8** frame fired after rows (CodeRabbit Major) | HIGH | `a0c7377` | **fixed a real bridge bug** — `blob`/`opaque`/`signature` read stale provenance; `frame()` now fires before rows in every branch; falsifiable ordering test |
| **G9** misleading direction test (CodeRabbit Minor) | LOW | `dc1969f` | corrected wording + `rtl`-preservation test asserting `RdfTextDirection::Rtl` |
| **F1** orphaned GtsEventSink doc (adversary) | MIN | `17cc73f` | doc corrected to match before-rows ordering |
| **doc gate** public→private intra-doc links (self-caught) | MIN | `c888cb9` | plain code spans; `make doc` green |

### Signed-off exception (recorded verbatim in `.deficiencies`)
**G7 / F3** — the `diff()` `Diverged` branches R2-exec-b (forged-`prev`) and R2-exec-c (re-encoded common segment) are covered by synthetic `Inventory` fixtures because they are **provably unreachable from real clean bytes** (deterministic writer never forges `prev`; equal content-id ⇒ equal bytes ⇒ equal length; tampering trips `BrokenChain`→`Error`). The safety invariant "a discontinuity is never a silent fetch" **is** demonstrated on real bytes (→ `Error`), and `Diverged` is reachable on real bytes via `diff_fork_is_diverged`. User signed off; plan approval recorded as the sign-off.

### Final gate demonstrations (adversary, real production surface)
- **R8 bounded memory** — `bounded_memory_across_many_segments` ✓ (maps return to O(one segment) after each flush; falsifiability confirmed by reverting a clear)
- **R6-exec2 no-swallow** — `dangling_term_ref_real_bytes_is_err` ✓ on both `stream_events` and `import_gts_events` (real Writer bytes → `Err`, no partial index)
- **R5/R7 frame ordering** — `frame_event_precedes_its_own_row_events_for_every_frame` ✓, `bridge_events_equal_offline_read` ✓
- **R1/R3** — `diff_reconstructs_remote_via_splice`, `diff_fork_is_diverged`, `diff_problem_inventory_is_error` ✓
- **Gates** — `make check` / `make doc` / `make wasm` all exit 0; `cargo test -p purrdf-gts -p purrdf-rdf` all green; no Cargo features; no issue/PR refs in source; determinism preserved.

**Ready for /stage3.**

paudley:
feat(gts): two-inventory diff/splice fetch-list + streaming RdfEventSink bridge on a shared decode core

Closes #92.

## Summary

Two additive GTS conveniences over existing machinery — no format/wire change:

1. **Two-inventory `diff` → fetch-list + `splice`.** `purrdf_gts::replication::diff(local, remote)`
   returns a minimal, deterministically-ordered `Vec<SegmentFetch>` of the remote byte ranges needed
   to bring `local` up to date, verifying id/`prev` and segment-head-prefix continuity at the splice
   (a discontinuity yields `Diverged`/`Error`, never a silent fetch). `splice(local, remote, result)`
   applies the fetch list and re-inventories the reconstruction, and `diff_json` renders the
   `gts-replication-diff-v1` schema. Removes the caller boilerplate that hand-rolled this over
   `replication::inventory` output.

2. **Streaming index-builder hook.** A `GtsEventSink: RdfEventSink` bridge (`stream_events`) drives an
   event sink from the bounded-memory streaming fold (GTS-SPEC §7.7), tagging each event with per-frame
   `(segment index, frame content-id, byte range, frame_type, valid)` provenance so external index
   builders (posting lists, ClaimId→byte-offset maps) build in one pass without materializing datasets.
   The diff's fetch ranges compose directly through the bridge for incremental re-indexing.

Both ride one **shared per-segment decode core** (`SegmentResolver` + `ResolvedSink`) in `purrdf-gts`,
onto which `purrdf-rdf`'s `SinkImporter` was refactored (SUBSUME) so the GTS decode logic exists once.

## Files

- `crates/gts/src/replication.rs` — `diff`/`splice`/`diff_json` + `DiffResult`/`SegmentFetch`/`DiffStatus`;
  shared `find_frame` primitive reused by `missing()` and `diff()`.
- `crates/gts/src/segment_decode.rs` (new) — target-agnostic `ResolvedSink` trait + `SegmentResolver<S>`:
  per-segment-at-close two-phase resolution, bounded to one segment's memory.
- `crates/gts/src/event_stream.rs` (new) — `GtsEventSink` bridge, `EventEmitter` (`ResolvedSink` →
  `RdfEventSink`), `stream_events`.
- `crates/gts/src/reader.rs` — per-frame `FrameContext` provenance on the streaming path; `StreamingSink::frame`
  fired **before** a frame's rows.
- `crates/gts/src/model.rs` — `ByteRange` hoisted as the shared provenance type.
- `crates/rdf/src/gts_import_sink.rs` — `SinkImporter` refactored onto the shared core (SUBSUME); old
  duplicated two-phase decode deleted.
- Tests: `crates/gts/tests/{replication_diff,streaming_provenance,event_bridge}.rs` — public-surface
  integration tests for every acceptance row.
- `crates/gts/Cargo.toml` — `purrdf-events` (bridge seam) + `ahash` (fixed-key hasher) workspace edges.

## Requirements satisfied (R1–R10)

All rows demonstrated on the public library surface (there is no GTS CLI — the library API is the
surface). R1 diff→SegmentFetch list; R2 continuity check (discontinuity → Diverged/Error, never silent
fetch); R3 edge cases; R4 `gts-replication-diff-v1` + public exports; R5 per-frame provenance + `verify`;
R6 shared decode core + SinkImporter SUBSUME; R6-exec2 dangling hard-fail; R7 bridge == offline `read()`;
R8 bounded-memory streaming; R9 incremental re-index; R10 no wire change, file-order determinism, fixed-
key ahash, no Cargo features.

## Stage-2 gap remediation (this branch, all fixed on-branch — no deferrals)

Adversarial gap analysis found the **R8 bounded-memory promise was structurally unmet** (three whole-file
maps never dropped at segment close) and two rows demonstrated only against hand-built input. Fixed:

- **G1/G2 (CRIT, R8):** `SegmentResolver.reifier_bindings` + `remaps` now cleared at each segment close in
  `resolve_buffered`; whole-file `remap()` test-inspector removed, its observation moved into a
  `#[cfg(test)]` RecordingSink.
- **G3 (CRIT, R8):** `EventEmitter.iri_map` cleared on segment transition in `ensure_scope`.
- **G5 (R8 guard):** falsifiable 12-segment `bounded_memory_across_many_segments` test + size accessors;
  falsifiability empirically confirmed by reverting each clear.
- **G4 (HIGH, R6-exec2, no-swallow):** found + fixed a genuine swallow — `stream_events` dropped an
  undeclared-term quad and returned `Ok`; now escalates the reader's `PositionConstraint`/`ForwardReference`
  diagnostics to a hard `EventError`; real-`Writer`-byte tests pin `Err`+no-partial on both public paths.
- **G6 (R10):** new maps switched to the workspace fixed-key `ahash` (`BuildHasherDefault<ahash::AHasher>`);
  wasm-clean (no getrandom).
- **G8 (CodeRabbit Major):** `StreamingSink::frame` fired *after* a frame's rows, violating the documented
  contract and feeding the bridge's blob/opaque/signature callbacks stale previous-frame provenance; now
  fired before rows in every branch; falsifiable ordering test added.
- **G9 (CodeRabbit Minor):** corrected a misleading base-direction test that claimed GTS drops direction
  (it preserves it); added an rtl-preservation test.
- Two doc-accuracy fixes (F1 adversary-caught GtsEventSink ordering doc; public→private intra-doc links
  tripping the doc gate).

## Standing constraints (`.goals`) respected

GREENFIELD/SOTA · RUST-FIRST/PYTHON-SURFACE · SUBSUME (SinkImporter onto shared core) · LOW/NO OPTIONALITY
+ HARD FAILS (no-swallow enforced — dangling ref, graph-scoped reifier, missing binding, nesting limit all
hard-`Err`; G4 closed the one swallow) · RDF 1.2 ONLY · MAXIMAL PORTABILITY (wasm-core — `make wasm` green,
fixed-key ahash, no getrandom). No Cargo features; deps in root `[workspace.dependencies]` only; no
nondeterminism (all order from explicit id-sorting); no oxigraph/PyO3 near core; no issue/PR refs in source.

## Completion-Adversary — STAGE 3 PRE-MERGE: PASS (merge-into-main result, HEAD merge of c888cb9 + b7b7136)

Every R1–R10 criterion DEMONSTRATED-IN-PRODUCTION on the merged tree (no open findings):
- R1 `diff_reconstructs_remote_via_splice`; R2-a `diff_appends_new_segment_is_fetch`; R3
  identical/empty-local/behind/fork/problem; R4 `diff_json_schema_and_shape`; R5
  `streaming_frame_provenance_matches_offline_inventory`; R5/R7-ordering
  `frame_event_precedes_its_own_row_events_for_every_frame`; R6 SinkImporter rides shared
  `SegmentResolver`/`ResolvedSink` (no dup path); R6-exec2 `dangling_term_ref_real_bytes_is_err` (real
  Writer bytes → Err, no partial); R7 `bridge_events_equal_offline_read`; R7-exec2
  `per_segment_blank_scopes_are_distinct`; R8 `bounded_memory_across_many_segments` (falsifiable); R8-exec
  `claimid_offset_index_single_pass`; R9 `incremental_reindex_from_diff_fetch`; R10
  `stream_ids_are_file_order_deterministic` + wasm-clean.
- Gates on merged tree: `make check` 0, `make doc` (`-D warnings`) 0, `make wasm` 0, `cargo test -p
  purrdf-gts -p purrdf-rdf` all green.

### Human-signed-off exception (recorded verbatim in `.deficiencies`)
**G7/F3** — the `diff` `Diverged` branches R2-exec-b (forged-`prev`) and R2-exec-c (re-encoded common
segment) are covered by synthetic `Inventory` fixtures because provably unreachable from real clean bytes
(deterministic writer never forges `prev`; equal content-id ⇒ equal bytes ⇒ equal length; tampering trips
`BrokenChain`→`Error`). The safety invariant "a discontinuity is never a silent fetch" IS demonstrated on
real bytes (→ `Error`), and `Diverged` is reachable on real bytes via `diff_fork_is_diverged`. User
(paudley) signed off; recorded in `.deficiencies` (2026-07-13 Issue #92 block).

## Deferral gate: CLEAN (all hits false positives, evidenced)

- Diff hits: "non-map **placeholder** frame" ×3 — the GTS domain term for a structurally non-map frame,
  not deferred work.
- PR hits: a CI "Out of Scope Changes check ✅ Passed"; the Completeness Contract's "nothing deferred/
  stubbed"; "not future work" (a closed design decline); and the Task-6 text quoting the deferral-scan
  pattern itself. None is deferred work.
- Three enhancements documented as DECLINED design decisions (N-way inventory `meet`, MMR-coordinate
  provenance, rolling-hash approximate diff) are not in issue #92's text and the headline functions
  without them — innocence test passes, not laundering.

## Conflicts resolved

`crates/gts/src/compact.rs` (`base64url_decode` doc, ~line 252): both this branch and main independently
demoted the same private intra-doc link `[base64url_unpadded]` (denied by the doc gate) and dropped a
banned `(Task 5 …)` process reference. Kept this branch's clearer wording ("the inverse of the private
`base64url_unpadded` encoder … for signature verification") — a plain code span, no link, no process ref,
doc-gate clean. No behavioral difference between the two sides.

- The decode core resolves **per-segment-at-close** (bounded memory); segments are independently rooted so
  segment-major key order reproduces the previous global two-phase term order — regression-proven by the
  unchanged `purrdf-rdf` import suite.
- Bridge fidelity boundary: graph-scoped reifiers/annotations have no slot on the neutral `RdfEventSink`
  protocol and are handled by hard-`Err` (no-swallow); such callers use the IR import path
  (`import_gts_events`).
- The bounded-memory property is guarded by a **falsifiable** test asserting map sizes stay O(one segment);
  keep it that way — the original R8 test only checked index equality and let the leak ship.

Issue #105, opened 2026-07-14
feat(rdf): unify JSON-LD/YAML-LD into the native format registry


paudley:
## Context
The standard RDF serialization superset should dispatch uniformly. Today JSON-LD-star / YAML-LD-star lives in a **standalone module** (`crates/rdf/src/native_codecs/jsonld.rs`) and is NOT in the `NativeRdfFormat` enum / `classify` registry, so generic dispatch (e.g. a `convert --to jsonld`) can't reach it.

## Deliverable
Fold JSON-LD/YAML-LD into the one registry:
- Add `NativeRdfFormat` variants + `classify` rows (`application/ld+json`, `jsonld`, `.jsonld`; a YAML-LD media type + `.yamlld`).
- Add an `RdfCodec` impl routing parse/serialize through the same chokepoint as the other 7 syntaxes.

## Acceptance criteria
- `classify("jsonld")` and `classify("yamlld")` resolve to the new variants.
- `assert_round_trips` passes for JSON-LD and YAML-LD alongside Turtle/N-Triples/N-Quads/TriG/RDF-XML/TriX/HexTuples.
- Generic `RdfSerializer`/`RdfParserBackend` reach JSON-LD/YAML-LD (no side-door entry point).

## Host / reuse
`crates/rdf/src/native_codecs/{media_type,codec,jsonld,mod}.rs`.

## Dependencies
None. Blocks the CLI (I-CLI) full standard matrix.

### Standing principles (inherited)
- **Bidirectional by default** (reader + writer) unless the reverse is genuinely very hard or semantically undefined — then it ships one-way with the reason recorded, never silently.
- **Loss ledgers, always-on in code, optional output:** every conversion computes a deterministic `LossLedger` (`crates/rdf-core/src/loss.rs`) in the code path (no feature, no behavior branch — LOW/NO OPTIONALITY); lossless ops yield `is_empty()`. Only CLI surfacing is optional (`--loss-ledger[=PATH]`).
- **Caller-supplied vocabulary** (purrdf mints no IRIs; fixtures use `example.org`), **byte-deterministic** output, **RDF-1.2-complete**, **wasm32-clean + no Cargo features** for every published library crate.

### Cross-cutting acceptance (stage-workflow completion-adversary)
[...]

PR #115, opened 2026-07-15
feat(cli): the purrdf CLI — convert / query / reason

### LINKED ISSUE #106: feat(cli): the purrdf CLI — convert / query / reason

paudley:
Closes #106

## Summary
The first user-facing PurRDF binary — new native-only crate `purrdf-cli` (binary `purrdf`) — glue over the existing codec, SPARQL, and entailment engines, structured as one `Source → [entail]? → Sink` pipeline. `convert` / `query` / `reason` + a global `--loss-ledger`. Query and convert stream an mmap'd `.purrpck` pack with **zero materialization**; the serializer was generalized over `DatasetView`, unifying convert's RDF sink with CONSTRUCT/DESCRIBE query results.

## Library changes (subsume/extend)
- **`DatasetView`-generic serializer** (`purrdf-rdf`): a `PackView` (or any view, incl. a CONSTRUCT result graph) serializes to any of the 9 syntaxes with zero materialization. A canonical value-keyed row sort makes output byte-identical across backends and insertion-order independent. + a byte-parity test.
- **Public `dataset_from_view<D: DatasetView>`** (`purrdf-core`): subsumes the private pack `reconstruct` loop into one shared, generic bridge (`verify_pack` unchanged).
- **Bug fix**: TriX/HexTuples base-direction drop is now recorded in the loss ledger (`rdf12-direction-dropped`) instead of being a silent loss; new `carries_direction()` capability. Serialized bytes unchanged.

## CLI
- `convert --from --to [--base] [--entailment] [--canonical] IN OUT` — any↔any over the 9 syntaxes + the native `pack` container; extension inference or explicit; stdin/stdout `-`; mmap'd zero-copy pack path; `--canonical` = RDFC-1.0 N-Quads.
- `query --data <file|pack> [--base] [--entailment] --results-format <json|xml|csv|tsv|…9 RDF> '<SPARQL>'` — SELECT/ASK → W3C results; CONSTRUCT/DESCRIBE → the unified RDF sink; zero-copy pack query; shape/format mismatches hard-error.
- `reason --regime <simple|rdf|rdfs|owl-rl> [--base] IN OUT` — via `materialize`; owl-direct/rif/d → per-regime hard error (exit 3).
- Global `--loss-ledger[=PATH]` (absent→silent, bare→stderr, =PATH→file; always computed; never leaks to stdout). Exit codes 0/1/2/3.

## Verification
- 61 CLI integration tests drive the **built binary** (full 9×9 convert matrix via canonical-N-Quads isomorphism; all four SELECT/ASK formats; CONSTRUCT/DESCRIBE; per-regime inferences; loss-ledger tri-state).
- Library: serialize↔PackView byte-parity; RDFC-digest reconstruct isomorphism.
- `make check` + `make wasm` + `make doc` all green. `purrdf-cli` is native-only (excluded from the wasm allowlist by omission; excluded from docs like capi/python). No minted IRIs; `example.org` fixtures; deterministic output; no Cargo features.

## Summary by CodeRabbit

* **New Features**
  * Added the `purrdf` command-line tool with `convert`, `query`, and `reason` commands.
  * Supports RDF format conversion, native pack files, SPARQL queries, entailment reasoning, canonical output, stdin/stdout pipelines, and base IRI handling.
  * Added optional loss-ledger reporting to stderr or a file.
  * Added deterministic serialization and improved preservation of RDF graphs, annotations, and direction metadata where supported.

* **Documentation**
  * Added comprehensive CLI usage documentation, supported formats, options, and exit codes.

* **Bug Fixes**
  * Improved format-loss reporting for unsupported graph, statement-layer, and directional-literal features.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1282,6 +1282,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "purrdf-cli"
+version = "0.6.0"
+dependencies = [
+ "clap",
+ "memmap2",
+ "purrdf-core",
+ "purrdf-entail",
+ "purrdf-rdf",
+ "purrdf-sparql-eval",
+ "purrdf-sparql-results",
+ "tempfile",
+]
+
+[[package]]
 name = "purrdf-core"
 version = "0.6.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@
 resolver = "2"
 members = [
     "bindings/python",
+    "crates/cli",
     "crates/entail",
     "crates/gts",
     "crates/iri",
@@ -72,6 +73,12 @@ blake3 = { version = "1", features = ["rayon"] }
 boon = "0.6"
 camino = "1"
 ciborium = "0.2"
+# Argument parsing for the `purrdf` CLI (native-only binary crate `crates/cli`).
+# `derive` supplies the `Parser`/`Subcommand`/`ValueEnum` macros the CLI models
+# its command tree with. Not a library dependency, so the wasm gate (which
+# builds `--lib` on an explicit `-p` allowlist that omits `purrdf-cli`) is
+# unaffected.
+clap = { version = "4", features = ["derive"] }
 criterion = { version = "0.5", default-features = false, features = ["cargo_bench_support"] }
 csv = "1"
 datatest-stable = "0.3"
--- a/Makefile
+++ b/Makefile
@@ -111,7 +111,7 @@ test: ## Run the workspace test suite.
 	cargo test --workspace --locked
 
 doc: ## Build docs for the 16 publishable crates with rustdoc warnings denied.
-	RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --exclude purrdf-capi --exclude purrdf-python --exclude purrdf-sparql-conformance
+	RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --exclude purrdf-capi --exclude purrdf-python --exclude purrdf-sparql-conformance --exclude purrdf-cli
 
 book-samples: ## Regenerate deterministic SVG visualization samples embedded in The PurRDF Book.
 	@set -eu; \
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -1,0 +1,49 @@
+# SPDX-FileCopyrightText: 2026 Blackcat Informatics® Inc. <paudley@blackcatinformatics.ca>
+# SPDX-License-Identifier: MIT OR Apache-2.0
+
+[package]
+name = "purrdf-cli"
+description = "PurRDF command-line interface: the `purrdf` convert/query/reason pipeline over the native codecs, pack container, SPARQL engine, and entailment closures"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+homepage.workspace = true
+rust-version.workspace = true
+keywords = ["rdf", "rdf-12", "sparql", "cli", "command-line"]
+categories = ["command-line-utilities"]
+publish = false
+
+[lints]
+workspace = true
+
+[[bin]]
+name = "purrdf"
+path = "src/main.rs"
+
+[dependencies]
+# The native RDF text/statement codec surface: `classify`, `parse_dataset`,
+# `serialize_dataset_to_format`, and `NativeRdfFormat`. This is the format
+# authority the CLI's `--from`/`--to` routing resolves against.
+purrdf-rdf = { workspace = true }
+# The oxigraph-free kernel: the frozen `RdfDataset`/`DatasetView` model, the
+# pack container (`PackBuilder`/`PackView`/`verify_pack`/`dataset_from_view`),
+# the loss ledger, and the `SparqlResult` model.
+purrdf-core = { workspace = true }
+# The native SPARQL engine driving the `query` subcommand over any `DatasetView`.
+purrdf-sparql-eval = { workspace = true }
+# The W3C SPARQL-Results serializers (JSON/XML/CSV/TSV) for SELECT/ASK results.
+purrdf-sparql-results = { workspace = true }
+# The entailment regimes driving the `reason` subcommand's closure materialization.
+purrdf-entail = { workspace = true }
+# Zero-copy memory-mapped reads of on-disk pack files: a pack source is mmap'd
+# and handed to `PackView::from_bytes` with no materialization. Native-only —
+# this crate is never built for wasm32 (it is absent from the `make wasm`
+# `-p` allowlist), so mmap's absence there is a non-issue.
+memmap2 = { workspace = true }
+# Command-tree argument parsing (`Parser`/`Subcommand`/`ValueEnum`).
+clap = { workspace = true }
+
+[dev-dependencies]
+# Isolated temp files/dirs for the integration tests that drive the built binary.
+tempfile = { workspace = true }
--- a/crates/cli/README.md
+++ b/crates/cli/README.md
@@ -1,0 +1,265 @@
+<!--
+SPDX-FileCopyrightText: 2026 Blackcat Informatics® Inc. <paudley@blackcatinformatics.ca>
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+<p align="center">
+  <a href="https://github.com/Blackcat-Informatics/purrdf">
+    <img src="https://raw.githubusercontent.com/Blackcat-Informatics/purrdf/main/docs/purrdf-logo.svg" alt="PurRDF logo" width="120" height="120">
+  </a>
+</p>
+
+# `purrdf` — the PurRDF command-line interface
+
+[![License](https://img.shields.io/badge/license-MIT%20OR%20Apache--2.0-blue.svg)](https://github.com/Blackcat-Informatics/purrdf/blob/main/LICENSE-MIT)
+[![Repository](https://img.shields.io/badge/repo-Blackcat--Informatics%2Fpurrdf-181717.svg)](https://github.com/Blackcat-Informatics/purrdf)
+
+`purrdf` is the native RDF 1.2 command-line tool of the PurRDF toolkit. It is a
+thin, deterministic shell over the same engines the library exposes — the native
+text/XML/JSON codecs, the pack container, the SPARQL 1.2 evaluator, and the
+entailment closures — so anything the CLI does, it does with byte-for-byte the
+same behavior as the Rust, Python, WebAssembly, and C surfaces.
+
+Every invocation is one `Source → [transform] → Sink` pipeline, exposed as three
+subcommands:
+
+| Subcommand | Pipeline |
+|---|---|
+| [`convert`](#convert) | transcode RDF between syntaxes and the native pack container |
+| [`query`](#query) | evaluate a SPARQL query over an RDF or pack data source |
+| [`reason`](#reason) | materialize an entailment regime's closure over a source graph |
+
+A single global flag, [`--loss-ledger`](#the-loss-ledger), surfaces the
+machine-readable transcode-loss record for whichever conversion ran.
+
+> **This tool mints no vocabulary.** PurRDF is a carrier, not an ontology: every
+> IRI in your data is yours. The `example.org` IRIs below are illustrative
+> fixtures only.
+
+## Installation
+
+`purrdf` is a native-only binary (it memory-maps pack files, so it is never built
+for `wasm32`). Build it from the workspace:
+
+```sh
+cargo build --release -p purrdf-cli
+# the binary is `purrdf`:
+./target/release/purrdf --help
+```
+
+## Formats
+
+Nine native RDF syntaxes plus the native pack container are accepted anywhere a
+format is named (`--from`, `--to`, `--results-format`, or inferred from a path):
+
+| Token | Syntax | Filename extensions |
+|---|---|---|
+| `turtle` (`ttl`) | Turtle | `.ttl` |
+| `trig` | TriG | `.trig` |
+| `ntriples` (`nt`) | N-Triples | `.nt` |
+| `nquads` (`nq`) | N-Quads | `.nq` |
+| `rdfxml` (`rdf`, `xml`) | RDF/XML | `.rdf`, `.xml` |
+| `trix` | TriX | `.trix` |
+| `hextuples` (`hext`) | HexTuples | `.hext` |
+| `jsonld` (`json-ld`) | JSON-LD | `.jsonld` |
+| `yamlld` (`yaml-ld`) | YAML-LD | `.yamlld` |
+| `pack` | PurRDF pack container | `.purrpck`, `.pack` |
+
+**Format inference.** When `--from`/`--to` is omitted, the format is inferred
+from the path's extension. An explicit `--from`/`--to` always wins over the
+extension.
+
+**stdin/stdout.** A path of `-` reads from stdin or writes to stdout. Because `-`
+has no extension, it **requires** an explicit `--from` (for input) or `--to` (for
+output). `convert` defaults both `IN` and `OUT` to `-`.
+
+**The pack container.** A pack is PurRDF's native, lossless RDF 1.2 container. On
+disk it is opened **read-only and memory-mapped**, verified end-to-end
+(`verify_pack`, fail-closed), and handed to the engine zero-copy — no intermediate
+materialization for `convert` passthroughs, `query`, or serialization. A pack
+arriving on stdin is read into a buffer and verified the same way. A `pack → pack`
+`convert` is a verified byte passthrough (no decode/re-encode churn).
+
+## `convert`
+
+```text
+purrdf convert [--from <F>] [--to <F>] [--base <IRI>] [--entailment <R>] [--canonical] [IN] [OUT]
+```
+
+Transcode a source into a target syntax or the pack container.
+
+- `--from <F>` / `--to <F>` — input/output format overrides; inferred from the
+  `IN`/`OUT` extension when omitted.
+- `--base <IRI>` — base IRI for resolving relative IRIs while parsing, also
+  threaded into the serializer as its base.
+- `--entailment <R>` — materialize a regime's closure **in memory** before
+  serializing (see [`reason`](#reason) for the supported regimes and the exit-3
+  boundary; the two lanes reject identically).
+- `--canonical` — emit the RDFC-1.0 canonical N-Quads document instead of `--to`.
+  Canonical output is **always** N-Quads, so `--canonical` overrides (and lets you
+  omit) `--to`.
+
+Transforms compose in a fixed order: entail first, then canonicalize.
+
+```sh
+# Turtle → N-Triples, formats inferred from the extensions.
+purrdf convert people.ttl people.nt
+
+# JSON-LD on stdin → Turtle on stdout (explicit formats required for `-`).
+cat people.jsonld | purrdf convert --from jsonld --to turtle - -
+
+# Pack a graph into the native lossless container, then unpack it.
+purrdf convert people.ttl people.purrpck
+purrdf convert people.purrpck restored.trig
+
+# Emit RDFC-1.0 canonical N-Quads (no `--to` needed; canonical is always N-Quads).
+purrdf convert --canonical people.ttl people.nq
+
+# Materialize the RDFS closure, then canonicalize it.
+purrdf convert --entailment rdfs --canonical people.ttl closure.nq
+
+# Resolve relative IRIs against a base while converting.
+purrdf convert --base http://example.org/ data.ttl data.nt
+```
+
+## `query`
+
+```text
+purrdf query --data <file|pack> [--base <IRI>] [--entailment <R>] [--results-format <FMT>] '<SPARQL>'
+```
+
+Evaluate a SPARQL 1.2 query over a data source. The source is opened as a view (a
+pack is queried **zero-copy**); the query text and the parsed data both resolve
+relative IRIs against `--base`.
+
+- `--data <file|pack>` — the data source (format inferred from its extension).
+- `--base <IRI>` — base IRI applied to both the data parse and the query text.
+- `--entailment <R>` — reconstruct an owned dataset, materialize the regime's
+  closure in memory, and run the query over **the closure** (a pack is rebuilt for
+  this; the zero-copy path is used only without `--entailment`).
+- `--results-format <FMT>` — the result serialization (default `json`).
+
+The **result shape** selects which half of `--results-format` is legal:
+
+- **SELECT / ASK** produce solutions / a boolean → a SPARQL-results format:
+  `json`, `xml`, `csv`, `tsv`.
+- **CONSTRUCT / DESCRIBE** produce a graph → one of the nine RDF syntaxes
+  (`turtle`, `trig`, `ntriples`, `nquads`, `rdfxml`, `trix`, `hextuples`,
+  `jsonld`, `yamlld`).
+
+A shape/format mismatch (e.g. SELECT solutions with `turtle`, or a CONSTRUCT graph
+with `csv`) is a hard runtime error (exit 1). Results always go to stdout.
+
+```sh
+# SELECT → SPARQL Results JSON (the default).
+purrdf query --data people.ttl \
+  'SELECT ?name WHERE { ?p <http://example.org/name> ?name }'
+
+# ASK → CSV.
+purrdf query --data people.ttl --results-format csv \
+  'ASK { ?p <http://example.org/name> "Alice" }'
+
+# CONSTRUCT → Turtle (a graph result serialized through an RDF syntax).
+purrdf query --data people.ttl --results-format turtle \
+  'CONSTRUCT { ?p <http://example.org/label> ?name } WHERE { ?p <http://example.org/name> ?name }'
+
+# Query a pack zero-copy (mmap'd, verified, no materialization).
+purrdf query --data people.purrpck --results-format tsv \
+  'SELECT * WHERE { ?s ?p ?o } LIMIT 10'
+
+# Query the RDFS closure rather than the raw graph.
+purrdf query --data people.ttl --entailment rdfs \
+  'SELECT ?type WHERE { <http://example.org/alice> a ?type }'
+```
+
+## `reason`
+
+```text
+purrdf reason --regime <R> [--base <IRI>] [IN] [OUT]
+```
+
+Materialize an entailment regime's closure over the source graph and write it out
+(the output format is inferred from `OUT`'s extension).
+
+- `--regime <R>` — the entailment regime to close under.
+- `--base <IRI>` — base IRI for the input parse, also threaded into the serializer.
+
+**Supported (materializable) regimes:**
+
+| `--regime` | Meaning |
+|---|---|
+| `simple` | Simple entailment (a faithful copy of the source) |
+| `rdf` | RDF entailment |
+| `rdfs` | RDFS entailment |
+| `owl-rl` | OWL 2 RL entailment |
+
+**The unsupported boundary (exit code 3).** Three regimes cannot be materialized
+by the CLI because they need inputs it has no way to supply, and each is rejected
+with a distinct diagnostic:
+
+- `owl-direct` — OWL Direct (DL) needs the query's class expressions;
+- `rif` — RIF-Core needs a parsed rule set;
+- `d` — datatype (D) entailment is a spec-inherent materialization boundary.
+
+`convert --entailment` shares this boundary and rejects identically.
+
+```sh
+# Materialize the RDFS closure and write it as N-Triples.
+purrdf reason --regime rdfs people.ttl closure.nt
+
+# OWL 2 RL closure from stdin to stdout.
+cat ontology.ttl | purrdf reason --regime owl-rl - closure.ttl
+
+# The unsupported boundary: exits 3 with an explanatory message.
+purrdf reason --regime owl-direct people.ttl out.ttl
+echo $?   # 3
+```
+
+## The loss ledger
+
+`--loss-ledger` is a global flag that surfaces the machine-readable transcode-loss
+record for whichever conversion ran. The ledger is **always computed**; the flag
+only controls where (if anywhere) it is written, via three states:
+
+| Form | Effect |
+|---|---|
+| absent | silent — the ledger is not surfaced |
+| `--loss-ledger` (bare) | render the ledger's JSON to **stderr** |
+| `--loss-ledger=PATH` | write the ledger's JSON to **PATH** |
+
+The `=PATH` spelling is required (the bare form takes no value), so the flag never
+swallows a following subcommand or query string.
+
+The ledger records two kinds of loss when a target syntax cannot carry what the
+source held: the **contract** losses inherent to a `(source-codec → target-codec)`
+pair, and the **realized** counts the serializer actually dropped — RDF 1.2
+statement-layer rows (reifier bindings + annotation triples) when the target has no
+star layer, and literal base directions when the target (TriX / HexTuples) has no
+direction surface. A pack target, a `pack → pack` passthrough, and RDFC-1.0
+canonical N-Quads are lossless, so their ledgers are empty.
+
+```sh
+# Convert to a star-incapable syntax and inspect what was dropped, on stderr.
+purrdf --loss-ledger convert star-data.ttl plain.rdf
+
+# Persist the ledger to a file alongside the output.
+purrdf --loss-ledger=convert.loss.json convert star-data.ttl plain.trix
+```
+
+## Exit codes
+
+| Code | Meaning |
+|---|---|
+| `0` | success |
+| `1` | runtime failure — a parse/serialize diagnostic, a pack-integrity failure, an I/O error, or a result/shape mismatch |
+| `2` | usage error — a malformed command line (clap), or a pipeline usage error such as `-` without an explicit format |
+| `3` | unsupported entailment regime — `owl-direct` / `rif` / `d` cannot be materialized by the CLI |
+
+On any failure the error's message is printed to stderr and its category becomes
+the process exit code; nothing is swallowed.
+
+## License
+
+Licensed under either of [MIT](https://github.com/Blackcat-Informatics/purrdf/blob/main/LICENSE-MIT)
+or [Apache-2.0](https://github.com/Blackcat-Informatics/purrdf/blob/main/LICENSE-APACHE)
+at your option.
--- a/crates/cli/README.md
+++ b/crates/cli/README.md
@@ -175,13 +175,16 @@ purrdf query --data people.ttl --entailment rdfs \
 ## `reason`
 
 ```text
-purrdf reason --regime <R> [--base <IRI>] [IN] [OUT]
+purrdf reason --regime <R> [--from <F>] [--to <F>] [--base <IRI>] [IN] [OUT]
 ```
 
-Materialize an entailment regime's closure over the source graph and write it out
-(the output format is inferred from `OUT`'s extension).
+Materialize an entailment regime's closure over the source graph and write it out.
 
 - `--regime <R>` — the entailment regime to close under.
+- `--from <F>` / `--to <F>` — input/output format overrides; inferred from the
+  `IN`/`OUT` extension when omitted. `IN`/`OUT` default to `-` (stdin/stdout); a
+  path of `-` has no extension, so it **requires** the matching explicit
+  `--from`/`--to`.
 - `--base <IRI>` — base IRI for the input parse, also threaded into the serializer.
 
 **Supported (materializable) regimes:**
@@ -207,8 +210,8 @@ with a distinct diagnostic:
 # Materialize the RDFS closure and write it as N-Triples.
 purrdf reason --regime rdfs people.ttl closure.nt
 
-# OWL 2 RL closure from stdin to stdout.
-cat ontology.ttl | purrdf reason --regime owl-rl - closure.ttl
+# OWL 2 RL closure from stdin to stdout (explicit formats required for `-`).
+cat ontology.ttl | purrdf reason --regime owl-rl --from ttl --to nt - -
 
 # The unsupported boundary: exits 3 with an explanatory message.
 purrdf reason --regime owl-direct people.ttl out.ttl
--- a/crates/cli/src/cli.rs
+++ b/crates/cli/src/cli.rs
@@ -142,14 +142,20 @@ pub(crate) enum Command {
         /// The entailment regime to close under.
         #[arg(long, value_enum)]
         regime: CliRegime,
+        /// Input format override; inferred from the input extension when omitted.
+        #[arg(long, value_enum)]
+        from: Option<CliRdfFormat>,
+        /// Output format override; inferred from the output extension when omitted.
+        #[arg(long, value_enum)]
+        to: Option<CliRdfFormat>,
         /// Base IRI for resolving relative IRIs while parsing the input; also
         /// threaded into the serializer as its base.
         #[arg(long, value_name = "IRI")]
         base: Option<String>,
-        /// Input path `IN`, or `-` for stdin (which requires a recognizable format).
+        /// Input path `IN`, or `-` for stdin (which requires `--from`).
         #[arg(value_name = "IN", default_value = "-")]
         input: String,
-        /// Output path `OUT` (format inferred from its extension), or `-` for stdout.
+        /// Output path `OUT`, or `-` for stdout (which requires `--to`).
         #[arg(value_name = "OUT", default_value = "-")]
         output: String,
     },
--- a/crates/cli/src/cli.rs
+++ b/crates/cli/src/cli.rs
@@ -1,0 +1,260 @@
+// SPDX-FileCopyrightText: 2026 Blackcat Informatics® Inc. <paudley@blackcatinformatics.ca>
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! The clap command tree: the `purrdf` binary's argument model.
+//!
+//! One pipeline, three subcommands ([`Command`]), and one global flag
+//! (`--loss-ledger`). The format / regime / results-format choices are modeled as
+//! [`clap::ValueEnum`] wrappers so `--help` enumerates the legal values and clap
+//! validates them at parse time, and each wrapper carries a total conversion into
+//! its library counterpart.
+//!
+//! ## The `--loss-ledger` tri-state
+//!
+//! `--loss-ledger` is an optional-value global flag whose three states are encoded
+//! as `Option<Option<PathBuf>>`:
+//!
+//! * absent → `None` — do not surface the ledger.
+//! * `--loss-ledger` (bare) → `Some(None)` — render the ledger to **stderr**.
+//! * `--loss-ledger=PATH` → `Some(Some(PATH))` — write the ledger to **PATH**.
+//!
+//! `require_equals` forces the `=PATH` spelling so the optional value never
+//! greedily swallows a following positional (e.g. a subcommand or a query string),
+//! keeping the three states unambiguous.
+
+use std::path::PathBuf;
+
+use clap::{Parser, Subcommand, ValueEnum};
+use purrdf_entail::Regime;
+use purrdf_sparql_results::SparqlResultsFormat;
+
+use crate::format::CliFormat;
+
+/// The `purrdf` command-line interface.
+#[derive(Parser, Debug)]
+#[command(
+    name = "purrdf",
+    version,
+    about = "PurRDF: convert, query, and reason over RDF 1.2 data and native packs",
+    propagate_version = true
+)]
+pub(crate) struct Cli {
+    /// The subcommand to run.
+    #[command(subcommand)]
+    pub(crate) cmd: Command,
+
+    /// Surface the transcode loss ledger: bare writes it to stderr,
+    /// `--loss-ledger=PATH` writes it to PATH.
+    //
+    // `Option<Option<PathBuf>>` is clap's idiom for an optional-value flag (the
+    // three states are: absent / present-bare / present-with-value); it is the
+    // only place this shape appears — `Cli::ledger_target` projects it into the
+    // self-documenting `LedgerTarget` the pipeline actually threads.
+    #[allow(clippy::option_option)]
+    #[arg(
+        long,
+        global = true,
+        value_name = "PATH",
+        num_args = 0..=1,
+        require_equals = true
+    )]
+    pub(crate) loss_ledger: Option<Option<PathBuf>>,
+}
+
+/// Where (if anywhere) the loss ledger should be surfaced — the decoded form of
+/// the `--loss-ledger` tri-state flag.
+#[derive(Debug, Clone)]
+pub(crate) enum LedgerTarget {
+    /// The flag was absent: do not surface the ledger.
+    Silent,
+    /// Bare `--loss-ledger`: render the ledger to stderr.
+    Stderr,
+    /// `--loss-ledger=PATH`: write the ledger to PATH.
+    File(PathBuf),
+}
+
+impl Cli {
+    /// Decode the raw `--loss-ledger` tri-state into a [`LedgerTarget`].
+    pub(crate) fn ledger_target(&self) -> LedgerTarget {
+        match &self.loss_ledger {
+            None => LedgerTarget::Silent,
+            Some(None) => LedgerTarget::Stderr,
+            Some(Some(path)) => LedgerTarget::File(path.clone()),
+        }
+    }
+}
+
+/// The three pipeline subcommands.
+#[derive(Subcommand, Debug)]
+pub(crate) enum Command {
+    /// Convert RDF between syntaxes, and to/from the native pack container.
+    Convert {
+        /// Input format override; inferred from the input extension when omitted.
+        #[arg(long, value_enum)]
+        from: Option<CliRdfFormat>,
+        /// Output format override; inferred from the output extension when omitted.
+        #[arg(long, value_enum)]
+        to: Option<CliRdfFormat>,
+        /// Input path `IN`, or `-` for stdin (which requires `--from`).
+        #[arg(value_name = "IN", default_value = "-")]
+        input: String,
+        /// Output path `OUT`, or `-` for stdout (which requires `--to`).
+        #[arg(value_name = "OUT", default_value = "-")]
+        output: String,
+    },
+    /// Evaluate a SPARQL query over an RDF or pack data source.
+    Query {
+        /// Data-source path (format inferred from its extension). A pack file is
+        /// queried zero-copy.
+        #[arg(long)]
+        data: String,
+        /// SPARQL-results serialization for SELECT/ASK results.
+        #[arg(long, value_enum, default_value_t = ResultsFormat::Json)]
+        results_format: ResultsFormat,
+        /// The SPARQL query text.
+        query: String,
+    },
+    /// Materialize an entailment regime's closure over a source graph.
+    Reason {
+        /// The entailment regime to close under.
+        #[arg(long, value_enum)]
+        regime: CliRegime,
+        /// Input path `IN`, or `-` for stdin (which requires a recognizable format).
+        #[arg(value_name = "IN", default_value = "-")]
+        input: String,
+        /// Output path `OUT` (format inferred from its extension), or `-` for stdout.
+        #[arg(value_name = "OUT", default_value = "-")]
+        output: String,
+    },
+}
+
+/// The input/output format choices `--from`/`--to` accept: the nine native RDF
+/// syntaxes plus the native `pack` container.
+///
+/// Each variant's canonical value is the one `--help` lists; the short
+/// extension/id spellings the native codec [`classify`](purrdf_rdf::classify)
+/// accepts (e.g. `ttl`, `nt`, `nq`) are registered as hidden aliases so the same
+/// name works on the command line and in a filename.
+#[derive(ValueEnum, Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum CliRdfFormat {
+    /// Turtle.
+    #[value(name = "turtle", alias = "ttl")]
+    Turtle,
+    /// TriG.
+    #[value(name = "trig")]
+    Trig,
+    /// N-Triples.
+    #[value(name = "ntriples", alias = "nt", alias = "n-triples")]
+    Ntriples,
+    /// N-Quads.
+    #[value(name = "nquads", alias = "nq", alias = "n-quads")]
+    Nquads,
+    /// RDF/XML.
+    #[value(name = "rdfxml", alias = "rdf", alias = "xml")]
+    Rdfxml,
+    /// TriX.
+    #[value(name = "trix")]
+    Trix,
+    /// HexTuples.
+    #[value(name = "hextuples", alias = "hext")]
+    Hextuples,
+    /// JSON-LD.
+    #[value(name = "jsonld", alias = "json-ld")]
+    Jsonld,
+    /// YAML-LD.
+    #[value(name = "yamlld", alias = "yaml-ld")]
+    Yamlld,
+    /// The native PurRDF pack container.
+    #[value(name = "pack")]
+    Pack,
+}
+
+impl CliRdfFormat {
+    /// Resolve this explicit choice to the pipeline's [`CliFormat`].
+    pub(crate) fn to_cli_format(self) -> CliFormat {
+        use purrdf_rdf::NativeRdfFormat as N;
+        match self {
+            Self::Turtle => CliFormat::Rdf(N::Turtle),
+            Self::Trig => CliFormat::Rdf(N::TriG),
+            Self::Ntriples => CliFormat::Rdf(N::NTriples),
+            Self::Nquads => CliFormat::Rdf(N::NQuads),
+            Self::Rdfxml => CliFormat::Rdf(N::RdfXml),
+            Self::Trix => CliFormat::Rdf(N::TriX),
+            Self::Hextuples => CliFormat::Rdf(N::HexTuples),
+            Self::Jsonld => CliFormat::Rdf(N::JsonLd),
+            Self::Yamlld => CliFormat::Rdf(N::YamlLd),
+            Self::Pack => CliFormat::Pack,
+        }
+    }
+}
+
+/// The SPARQL-results serialization choices `--results-format` accepts.
+#[derive(ValueEnum, Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum ResultsFormat {
+    /// SPARQL Results JSON.
+    #[value(name = "json")]
+    Json,
+    /// SPARQL Results XML.
+    #[value(name = "xml")]
+    Xml,
+    /// SPARQL Results CSV.
+    #[value(name = "csv")]
+    Csv,
+    /// SPARQL Results TSV.
+    #[value(name = "tsv")]
+    Tsv,
+}
+
+impl ResultsFormat {
+    /// The library [`SparqlResultsFormat`] this choice maps to.
+    pub(crate) fn to_native(self) -> SparqlResultsFormat {
+        match self {
+            Self::Json => SparqlResultsFormat::Json,
+            Self::Xml => SparqlResultsFormat::Xml,
+            Self::Csv => SparqlResultsFormat::Csv,
+            Self::Tsv => SparqlResultsFormat::Tsv,
+        }
+    }
+}
+
+/// The entailment-regime choices `--regime` accepts.
+#[derive(ValueEnum, Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum CliRegime {
+    /// Simple entailment (a faithful copy of the source).
+    #[value(name = "simple")]
+    Simple,
+    /// RDF entailment.
+    #[value(name = "rdf")]
+    Rdf,
+    /// RDFS entailment.
+    #[value(name = "rdfs")]
+    Rdfs,
+    /// OWL 2 RL entailment.
+    #[value(name = "owl-rl")]
+    OwlRl,
+    /// OWL Direct (DL) entailment — not materializable without query class
+    /// expressions.
+    #[value(name = "owl-direct")]
+    OwlDirect,
+    /// RIF-Core entailment — not materializable without a rule set.
+    #[value(name = "rif")]
+    Rif,
+    /// Datatype (D) entailment — a spec-inherent boundary for materialization.
+    #[value(name = "d")]
+    D,
+}
+
+impl CliRegime {
+    /// The library [`Regime`] this choice maps to.
+    pub(crate) fn to_native(self) -> Regime {
+        match self {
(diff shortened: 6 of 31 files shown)


paudley:
# Implementation Plan for Issue #106 — the `purrdf` CLI (convert / query / reason)

## Context

PurRDF has no user-facing command-line tool. The only binaries today are three
maintainer-only golden/vector freezers in `crates/rdf/Cargo.toml`. Every engine a
real CLI needs already exists and is production-tested: the native codec layer
(`parse_dataset` / `serialize_dataset*`), the `NativeSparqlEngine`, the
`purrdf-entail::materialize` reasoner, the SPARQL-results serializer, and the
mmap-able `PackView`. Issue #106 asks for the first user-facing binary that glues
these together into three subcommands — `convert`, `query`, `reason` — plus a
global `--loss-ledger` flag that surfaces the always-computed codec loss ledger.

Prerequisites #104 (unified loss-ledger surface) and #105 (JSON-LD/YAML-LD in the
native registry) are both CLOSED, so #106 is unblocked.

The design goal (from `.goals`): a native-only crate, glue over existing engines, but
with **no compromise**. Exploration surfaced a richer structure the three subcommands
are a shadow of — a single dataflow:

```
Source ─ {text file │ .purrpck pack (mmap, zero-copy) │ stdin '-'}
  │        (yields a concrete DatasetView: Arc<RdfDataset> or PackView)
  ├─ [optional Transform] entail(regime)   ← in-memory closure via materialize
  └─ Terminal:
       • RDF Sink   → {9 text formats │ .purrpck pack │ RDFC-1.0 canonical} + always-on LossLedger
       • SPARQL     → { SELECT/ASK → SparqlResultsFormat {json│xml│csv│tsv}
                        CONSTRUCT/DESCRIBE → the SAME RDF Sink (+ LossLedger) }
```

Three findings drove this:

1. **Query over an mmap'd pack is already zero-materialization** via the engine's
   generic `query_prepared_view<D: DatasetView>` (`crates/sparql-eval/src/engine.rs:200`),
   proven by `crates/sparql-eval/tests/pack_query_e2e.rs`.
2. **The serializer can be `DatasetView`-generic with no trait change** — the
   `DatasetView` trait (`crates/rdf-core/src/dataset_view.rs:148`) already exposes every
   accessor the serializer uses (`quads`, `resolve`, `reifier_quads`, `annotation_quads`,
   `term_id_by_value`, `term_count`, `len_hint`). So `PackView` serializes to text with
   zero materialization (Task 1), and a CONSTRUCT/DESCRIBE result graph reuses the exact
   same RDF sink as `convert`.
3. **Entailment is a query parameter, not only a subcommand.** SPARQL is parameterized
   by `sparql:entailmentRegime`; `reason` is the serialize-terminal projection of the
   same pipeline. `--entailment <regime>` on `query`/`convert` materializes the closure
   in memory (reusing `materialize`) and needs zero new engine code.

Outcome: all three subcommands accept `{text, mmap'd pack, stdin}` uniformly; `query`
handles every SPARQL result shape (SELECT/ASK/CONSTRUCT/DESCRIBE); the loss ledger is a
universal RDF-sink invariant; and both `convert` and `query`-over-a-pack stream with no
full in-memory load.

- Adversarial gap analysis; build executable production-surface checks, not prose.
- **No deferrals / follow-ups / future-work.** Everything is done on this branch.
- **No GitHub issue/PR numbers in code, ontology, or docs** (`check-issue-refs.py`).
- When you see a bug, you own it and fix it now.

- GREENFIELD-FIRST, SOTA, NO COMPROMISES.
- RUST-FIRST, PYTHON-SURFACE — all core work in Rust (native Rust binary here).
- SUBSUME, EXTEND, ENHANCE — generalize existing engines rather than bolt on.
- LOW/NO OPTIONALITY, HARD FAILS — no Cargo features on member crates; fail fast on bad
  input / unsupported regimes; no swallowed errors; no modal behavior branches.
- RDF 1.2 ONLY — RDF-1.2 / RDF-star first-class; the star layer round-trips.
- MAXIMAL UTILITY / PERFORMANCE / PORTABILITY — the pack path stays zero-copy; every
  *library* crate stays wasm32-clean (the CLI is native-only, excluded by omission).

Plus repo hard constraints (AGENTS.md / CLAUDE.md): `make check` warning-free
(pedantic/nursery clippy); no Cargo features; dependency versions only in root
`[workspace.dependencies]`; deterministic serializer output; PurRDF mints **no**
vocabulary IRIs (fixtures use `example.org`); do not break `make wasm`.

## Completeness Contract

Every requirement and acceptance criterion from issue #106 maps to a numbered task.
Nothing is deferred, stubbed, or left for a future PR.

| New native binary crate `purrdf-cli` (binary `purrdf`), must not break `make wasm --lib` | 3, 8 |
| `convert` any↔any over the 9-format matrix (ttl/nt/nq/trig/rdfxml/trix/hext/jsonld/yamlld) | 4 |
| convert: format inferred from extension OR set explicitly via `--from`/`--to` | 3, 4 |
| convert: large **pack** input via mmap'd `PackView`, **zero materialization** (pack→text) | 1, 4 |
| convert: large **text** input parses correctly into IR (zero-copy holds only for pack input — AC scope) | 4 |
| `query --data <file\|pack> '<SPARQL>'` over a file AND an mmap'd pack | 5 |
| query: SELECT/ASK results in SRJ/CSV/TSV/XML | 5 |
| query: CONSTRUCT/DESCRIBE results serialize via the RDF sink (core SPARQL form, not excluded) | 1, 5 |
| `reason --regime <rdfs\|owl-rl\|...> IN OUT` via `materialize`, for each **supported** regime | 6 |
| unsupported regimes (owl-direct/rif/d) → clean HARD ERROR (spec-inherent boundary), distinct exit code | 6 |
| Global `--loss-ledger[=PATH]`: `render_json()`; empty when lossless; non-empty when lossy; absent ⇒ silent | 7 |
| Loss ledger always computed in code (no feature/branch); only CLI surfacing is optional; universal RDF-sink invariant | 4, 5, 6, 7 |
| Demonstrated conversion matrix (production binary) | 4 |
| SPARQL query over BOTH a plain file and an mmap'd pack (production binary) | 5 |
| `reason` run for each supported regime (production binary) | 6 |
| `--loss-ledger` non-empty on a lossy op, empty on a lossless op (production binary) | 7 |
| Bidirectional codecs pass round-trip (lossless, RDFC-digest isomorphism) / write→read→write + ledger (lossy) | 4 |
| No minted IRIs; `example.org`-only fixtures; byte-deterministic output | 4,5,6,7,8 |
| `make check` + `make wasm` + `make doc` stay green; no issue refs; no Cargo features | 8 |

### Adopted enhancements (folded into the tasks above; not separate deliverables)
- Unified RDF sink for convert / reason / query-CONSTRUCT (Tasks 1,4,5,7).
- `--entailment <regime>` in-memory pre-transform on `query` and `convert` — reuses
  `materialize` (Tasks 4,5,6).
- `--base <IRI>` ingress for relative-IRI inputs via `parse_dataset_with` (Tasks 3,4,5,6).
- `--canonical` RDFC-1.0 N-Quads egress mode + RDFC-digest isomorphism asserts in tests
  (Tasks 4,8).
- clap `ValueEnum` for `--from/--to/--results-format/--regime` (validation + `--help`
  enumeration + completion; replaces hand-rolled matches) (Task 3).
- Unconditional fail-closed pack ingest (`verify_pack`/`from_bytes` errors → hard exit)
  on EVERY command, with a corrupt-pack test per command; mmap read-only caveat documented
  (Tasks 3,4,5,6).
- Exit-code taxonomy: usage=2, runtime=1, unsupported-regime-boundary=3 (Tasks 3,6).

- **`DatasetView`-generic `materialize` (query a pack for reason with zero
  materialization).** DECLINED — the RDFS/OWL-RL chase in `crates/entail` is inherently
  whole-dataset and materializes its closure regardless, and its native semi-naive
  evaluator is written against concrete `RdfDataset` term ids; genericizing it is a
  large, high-risk refactor that buys nothing (the closure is allocated anyway).
  Reconstruct-then-materialize (Task 2) reaches the identical result at S cost/low risk
  and reuses proven code.
- **Register `.purrpck` as a 10th `NativeRdfFormat`.** DECLINED — the pack is a binary
  container (magic/dictionary/side-tables), not an RDF *text* codec; forcing it into
  `classify`/`FORMATS` pollutes a registry whose invariant is "one row per text syntax."
  The correct seam is a CLI-level `CliFormat = Rdf(NativeRdfFormat) | Pack`. Pack
  read/write is first-class in the CLI regardless.
- **GTS bundle (`.gts`) as a fourth Source kind.** DECLINED — a signed GTS bundle carries
  its own certification/verification semantics (digests, signatures) distinct from the
  issue's plain-RDF I/O contract; treating it as "just another input" would elide those
  checks. It is outside #106's stated I/O surface (the 9 text formats + the pack
  container); the pipeline structure admits it later as a first-class verb without
  rework, but it is not this issue's work.
- **Multiple input files → union dataset (`convert a.ttl b.nq -o out`).** DECLINED —
  merging distinct documents requires deliberate blank-node scoping/relabelling decisions
  (label collisions across inputs) that the single-document `IN → OUT` contract does not
  raise; it is a distinct merge operation, not part of #106's convert contract.
- **Streaming text→text line codec bypassing the IR.** REJECTED — only sound for
  base-quad line formats; breaks the RDF-1.2 statement-layer two-pass fold and the
  one-engine/one-behavior/byte-deterministic invariant (`.goals`).
- **gzip/zstd auto-decompression of inputs.** REJECTED — an implicit modal behavior
  branch on input bytes; conflicts with LOW/NO OPTIONALITY and no-magic-modes.
- **`validate`/`shex`/`describe` subcommands.** REJECTED for this issue — #106 defines
  exactly convert/query/reason + loss ledger; the pipeline makes them drop-in terminals
  later, but adding them now expands the contract beyond the issue.

---

1. Resolve the top-level repo and names:
   ```bash
   BRANCH="paudley/106-purrdf-cli"
   WORKTREE="$REPO_ROOT/.worktrees/106-purrdf-cli"
   ```
   ```bash
   ```
   ```bash
   cd "$WORKTREE"
   git branch --show-current   # MUST show paudley/106-purrdf-cli, NOT main
   ```
4. Warm the build environment: `cargo fetch` + a baseline `cargo build --workspace`.
   ```bash
   gh issue comment 106 --body-file <this-plan-file-path>
   ```

**⚠️ DO NOT proceed to Task 1 until you are on the feature branch inside the worktree.**

---

## Task 1: `DatasetView`-generic serializer (library — `purrdf-rdf`)

**The transformational core.** Generalize the native egress path so a `PackView` (or any
`DatasetView`, e.g. a CONSTRUCT result graph) serializes straight to RDF text with no
materialization.

Files: `crates/rdf/src/native_codecs/serialize.rs` (primary); check callers in
`crates/rdf/src/native_codecs/jsonld*` and `dataset_io.rs`.

Changes:
- Parametrize `serialize_dataset`, `serialize_dataset_base_only`,
  `serialize_dataset_inner`, `serialize_into`, `serialize_dataset_to_format`, and
  `build_ser_graph` over `D: DatasetView` (was `&RdfDataset`), using `D::Id` in place of
  `TermId` throughout `SerGraphInterner`, `term_value`, `iri_of`,
  `intern_triple_components`. `TermRef<'_, D::Id>` is already generic.
- Switch the RDF-1.2 statement layer from concrete `dataset.reifiers_with_graph()` /
  `annotations_with_graph()` to the trait methods `DatasetView::reifier_quads()` /
  `annotation_quads()` (both yield `QuadIds<D::Id>` carrying the graph slot). In
  `serialize_dataset_to_format`, compute the dropped count from
  `view.reifier_quads().count() + view.annotation_quads().count()`.
- Replace `dataset.quad_count()` hint with `view.len_hint().unwrap_or(0)`; keep
  `view.term_count()`.
- Backward compatible: `&RdfDataset` still satisfies `DatasetView`, so all current call
  sites compile unchanged (source-compatible generalization, ENHANCE not break).

Executable acceptance (new `crates/rdf/tests/serialize_pack_parity.rs`):
- Build the rich RDF-1.2 fixture (mirror `pack_query_e2e.rs`: default + ≥2 named graphs,
  every literal shape, a scoped blank node, a quoted-triple term, a reifier + a
  default-scoped AND a graph-scoped annotation; `example.org` only).
- For **every** `NativeRdfFormat`, assert `serialize_dataset_to_format(&rdf_dataset, fmt)`
  == `serialize_dataset_to_format(&PackView::from_bytes(&pack_bytes)?, fmt)`
  **byte-identical**, AND equal `statement_rows_dropped`. Non-vacuous (output len > 0).
  If an ordering divergence surfaces, fix it here so both paths are deterministic and
  identical (you own it).
- **Zero-materialization is a structural guarantee of this generic path, not a runtime
  claim**: the pack→text egress serializes a `&PackView` directly through
  `serialize_dataset_to_format::<PackView>` and NEVER calls `dataset_from_view`
  (Task 2's reconstructor). Pin this structurally in the CLI convert path (Task 4): the
  pack→text arm must be statically unable to reach `dataset_from_view` (assert by
  construction / a targeted test that the reconstructor is not on that arm), so "no full
  in-memory load" for pack input is guaranteed by the type-directed dispatch, not merely
  by output byte-identity.

### After Task 1: commit (github-commit-agent) → push (github-push-agent) → wiring-verifier → post update to issue #106.

---

## Task 2: Public `DatasetView`-generic reconstructor (library — `purrdf-core`)

Subsume the private `certify::reconstruct` (`crates/rdf-core/src/ir/pack/certify.rs:178`)
into a public, generic bridge so a pack can feed a transform that needs a concrete
`RdfDataset` (the reasoner).

Files: `crates/rdf-core/src/ir/pack/certify.rs`, `crates/rdf-core/src/lib.rs` (re-export),
and the `purrdf-rdf` re-export surface.

Changes:
- Add `pub fn dataset_from_view<D: DatasetView>(view: &D) -> Result<Arc<RdfDataset>, RdfDiagnostic>`
  generalizing the reconstruct loop (`quads` → re-intern → `push_quad`;
  `reifier_quads`/`annotation_quads` → the RDF-1.2 side tables) over any `DatasetView`,
  then `freeze()`.
- Refactor `certify::reconstruct` / `verify_pack` to call the new public fn — **no
  duplicated loop** (SUBSUME).
- Re-export from `purrdf_core` and `purrdf_rdf`.

Executable acceptance (extend pack tests):
- `dataset → PackBuilder::build_bytes → PackView::from_bytes → dataset_from_view` yields a
  dataset **isomorphic** to the original by RDFC-1.0 digest, and its
  `serialize_dataset_to_format` output equals the original's for a star-capable format.

### After Task 2: commit → push → wiring-verifier → post update to issue #106.

---

## Task 3: `purrdf-cli` crate scaffold, workspace wiring, pipeline plumbing

Create the native binary crate and the shared Source→Transform→Sink plumbing.

Files:
- `crates/cli/Cargo.toml` — package `purrdf-cli`; `[[bin]] name = "purrdf"`,
  `path = "src/main.rs"`. Deps (all `{ workspace = true }`): `purrdf-rdf`, `purrdf-core`,
  `purrdf-sparql-eval`, `purrdf-sparql-results`, `purrdf-entail`, `clap` (derive),
  `memmap2`. No `[features]` table.
- Root `Cargo.toml` — add `crates/cli` to `[workspace.members]`; add
  `clap = { version = "4", features = ["derive"] }` to `[workspace.dependencies]` (match
  `check-versions.py`'s expected form; `memmap2` is already a workspace dep).
- `Makefile` — add `--exclude purrdf-cli` to the `make doc` recipe (match the app-crate
  pattern for `purrdf-capi`/`purrdf-python`/`purrdf-sparql-conformance`). `make wasm` needs
  no change (explicit `-p` allowlist excludes by omission).
- `crates/cli/src/main.rs` + `cli.rs` — clap derive. Top-level `Cli { global
  --loss-ledger[=PATH]; subcommand }`. Subcommands:
  - `convert --from <F> --to <F> [--base IRI] [--entailment <R>] [--canonical] IN OUT`
  - `query --data <file|pack> [--base IRI] [--entailment <R>] --results-format <RF> '<SPARQL>'`
  - `reason --regime <R> [--base IRI] IN OUT`
  Use `clap::ValueEnum` wrappers for `<F>` (9 formats + `pack`), `<RF>`
  (json/xml/csv/tsv + the 9 RDF formats for CONSTRUCT/DESCRIBE), and `<R>`
  (simple/rdf/rdfs/owl-rl + the unsupported ones so `--help` lists them and the boundary
  is explicit).
- Error/exit funnel: a single `enum CliError` → exit codes **2** (usage/parse of args),
  **3** (unsupported-regime boundary), **1** (all other runtime errors). Every error
  writes a diagnostic to stderr; nothing swallowed.
- `crates/cli/src/format.rs` — `CliFormat` resolution: explicit `--from`/`--to` wins;
  else infer from the path extension (extract `path.extension()`, prepend `.`, feed
  `classify`; `.purrpck`/`.pack` → `Pack`). Unresolvable ⇒ hard error (no default).
- `crates/cli/src/source.rs` — open an input as a concrete `DatasetView`. `DatasetView`
  is not object-safe (RPITIT), so dispatch by kind and call a **generic** op:
  `TextDataset(Arc<RdfDataset>)` (parse text file / stdin `-`, honoring `--base` via
  `parse_dataset_with`) vs `Pack(Mmap)` (`unsafe { Mmap::map(&File::open(p)?) }` held in a
  local; `PackView::from_bytes(&mmap)` at the call site). **Pack ingest is unconditionally
  fail-closed**: a `from_bytes`/magic error → hard exit (code 1). Document the mmap
  read-only / no-concurrent-mutation caveat in the module docs. Expose
  `run_over_input<R>(spec, f)` where `f: impl FnOnce(&D) for D: DatasetView`, monomorphized
  for both arms.
- `crates/cli/src/sink.rs` — the unified RDF sink: given any `&D: DatasetView`, a target
  `CliFormat`/`--canonical`, and `--base`, serialize (via Task 1) and compute the always-on
  `LossLedger`; write to OUT (file or stdout `-`). Pack output → `PackBuilder::build_bytes`
  from a concrete dataset (parse/reconstruct as needed). Canonical → RDFC-1.0 N-Quads.

Executable acceptance:
- `cargo build -p purrdf-cli` succeeds; `purrdf --help` and each subcommand `--help` print
  usage listing the global `--loss-ledger`, the ValueEnum choices, `--base`,
  `--entailment`, `--canonical` (assert in a test driving the binary).
- `make wasm` remains green (crate excluded by omission).

### After Task 3: commit → push → wiring-verifier → post update to issue #106.

---

## Task 4: `convert` subcommand + conversion matrix

`crates/cli/src/convert.rs`.

Behavior (Source → [entail?] → RDF Sink):
- Resolve `CliFormat` for IN and OUT (Task 3).
- Optional `--entailment <R>`: materialize the closure in memory first (reuses Task 6's
  regime mapping + `materialize`, reconstructing a pack input via Task 2), then serialize
  the closure.
- **text/pack → text**: `serialize_dataset_to_format::<D>(view, fmt, base)` from the view
  — zero-copy when the view is a `PackView`.
- **→ pack**: from a concrete dataset (parse; or `dataset_from_view` for a pack input) →
  `PackBuilder::build_bytes` → write OUT. **pack → pack**: `verify_pack` (unconditional
  fail-closed), then pass the bytes through.
- **`--canonical`**: RDFC-1.0 N-Quads egress (interchange/diff form).
- Compute the loss ledger (`pair_loss_ledger(src_codec, dst_codec)` + realized
  `statement_rows_dropped`); hand to `--loss-ledger` (Task 7). Always computed.
- Write OUT (file or stdout `-`).

Executable acceptance (`crates/cli/tests/convert_cli.rs`, driving the built binary):
- **Full any↔any matrix (not representative pairs).** From one `example.org` seed
  dataset (with named graphs + an RDF-1.2 reifier/annotation so star loss is exercised),
  materialize a canonical form in each of the 9 formats, then loop over **every ordered
  (src, dst) pair of the 9 formats** and run `purrdf convert --from <src> --to <dst>` on
  the built binary, asserting:
  - dst is star-capable & dataset-preserving ⇒ **RDFC-digest isomorphism** of the output
    with the seed (semantics preserved) AND write→read→write byte-stability;
  - dst is single-graph or star-incapable ⇒ the guaranteed projection holds (base quads
    survive; named-graph/statement-layer drop equals the declared contract) AND the loss
    ledger records exactly those losses (per Task 7).
  This loop is the executable check for the "any↔any over the 9-format matrix" criterion —
  all 9 formats appear as BOTH source and target.
- **Explicit `--from`/`--to` selection leg** (distinct from extension inference): (a)
  convert an **extensionless** file and a **stdin `-`** input using explicit `--from`/
  `--to`, asserting correct format selection; (b) convert a file whose extension is
  **deliberately misleading** (e.g. `data.txt` holding Turtle, or `.nt` holding N-Quads),
  asserting `--from`/`--to` overrides the extension. Also assert a truly unresolvable
  format (no extension, no `--from`) exits non-zero.
- text→pack→text: `in.ttl → mid.purrpck → out.nt` equals a direct `in.ttl → out.nt`.
- **pack→text via mmap** (zero-copy, Task 1): `big.purrpck → out.trig` byte-identical to
  the source dataset serialized directly.
- **Large-text ingress correctness**: a large `.nt`/`.ttl` input converts *correctly*
  (the "no full in-memory load" AC holds for *pack* inputs; large text still parses to IR
  — this test pins correctness, not zero-copy, so the AC is not over-read).
- `--canonical` produces stable RDFC-1.0 N-Quads. Deterministic: every conversion run
  twice is byte-identical. `example.org` only.
- **`convert --entailment rdfs`** (the entail→RDF-sink terminal): run once on a text
  input and once on a `.purrpck` **pack** input (exercising reconstruct-then-entail),
  asserting the RDFS-inferred triple appears in the converted output that is absent from a
  plain `--entailment`-less convert of the same seed.

### After Task 4: commit → push → wiring-verifier → post update to issue #106.

---

## Task 5: `query` subcommand (file + mmap'd pack; all result shapes)

`crates/cli/src/query.rs`.

Behavior (Source → [entail?] → SPARQL → shape-dispatched sink):
- Optional `--entailment <R>`: materialize the closure in memory (Task 6 mapping +
  `materialize`; reconstruct a pack input via Task 2), then query the closure.
- `NativeSparqlEngine::new().prepare_query(q, base)` then
  `engine.query_prepared_view(&view, &prepared, &[])` — text parses to `Arc<RdfDataset>`
  (honoring `--base`); pack is `PackView::from_bytes(&mmap)` (**zero-copy**, unconditional
  fail-closed).
- Dispatch on `SparqlResult`:
  - **Solutions (SELECT) / Boolean (ASK)** → `purrdf_sparql_results::serialize(&result,
    fmt, &ResultProvenance::default())` for `--results-format json|xml|csv|tsv`.
  - **Graph (CONSTRUCT/DESCRIBE)** → the unified RDF sink (Task 3 `sink.rs`) using an RDF
    `--results-format` (one of the 9), with the always-on loss ledger.
  - HARD FAIL (clear diagnostic) on a shape/format mismatch (CSV of a graph, Turtle of a
    boolean).

Executable acceptance (`crates/cli/tests/query_cli.rs`, driving the built binary):
- The SAME `SELECT` over (a) a plain text file and (b) an mmap'd `.purrpck` pack built
  from identical data ⇒ **byte-identical** results; non-vacuous (≥1 row).
- **All four SELECT/ASK result formats driven on the binary**: a SELECT emitted with
  `--results-format` = each of `json`, `xml`, `csv`, `tsv` (assert non-vacuous +
  deterministic output for every one — TSV and XML are explicit issue criteria, not just
  json/csv), and an ASK emitted as `json` (boolean).
- **CONSTRUCT → Turtle AND DESCRIBE → Turtle** (both graph-shaped results): assert the
  constructed/described triples appear in the RDF-sink output, and, for a star-incapable
  target carrying a reifier, that the ledger records the drop.
- **Shape/format mismatch is a hard error**: `--results-format csv` on a CONSTRUCT (graph)
  and `--results-format turtle` on an ASK (boolean) each exit non-zero with a diagnostic.
- `--entailment rdfs`: a query that returns an inferred triple only under RDFS closure.
- Corrupt-pack input ⇒ non-zero exit. `example.org` fixtures; deterministic.

### After Task 5: commit → push → wiring-verifier → post update to issue #106.

---

## Task 6: `reason` subcommand (all supported regimes; hard-fail the rest)

`crates/cli/src/reason.rs`. Reason is the entail-then-serialize terminal of the
shared pipeline.

Behavior:
- Map `--regime` (clap `ValueEnum`): `simple|rdf|rdfs|owl-rl` (and full W3C entailment
  IRIs via `Regime::from_iri`). `owl-direct|rif|d` → **hard error, exit code 3**, message
  stating they require the query's class expressions / a rule set / are a
  datatype-entailment boundary (mirrors `EntailError::Unsupported`). Spec-inherent
  boundary surfaced as a distinct machine-detectable exit, never a silent skip.
- Input → `&RdfDataset`: text parses (honoring `--base`); a pack is reconstructed via
  `dataset_from_view` (Task 2). Then `materialize(&ds, regime)` → closed `Arc<RdfDataset>`
  → the unified RDF sink (Task 3) to OUT (text or pack), with the always-on loss ledger.

Executable acceptance (`crates/cli/tests/reason_cli.rs`, driving the built binary):
- For EACH supported regime (Simple/RDF/RDFS/OWL-RL) over an `example.org` fixture with a
  known entailment (e.g. `rdfs:subClassOf` chain → inferred `rdf:type`; an
  `owl:TransitiveProperty` closure for OWL-RL), assert OUT contains the inferred
  triple(s); Simple asserted identity.
- `reason --regime owl-direct` (and `rif`, `d`) exits with code **3** and the specific
  diagnostic.
- At least one regime over a `.purrpck` pack input (exercises Task 2). Deterministic.

### After Task 6: commit → push → wiring-verifier → post update to issue #106.

---

## Task 7: Global `--loss-ledger[=PATH]` — universal RDF-sink invariant

`crates/cli/src/sink.rs` / a `ledger.rs` helper; consumed by convert (Task 4),
reason (Task 6), and query-CONSTRUCT (Task 5).

Behavior:
- The ledger is **always computed** by every RDF-sink path (convert, reason,
  query-CONSTRUCT). The flag only controls surfacing:
  - absent ⇒ emit nothing (stdout carries only the conversion output).
  - `--loss-ledger` (no PATH) ⇒ write `ledger.render_json()` to **stderr**.
  - `--loss-ledger=PATH` ⇒ write `render_json()` to PATH.
- Lossless ⇒ empty ledger (`{"schema_version":1,"losses":[]}`); lossy ⇒ enumerated
  losses. Output-only; no behavior branch on the flag.

Executable acceptance (`crates/cli/tests/loss_ledger_cli.rs`, driving the built binary):
- Lossy conversion (`nq → rdfxml` with a reifier + annotation, or `trig → turtle`
  dropping a named graph) with `--loss-ledger` ⇒ **non-empty** JSON; assert the specific
  loss codes appear.
- Lossless conversion (`nq → trig`, or star-free default-graph `ttl → nt`) with
  `--loss-ledger` ⇒ **empty** losses array.
- Omitting `--loss-ledger` ⇒ no ledger bytes anywhere. `--loss-ledger=PATH` writes the
  identical JSON to the file.
- A CONSTRUCT-to-a-star-incapable-format query with `--loss-ledger` ⇒ non-empty (proves
  the universal-sink invariant). Byte-deterministic.

### After Task 7: commit → push → wiring-verifier → post update to issue #106.

---

## Task 8: Docs, gates, and final hygiene

Files: `crates/cli/README.md` and/or `crates/cli/src/main.rs` module docs (the three
subcommands, `--from/--to`, `--results-format`, `--regime`, `--entailment`, `--base`,
`--canonical`, `--loss-ledger`, the pack/mmap path, stdin/stdout `-`). No issue/PR numbers.

Verification (all must pass):
- `make check` — fmt, pedantic/nursery clippy warning-free, `cargo check/test
  --workspace`, all Python guards (`check-no-features`, `check-licenses` incl. clap's
  license, `check-issue-refs`, `check-versions`, `check-corpus-frozen`,
  `check-generated`), and the embedded `make wasm`.
- `make wasm` — the wasm32 `-p` allowlist still builds clean (CLI excluded).
- `make doc` — `RUSTDOCFLAGS=-D warnings` clean with `--exclude purrdf-cli`.
- Grep the new crate for minted vocabulary IRIs (only `example.org` fixtures /
  caller-supplied IRIs); no fabricated defaults.
- Confirm `purrdf --help` documents all three subcommands + the global flag.

### After Task 8: commit → push → wiring-verifier → post update to issue #106.

---

1. Verify all changes committed and pushed; run the Mechanical Deferral Scan over the
   branch diff (no `TODO`/`unimplemented!`/"future"/stub) and resolve any hit.
2. Create the PR:
   ```bash
   gh pr create --title "feat(cli): the purrdf CLI — convert / query / reason" --body "Closes #106

   ## Summary
   First user-facing PurRDF binary (crate purrdf-cli, binary \`purrdf\`): native-only glue
   over the existing codec, SPARQL, and entailment engines, structured as one
   Source → [entail]? → Sink pipeline. convert / query / reason + a global --loss-ledger.
   Query and convert stream an mmap'd .purrpck pack with zero materialization; the
   serializer was generalized over DatasetView, unifying convert's and CONSTRUCT's RDF sink.

   ## Changes
   - DatasetView-generic serializer (zero-copy pack→text egress; CONSTRUCT reuses it) + parity test
   - Public DatasetView-generic reconstructor (subsumes the private pack reconstruct)
   - New purrdf-cli crate: convert / query / reason + global --loss-ledger
   - Uniform {text, mmap'd pack, stdin} source; --base ingress; --entailment pre-transform;
     --canonical RDFC egress; ValueEnum flags; exit-code taxonomy; fail-closed pack ingest
   - Workspace/Makefile wiring (clap dep, doc exclusion); make check/wasm/doc green
   "
   ```
3. Post this plan to the PR:
   ```bash
   ```
4. Post to BOTH the PR and issue #106:
   - What am I least confident about, and why?
5. Report: "Stage 1 complete. PR #<num> created for issue #106. Ready for /stage2 106".

## Execution Rules
1. Task 0 (worktree) is MANDATORY first; never work on `main`.
2. Commit after EACH task via github-commit-agent; never batch.
4. Every implementation sub-agent prompt carries the NO-DEFERRALS clause; reject any
   partial/"remaining work" summary and re-dispatch.
5. Task N (PR) is MANDATORY last.

paudley:
## Stage 1 reflections

**What am I least confident about, and why?**
The `sort_canonical` row-ordering change in the serializer (Task 1). It is a *universal* egress change — every `serialize_dataset*` call now emits rows in value-keyed order, not interning order. The full workspace suite (incl. sparql-conformance and the frozen GTS vectors) stays green and no golden moved, which is strong evidence it reproduces prior output for existing fixtures — but it is the change with the widest blast radius, and a byte-order regression would surface far from the CLI. The parity test pins RdfDataset↔PackView equality across all 9 formats, but broad byte-stability across the whole corpus rests on the existing suite rather than an exhaustive cross-product.

**What do I not know but should?**
Whether the RDF-1.2 base-direction drop I fixed for TriX/HexTuples has *siblings* — i.e., are there other structural drops (e.g. bnode scoping edge cases, or a datatype nuance) that a non-star-capable or non-dataset-capable format performs without a loss-ledger entry? I fixed the one I could demonstrate (direction), and the ledger contract for statement-rows and named-graphs is tested, but I have not exhaustively audited every (term-shape × target-format) pair for silent loss. A property-test enumerating all term shapes against all 9 targets and asserting "lossy ⇒ non-empty ledger" would close that gap and is a natural next hardening.

paudley:
# Stage 2 Gap Analysis & Remediation — Issue #106 / PR #115 (`purrdf-cli`)

**Worktree:** `.worktrees/106-purrdf-cli` · **Branch:** `paudley/106-purrdf-cli` · **PR:** #115
**Base:** `origin/main` · 30 files (+5147/−49) · Deferral scans (1 & 2): CLEAN · Clippy `-p purrdf-cli`: clean

## Context

Issue #106 asks for the first user-facing PurRDF binary — `purrdf convert | query | reason`
+ a global `--loss-ledger` — as glue over the existing codec / SPARQL / entailment engines.
PR #115 delivers it. Two independent adversarial audits (gap-analyzer + completion-adversary)
**built and drove the real binary** and confirmed the four written acceptance criteria all
execute in production:

- **convert** — full 81-pair (9×9) isomorphic matrix, extension-inference + explicit
  `--from/--to` override, pack→text zero-copy, `--canonical`, 5000-triple large input — all clean.
- **query** — identical SELECT over a text file and an mmap'd pack → byte-identical; all four
  W3C result formats (json/xml/csv/tsv); ASK; CONSTRUCT→turtle via the RDF sink; shape/format
  mismatch → non-zero; COUNT over the 5000-triple pack (zero-copy).
- **reason** — real inferences for `simple/rdf/rdfs/owl-rl`; boundary regimes `owl-direct/rif/d`
  exit 3 with a per-regime diagnostic; pack input reconstructs + reasons.
- **--loss-ledger** — empty on lossless, populated on lossy, `=PATH` writes the file, never leaks
  to stdout.

The completion-adversary verdict is **FINDINGS-OPEN** on exactly two defects (below). This plan
fixes every confirmed gap on this branch — no deferrals — and records the reviewer items that are
false positives / defend-not-fix with evidence.

---

## Confirmed gaps to fix (ranked)

### Gap A — [CRITICAL] `reason` stdin/stdout is completely unreachable
`crates/cli/src/cli.rs:141-155`, `crates/cli/src/reason.rs:56-63`, `crates/cli/src/format.rs:57-62`

The `Reason` variant has **no `--from`/`--to`** (unlike `Convert`, cli.rs:93-98), yet defaults both
`input` and `output` positionals to `"-"`, and its `--help` advertises `-` as stdin/stdout.
`reason::run` calls `format::resolve(None, input)` / `(None, output)`; `resolve` hard-rejects `"-"`
without an explicit format. Verified on the binary — **every** `-` path exits 2:
- `echo … | purrdf reason --regime rdfs` → exit 2 (the *default* invocation fails)
- `… | purrdf reason --regime rdfs - -` → exit 2
- `purrdf reason --regime rdfs g.ttl -` (file→stdout) → exit 2

Three compounding defects: (a) `reason` can never read stdin or write stdout, breaking the uniform
`{text, mmap'd pack, stdin}` pipeline the issue and plan promise; (b) the documented default
`reason --regime R FILE` fails; (c) the diagnostic tells the user to pass `--from/--to` flags that
**do not exist on `reason`** — a misleading error (violates `.baseline` "no misleading errors").
Untested: `reason_cli.rs` exercises only file paths with recognized extensions, masking the defect.
Root cause is the `convert`-has-overrides / `reason`-doesn't asymmetry (a One-Path/DRY break; both
share the `source`/`sink` seam and the identical `-` defaults).

**Fix (one commit):**
1. Add `--from`/`--to` `CliRdfFormat` args to the `Reason` variant, mirroring `Convert`.
2. Thread both through `main.rs` dispatch → `reason::run` → both `format::resolve` calls (mirror
   `ConvertOptions`). **Resolve both formats up front, before `load_dataset`/`materialize`** — this
   also fixes the fail-slow ordering (old Gap #5: OUT was resolved *after* the closure was computed).
3. Correct the `Reason` doc-comments and the misleading error text so an unresolvable `-` names the
   flags that actually exist.
4. Fix `crates/cli/README.md:178, 208-211` — the `reason` stdin/stdout example (`-` input, "to
   stdout" caption vs the `closure.ttl` target). Coupled to the same behavior; same commit.
5. Add binary-driven `reason_cli.rs` tests: stdin→stdout with `--from/--to`, and file→stdout with
   `--to`, asserting the inferred triple appears in stdout.

### Gap B — [MEDIUM] stdout `BrokenPipe` surfaces as a runtime error
`crates/cli/src/sink.rs:41-51` (and any other stdout/results/ledger writer)

Verified: `purrdf convert g.ttl - --to nt | head -c 10` → exit 1 with `purrdf: Broken pipe (os error
32)`. `write_out`'s `write_all`/`flush` propagate `ErrorKind::BrokenPipe` as `CliError::Runtime`.
The ubiquitous `purrdf … | head`/`| grep -q` idiom must exit 0 silently on downstream `EPIPE`.

**Fix (one commit):** in `write_out` (and the SPARQL-results stdout path if separate), match
`std::io::ErrorKind::BrokenPipe` → `Ok(())` / clean exit; propagate all other error kinds. Add a
binary-driven test piping into a short-reading consumer and asserting exit 0.

### Gap C — [LOW] `pack → pack` passthrough heap-buffers the whole pack
`crates/cli/src/convert.rs:92` (`source::read_bytes(input)`)

Not a zero-copy *contract* violation (that contract is "don't build the dataset IR"; the query and
pack→text paths **are** zero-copy, verified) — but the disk-pack passthrough materializes a
`Vec<u8>` where the existing `source.rs` mmap seam could borrow it. Aligns with `.goals` MAXIMAL
PERFORMANCE and the reviewer note.

**Fix (one commit):** for a disk pack input, mmap-borrow and `write_out(&mmap[..])`, buffering only
stdin. Output is already byte-identical (verified) — the existing passthrough test pins correctness;
no new bench (repo discipline: benches are report-only, never assert a speedup).

### Gap D — [LOW] test-coverage tightening
- `query --entailment owl-direct` exit-3 parity is asserted only for `reason`; `query.rs:10-16`
  claims it. Add the exit-3 boundary test to `crates/cli/tests/query_cli.rs`.
- Duplicate ~90-line `build_fixture()` in `crates/rdf/tests/pack_reconstruct_roundtrip.rs:33-124`
  and `serialize_pack_parity.rs:31-122` — extract a shared `tests/common` module (drift risk).

**Fix (one commit).**

---

## Reviewer items that are FALSE POSITIVES / defend-not-fix (respond on PR, no code change)

- **gemini HIGH `convert.rs:137`** — FALSE POSITIVE. Claims `canonical_flat_nquads` returns
  `Result<String, RdfDiagnostic>` needing `?`. Actual signature is `Result<String, String>`
  (`crates/rdf/src/native_quads.rs:171`); the current `.map_err(CliError::Runtime)` is correct and
  there is no `From<String> for CliError`, so `?` would not compile. Reply with the signature.
- **coderabbit MAJOR `source.rs:110` (TOCTOU/SIGBUS)** — DEFEND. The issue explicitly mandates
  "large inputs via mmap'd `PackView` without full load." The suggested fix ("read into an owned
  buffer… remove the unsafe mmap path") directly contradicts that acceptance criterion. The window
  is inherent to mmap, documented in the `SAFETY` block (`source.rs:33-38, 101-106`), behind a
  fail-closed `verify_pack` on a read-only, short-lived mapping. Same defend-not-fix stance as the
  fixed-key-ahash wasm rationale. Reply; keep the code.
- **coderabbit refactor `source.rs:110` (add Criterion bench)** — OPTIONAL, not a gap. Repo
  discipline: benches are report-only and we do not benchmark to optimize; library pack throughput
  is already covered by `crates/rdf-core/benches/pack_query.rs`. Reply noting this; no bench added.
- **coderabbit nitpick `ser_model.rs:87-134` (memoize `render_term`)** — SKIP. Perf micro-opt on the
  canonical sort; marginal value, would need a bench per repo rules.

---

## Execution (per-gap: fix → commit → push → verify; never batch)

For **each** of Gap A → B → C → D, in order:
1. **Task → general-purpose (sonnet)** implements the fix in `$WORKTREE`. Prompt carries the
   NO-DEFERRALS clause; a stub/downgrade/re-defer is a FAILED fix → re-dispatch.
2. **Task → github-commit-agent (Bash, sonnet)** — reads `PRECOMMIT_GUIDANCE.md`, iterates until
   `make check` passes (fmt + pedantic/nursery clippy + build + tests + hygiene).
3. **Task → github-push-agent (Bash, sonnet)** — pushes.
4. Post a per-gap summary (description, priority, fix, commit hash) to **issue #106 AND PR #115**.
5. **Task → wiring-verifier (sonnet)** — confirms the fix is wired into the production path.

After Gap A specifically: re-run the reason stdin→stdout invocation on the rebuilt binary and paste
the output into the PR summary (production-surface demonstration, not a unit-test claim).

Then post the four false-positive/defend replies to PR #115 with the evidence above.

## Verification (end-to-end, on the built binary)

- `cd $WORKTREE && cargo build -p purrdf-cli`
- **Gap A:** `printf '@prefix ex: <http://example.org/> .\nex:a a ex:Sub .\nex:Sub rdfs:subClassOf ex:Super .' | ./target/debug/purrdf reason --regime rdfs --from ttl --to nt - -` → exit 0, stdout contains the inferred `ex:a a ex:Super`. Also `./target/debug/purrdf reason --regime rdfs g.ttl - --to nt` (file→stdout) → exit 0.
- **Gap B:** `./target/debug/purrdf convert g.ttl - --to nt | head -c 10; echo "exit=${PIPESTATUS[0]}"` → `exit=0`, no "Broken pipe" on stderr.
- **Gap C:** `./target/debug/purrdf convert in.purrpck out.purrpck` → byte-identical to input; existing passthrough test green.
- **Gap D:** `cargo test -p purrdf-cli` + `cargo test -p purrdf-rdf` green; new query exit-3 test present.
- **Gates:** `make check` + `make wasm` + `make doc` all green (CLI native-only, excluded from wasm/doc by omission).

## Exit gate (Phase 6)

Re-dispatch the completion-adversary (STAGE=2) against the new HEAD. The stage is **not** ready for
`/stage3` until it returns PASS (every criterion `DEMONSTRATED-IN-PRODUCTION`) or every remaining
item carries verbatim human sign-off. Then post the final summary (with the pasted per-criterion
demonstrations) to issue #106 and PR #115, and append any signed-off exceptions to `.deficiencies`.

paudley:
## Gap A resolved — [CRITICAL] `reason` stdin/stdout now reachable

**Commit:** `d7e14b0` — *fix(cli): reach reason stdin/stdout via --from/--to + fail-fast format resolve*

**Defect:** `reason` had no `--from`/`--to` (unlike `convert`), yet defaulted both positionals to `-` and advertised stdin/stdout in `--help`. `format::resolve` rejects `-` without an explicit format, so **every** `-` path — including the default `reason --regime R FILE` (OUT defaults to `-`) and `… | reason --regime R` — exited 2, with an error naming flags that did not exist.

**Fix:**
- Added `--from`/`--to` `CliRdfFormat` overrides to the `Reason` variant, mirroring `Convert`; threaded through `main.rs` → `reason::run` → both `format::resolve` calls.
- **Both formats now resolved up front, before `load_dataset`/`materialize`** (also fixes the fail-slow ordering where OUT was resolved after the closure).
- Corrected the `Reason` doc-comments, the README `reason` usage line, and the mislabeled "to stdout" example.
- Added binary-driven `reason_cli.rs` tests: stdin→stdout and file→stdout, asserting the inferred triple in stdout.

**Production-surface demonstration (rebuilt binary):**
```
$ printf '@prefix ex: <http://example.org/> .\n@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .\nex:a a ex:Sub .\nex:Sub rdfs:subClassOf ex:Super .\n' \
    | purrdf reason --regime rdfs --from ttl --to nt - -
<http://example.org/a> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://example.org/Super> .
# exit 0
```

**Gate:** `make check` (fmt + pedantic/nursery clippy + workspace build/tests + hygiene guards + embedded `make wasm`) green. `cargo test -p purrdf-cli`: 63 tests pass (reason_cli now 11, incl. the 2 new stdin/stdout cases).

paudley:
## Gap B resolved — [MEDIUM] stdout `BrokenPipe` now a clean exit

**Commit:** `535c550` — *fix(cli): treat stdout BrokenPipe as a clean exit, not a runtime error*

**Defect:** writing to stdout (`-`) when the downstream consumer closes the pipe early (`purrdf … | head`, `| grep -q`) failed with `ErrorKind::BrokenPipe`, propagated as `CliError::Runtime` → exit 1 + `purrdf: Broken pipe (os error 32)` on stderr.

**Fix:** `sink::write_out` now silences `BrokenPipe` **only** on the stdout sink (every other error kind still propagates; file writes unchanged — no swallowed errors). Covers all stdout data paths: convert/reason via the shared sink, and query's SELECT/ASK results + CONSTRUCT/DESCRIBE graph result. The `--loss-ledger` path is stderr and intentionally untouched.

**Production-surface demonstration (rebuilt binary):**
```
$ purrdf convert g.ttl - --to nt | head -c 5 ; echo " exit=${PIPESTATUS[0]}"
<http exit=0        # (no "Broken pipe" on stderr)
```

**Test:** `stdout_broken_pipe_exits_clean` drives the built binary through a real shell pipe into `head -c 5` with a >64 KiB (20k-triple) fixture so the OS delivers EPIPE mid-write; asserts purrdf's own `PIPESTATUS[0] == 0` and no "broken pipe" on its (isolated) stderr. Verified falsifiable: reverting the fix makes it fail with exit 1.

**Gate:** `make check` green (fmt + pedantic/nursery clippy + workspace tests + hygiene + `make wasm`).

paudley:
## Gap C resolved — [LOW] `pack → pack` passthrough now mmap-borrows the disk pack

**Commit:** `e072a3d` — *perf(cli): mmap-borrow disk pack→pack passthrough instead of heap-buffering*

**Defect:** the disk pack→pack passthrough read the entire pack into a `Vec<u8>` before `verify_pack`/`write_out`, inconsistent with the existing zero-copy mmap seam used by the query/serialize pack paths (`.goals` MAXIMAL PERFORMANCE). Not a zero-copy *contract* violation (the contract is "don't build the dataset IR"), but an avoidable heap copy.

**Fix:** added `source::verified_pack_mmap(path)` (open → `Mmap::map` → `verify_pack`, carrying the read-only SAFETY caveat) and routed the disk pack passthrough through it — the borrowed `&mmap[..]` goes straight to `write_out`, no `Vec` copy of pack contents. The new helper also **subsumes** the previously-inline open+mmap+verify in `run_over_input`/`load_dataset` (DRY). Only stdin (`-`, no file to map) still buffers. Corrupt packs still fail-closed unchanged; output byte-identical.

**Tests:** `disk_pack_to_pack_is_byte_identical_passthrough` (build pack → `convert in.purrpck out.purrpck` → assert bytes equal) and `disk_pack_to_pack_corrupt_input_fails_closed` (garbage pack → non-zero exit, no partial output). Both drive the built binary.

**Gate:** `make check` green (fmt + pedantic/nursery clippy + workspace tests + hygiene + `make wasm`).

paudley:
## Gap D resolved — [LOW] test-coverage tightening

**Commit:** `0286643` — *test(cli,rdf): pin query --entailment exit-3 boundary + dedup pack test fixture*

**D1 — query `--entailment` boundary was unasserted.** `reason`'s unsupported-regime boundary (owl-direct/rif/d → exit 3) was tested, but the identical boundary on `query --entailment` was not, though `query.rs` relies on it. Added `entailment_boundary_regimes_exit_three` (`query_cli.rs`): drives the built binary with `query --data <ttl> --entailment {owl-direct,rif,d} --results-format json '<SELECT>'` and asserts `status.code() == Some(3)` + the "cannot be materialized" diagnostic, mirroring reason's boundary test.

**D2 — duplicated `build_fixture()`.** The ~90-line rich RDF-1.2 fixture builder was copy-pasted in `pack_reconstruct_roundtrip.rs` and `serialize_pack_parity.rs` (drift risk). Diffed byte-for-byte — **identical, no drift** — and extracted verbatim into a shared `crates/rdf/tests/common/mod.rs` (`pub(crate) fn build_fixture`); both tests now `mod common;` and call it. Pure DRY extraction; parity/roundtrip assertions unchanged.

**Gate:** `make check` green (fmt + pedantic/nursery clippy + workspace tests + hygiene + `make wasm`); both pack tests + the new query test pass.

paudley:
## Reviewer-comment adjudication (false positive / defend-not-fix)

Four automated-review items were evaluated adversarially against the built binary and the issue #106 acceptance criteria. Each is **not** a code change, with evidence:

**1. gemini (HIGH) `convert.rs:137` — "type mismatch, use `?` on `canonical_flat_nquads`" — FALSE POSITIVE.**
The suggestion assumes `canonical_flat_nquads` returns `Result<String, RdfDiagnostic>`. Its actual signature is `pub fn canonical_flat_nquads(dataset: &RdfDataset) -> Result<String, String>` (`crates/rdf/src/native_quads.rs`). The error is already a `String`, so the current `.map_err(CliError::Runtime)` is correct; there is no `From<String> for CliError`, so the proposed `?` would not compile. No change.

**2. coderabbit (MAJOR) `source.rs` — "TOCTOU/SIGBUS on mmap'd pack; read into an owned buffer / remove the unsafe mmap path" — DEFEND, do not fix.**
Issue #106 explicitly requires "large inputs via mmap'd `PackView` **without full load**." The proposed remediation (buffer the bytes, drop mmap) directly contradicts that acceptance criterion and the `.goals` zero-copy / MAXIMAL-PERFORMANCE mandate. The window is inherent to mmap and is documented in the module's `SAFETY` block (read-only, short-lived mapping behind a fail-closed `verify_pack`). This is the same defend-not-fix posture the project takes elsewhere for deliberate low-level choices. The mmap path stays.

**3. coderabbit (refactor) `source.rs` — "add a Criterion bench for the mmap pack path" — OPTIONAL, not adopted.**
Repo discipline: benches are report-only and we do not benchmark to optimize; library pack throughput is already covered by `crates/rdf-core/benches/pack_query.rs`. Adding a CLI-level throughput bench asserts nothing and is not a gap.

**4. coderabbit (nitpick) `ser_model.rs` — "memoize `render_term`" — skipped.**
A micro-optimization on the canonical sort with marginal value that would require extending a Criterion bench per repo rules. Not adopted.

The remaining review items were **confirmed defects and fixed** on this branch (see the per-gap comments above): `reason` stdin/stdout (`cli.rs`, MAJOR), README `reason` docs, and stdout `BrokenPipe` (`sink.rs`, MEDIUM).

paudley:
## Stage 2 complete — gap remediation + exit-gate PASS

All confirmed gaps fixed on `paudley/106-purrdf-cli`; every fix committed, pushed, and independently wiring-verified. The Stage-2 completion-adversary was **re-run against the new HEAD as the exit gate and returned PASS** — all nine acceptance criteria DEMONSTRATED-IN-PRODUCTION on the built `purrdf` binary.

### Gaps resolved (each: fix → commit → push → wiring-verify)
| # | Pri | Gap | Commit |
|---|-----|-----|--------|
| A | CRITICAL | `reason` stdin/stdout unreachable (no `--from/--to`; every `-` path exited 2 with a misleading error) → added `--from/--to`, resolve up front, fixed docs/README + tests | `d7e14b0` |
| B | MEDIUM | stdout `BrokenPipe` surfaced as exit 1 → clean exit on EPIPE (stdout only; other errors still propagate) + falsifiable test | `535c550` |
| C | LOW | `pack→pack` passthrough heap-buffered → mmap-borrow via shared `verified_pack_mmap` (also DRY-subsumed the inline mmap arms) | `e072a3d` |
| D | LOW | untested `query --entailment` exit-3 boundary + duplicated test fixture → exit-3 test + shared `tests/common/mod.rs` | `0286643` |

Four automated-review items were adjudicated as false-positive / defend-not-fix with evidence (gemini `convert.rs:137` wrong signature; coderabbit TOCTOU defended per the mmap acceptance criterion; bench + memoize declined per repo bench discipline) — see the adjudication comment above.

### Exit-gate completion-adversary verdict — PASS (per-criterion demonstrations)
- **[C1] convert 9×9 matrix** — all 81 ordered (src,dst) pairs converted + re-normalized + diffed vs canonical: `MATRIX: pass=81 fail=0`.
- **[C2] extension-inferred OR explicit `--from/--to`** — `convert base.ttl inferred.nt` (no flags) → N-Triples; explicit flags throughout C1.
- **[C3] large input via mmap'd pack, no full load** — `ttl→pack→nq` matched reference; `pack→pack` byte-identical via mmap-borrow (`source.rs:89-97`).
- **[C4] query over file AND mmap'd pack; SRJ/CSV/TSV/XML** — SELECT over file & pack byte-identical; all four result formats exit 0; ASK→`{"boolean":true}`; CONSTRUCT→turtle & DESCRIBE→ntriples via the RDF sink; SELECT+turtle mismatch hard-fails exit 1.
- **[C5] reason per supported regime; unsupported → exit 3** — simple/rdf/rdfs/owl-rl exit 0 (2/4/8/8 triples); owl-direct/rif/d exit 3; `query --entailment owl-direct` exit 3; `query --entailment rdfs` materializes the closure.
- **[C6] F1 CLOSED** — `… | purrdf reason --regime rdfs --from ttl --to nt - -` → exit 0, output has `ex:a a ex:Super`; file→stdout reason → exit 0.
- **[C7] F2 CLOSED** — `convert g.ttl - --to nt | head -c 5` → `exit=0`, stderr empty (no "Broken pipe").
- **[C8] `--loss-ledger` tri-state** — lossy → non-empty ledger on stderr (data-only on stdout); `=PATH` writes the file; absent → nothing anywhere; lossless → `"losses":[]`; never on stdout.
- **[C9] gates + determinism + no minted IRIs** — `make check` / `make doc` / wasm all exit 0; repeated conversions byte-identical; `example.org`-only, no fabricated IRIs.

**Open findings: none. Laundering scans: clean. Follow-up hunt: clean (no #106 work laundered).**

Ready for **/stage3 106**.

paudley:
feat(cli): the purrdf CLI — convert / query / reason

Closes #106. First user-facing PurRDF binary: new native-only crate `purrdf-cli`
(binary `purrdf`), glue over the existing codec / SPARQL / entailment engines,
structured as one `Source → [entail]? → Sink` pipeline. `convert` / `query` /
`reason` + a global `--loss-ledger`. Query and convert stream an mmap'd `.purrpck`
pack with zero materialization; the native serializer was generalized over
`DatasetView`, unifying convert's RDF sink with CONSTRUCT/DESCRIBE query results.

## What this delivers
- **convert** `--from --to [--base] [--entailment] [--canonical] IN OUT` — any↔any
  over the 9 syntaxes (ttl/nt/nq/trig/rdfxml/trix/hext/jsonld/yamlld) + the native
  `pack` container; extension inference or explicit; stdin/stdout `-`; mmap'd
  zero-copy pack path; `--canonical` = RDFC-1.0 N-Quads.
- **query** `--data <file|pack> [--base] [--entailment] --results-format <…>` —
  SELECT/ASK → W3C results (json/xml/csv/tsv); CONSTRUCT/DESCRIBE → the unified RDF
  sink; zero-copy pack query; shape/format mismatch hard-errors.
- **reason** `--regime <simple|rdf|rdfs|owl-rl> [--from] [--to] [--base] IN OUT` —
  via `materialize`; owl-direct/rif/d → per-regime hard error (exit 3).
- **--loss-ledger[=PATH]** — absent→silent, bare→stderr, =PATH→file; always
  computed, never leaks to stdout. Exit codes 0/1/2/3.

## Library changes (subsume / extend, source-compatible)
- `DatasetView`-generic native serializer (`purrdf-rdf`): a `PackView` or any view
  (incl. a CONSTRUCT result graph) serializes to any of the 9 syntaxes with zero
  materialization; canonical value-keyed row sort → byte-identical, insertion-order
  independent output. + byte-parity test.
- Public `dataset_from_view<D: DatasetView>` (`purrdf-core`): subsumes the private
  pack `reconstruct` loop into one shared generic bridge.
- Bug fix: TriX/HexTuples base-direction drop now recorded in the loss ledger
  (`rdf12-direction-dropped`) instead of being a silent loss.

## Files
- NEW crate `crates/cli/` (src: main/cli/format/source/sink/convert/query/reason/
  ledger/error; tests: convert/query/reason/loss_ledger/help/smoke_cli).
- `crates/rdf/src/native_codecs/{serialize,ser_model,media_type}.rs` — DatasetView-
  generic serialize + deterministic ordering + direction-loss accounting.
- `crates/rdf-core/src/ir/pack/*` — public `dataset_from_view`, shared
  `verified_pack_mmap`.
- `crates/rdf/tests/common/mod.rs` (new shared fixture), pack parity/roundtrip tests.
- `Cargo.toml` (clap workspace dep + `crates/cli` member), `Makefile` (`--exclude
  purrdf-cli` from `make doc`; wasm allowlist excludes the CLI by omission).

## Stage-2 gap remediation (all fixed on this branch, none deferred)
- [CRITICAL] `reason` stdin/stdout was unreachable (no `--from/--to`; every `-` path
  exited 2 with a misleading error) → added `--from/--to`, resolve formats up front
  (also fixes fail-slow), corrected docs/README, added stdin→stdout + file→stdout
  tests. (d7e14b0)
- [MEDIUM] stdout `BrokenPipe` surfaced as exit 1 → clean exit on EPIPE for the
  stdout sink only; all other error kinds still propagate; falsifiable test. (535c550)
- [LOW] `pack→pack` heap-buffered → mmap-borrow via shared `verified_pack_mmap`
  (also DRY-subsumed the inline mmap arms); corrupt-pack still fail-closed. (e072a3d)
- [LOW] untested `query --entailment` exit-3 boundary + duplicated test fixture →
  exit-3 test + shared `tests/common/mod.rs`. (0286643)

Four automated-review items were adjudicated false-positive / defend-not-fix with
evidence: gemini `convert.rs:137` (wrong signature; `canonical_flat_nquads` returns
`Result<String, String>`); coderabbit TOCTOU on the mmap pack (DEFENDED — buffering
would contradict #106's "mmap'd PackView without full load" acceptance criterion +
.goals zero-copy; documented in the SAFETY block); the Criterion-bench and
render_term-memoize suggestions declined per repo bench discipline.

## `.goals` standing constraints — respected
GREENFIELD/no-compromise; RUST-FIRST (native binary, Python surface untouched);
SUBSUME/EXTEND/ENHANCE (generalized the serializer + reconstructor rather than
bolting on); LOW/NO OPTIONALITY + HARD FAILS (no Cargo features on any member crate;
fail-closed pack ingest; unsupported regimes exit 3; no swallowed errors —
BrokenPipe is silenced only on the stdout sink, by Unix-filter convention);
RDF-1.2-ONLY (star layer round-trips; RDF-1.2 statement layer preserved); MAXIMAL
PERFORMANCE (zero-copy pack query + pack→text egress + pack→pack passthrough) /
PORTABILITY (every library crate stays wasm32-clean; `purrdf-cli` is native-only,
excluded from the wasm allowlist by omission).

## Completion-Adversary — Stage-2 exit gate returned PASS
Re-run against HEAD; all nine acceptance criteria DEMONSTRATED-IN-PRODUCTION on the
built binary (no unit-test/code-reading proxies):
- [C1] convert 9×9 matrix — all 81 ordered (src,dst) pairs converted, re-normalized,
  diffed vs canonical: `MATRIX: pass=81 fail=0`.
- [C2] extension-inferred OR explicit `--from/--to` — both driven.
- [C3] large input via mmap'd pack, no full load — ttl→pack→nq matched ref; pack→pack
  byte-identical via mmap-borrow.
- [C4] query over file AND mmap'd pack; SRJ/CSV/TSV/XML — file & pack SELECT
  byte-identical; all four result formats; ASK→boolean; CONSTRUCT→turtle,
  DESCRIBE→ntriples via the RDF sink; SELECT+turtle mismatch hard-fails exit 1.
- [C5] reason per supported regime; unsupported → exit 3 — simple/rdf/rdfs/owl-rl
  exit 0 (2/4/8/8 triples); owl-direct/rif/d and `query --entailment owl-direct`
  exit 3.
- [C6] reason stdin↔stdout reachable — `… | reason --regime rdfs --from ttl --to nt
  - -` → exit 0 with the inferred superclass triple; file→stdout → exit 0.
- [C7] stdout BrokenPipe clean — `convert g.ttl - --to nt | head -c 5` → exit 0,
  empty stderr.
- [C8] `--loss-ledger` tri-state — lossy→non-empty ledger on stderr (data-only on
  stdout); `=PATH` writes the file; absent→nothing; lossless→`"losses":[]`; never on
  stdout.
- [C9] gates + determinism + no minted IRIs — `make check` / `make doc` / `make wasm`
  exit 0; repeated conversions byte-identical; `example.org`-only, no fabricated IRIs.
Open findings: none. Laundering scans: clean. Follow-up hunt: clean.

## Deferral gate — clean
Mechanical deferral scan over the branch diff: no TODO/FIXME/unimplemented/"future
work"/stub hits. No `.goals` compromise, no self-authored follow-up carrying #106's
own work.

## CI note (human-authorized merge over pre-existing red)
The **Build The PurRDF Book** job (`playground-smoke`) is red with two stale JSON-LD
assertions in `docs/playground/smoke/engine.smoke.test.mjs` ("jsonld parse must
throw" / "JSON-LD serialize must hard-fail for a quoted-triple"). These are
**pre-existing on `main`** — introduced by the JSON-LD/YAML-LD native-registry
unification (the prior `main` push was green on this job) and failing identically
there before #106 existed. The failure is NOT a #106 regression: this branch does
not touch `docs/playground/`. Tracked separately as #116; merge of #106 authorized
by the maintainer.

- `purrdf-cli` is native-only by design (memmap2); it is kept out of `make wasm` by
  omission from the `-p` allowlist and out of `make doc` via `--exclude`. Do not add
  it to either.
- The serializer is now generic over `DatasetView` (`D::Id` throughout); `&RdfDataset`
  still satisfies the trait so all prior call sites compile unchanged.
- The mmap pack path is read-only / short-lived behind a fail-closed `verify_pack`;
  the SAFETY block documents the no-concurrent-mutation caveat — do not "fix" it into
  a buffered read (that would violate the zero-copy acceptance criterion).

Issue #142, opened 2026-07-17
structured-zstd 0.0.40 huff0 encoder panics ("internal error") composing large gts bundles — bump past 0.0.40


paudley:
## Summary

`purrdf`'s GTS bundle composer (`purrdf::gts_compose::emit_gts`) **panics deterministically** inside the pinned `structured-zstd 0.0.40` Huffman encoder when compressing a large multi-blob bundle:

```
thread '<unnamed>' panicked at
  structured-zstd-0.0.40/src/huff0/huff0_encoder.rs:643:13:
This is an internal error
```

It reproduces on every run at the same spot (two consecutive full runs), in a worker thread during the final `emit_gts` compression step. It is not a resource/OOM condition — it is an internal invariant violation in the `huff0` (entropy/literals) encoder, triggered by the specific byte distribution of the compressed content.

## Where it's pinned

`crates/purrdf-columnar/Cargo.toml` pins:

```toml
structured-zstd = "0.0.40"
```

For a `0.0.x` crate this resolves to **exactly 0.0.40**, so a downstream `cargo update -p structured-zstd --precise 0.0.49` is rejected:

```
error: failed to select a version for the requirement `structured-zstd = "^0.0.40"`
candidate versions found which didn't match: 0.0.49
required by package `purrdf-columnar v0.7.0`
```

and a downstream `[patch.crates-io]` to a newer version is ignored for the same requirement mismatch. Newer `structured-zstd` releases exist (0.0.45, 0.0.49). `rust-v0.7.0` is the latest purrdf tag and every tag pins `0.0.40`, so downstreams have no way to move off the panicking encoder without a purrdf change.

## Impact

A downstream consumer (the `gmeow-ontology` `gmeow.gts` terminal bundle) can no longer regenerate its bundle: `make sync` fails at the `emit_gts` compression step. The content that triggers it is legitimate (a growing multi-graph RDF-1.2 + docs/archive payload); the same content compressed fine at a slightly smaller size, so the trigger appears to be a data-distribution edge case in `huff0` that larger/different input reaches.

## Request

1. **Bump `structured-zstd` past 0.0.40** (0.0.49 is current) in `purrdf-columnar` and cut a `rust-v0.7.1` tag — if the newer `huff0` encoder fixes the internal-error path. (Please verify against a bundle large enough to trip 0.0.40.)
[...]

Issue #12, opened 2026-07-02
SHACL-SPARQL: custom constraint components (sh:ConstraintComponent)


paudley:
## Goal

Implement SHACL-SPARQL **custom constraint components** — the plugin system that lets shape authors declare reusable, parameterized components with their own SPARQL validators.

Covers the last 3 SHACL-SPARQL component tests in the W3C ledger (`crates/shapes/tests/w3c_conformance.rs` `XFAIL`):
- `sparql/component/optional-001` — `sh:SPARQLAskValidator` (`sh:ask`) with optional parameters
- `sparql/component/propertyValidator-select-001` — `sh:propertyValidator` SELECT with `$PATH` substitution
- `sparql/component/validator-001` — ASK validator over `$value`

## Deliverables

- [ ] Component registry parsed from the shapes graph (`sh:ConstraintComponent`, `sh:parameter` with `sh:path`/`sh:optional`)
- [ ] Validator selection per spec §6 (`sh:validator` / `sh:nodeValidator` / `sh:propertyValidator`; ASK vs SELECT semantics)
- [ ] `$PATH` textual substitution for property-shape validators; parameter pre-binding
- [ ] The 3 ledger entries removed (XPASS discipline enforces this); scoreboard 117/120+

## Gates

`cargo test -p purrdf-shapes` green incl. the W3C harness exact counts; clippy -D warnings; wasm32-clean; no Cargo features.

paudley:
# Implementation Plan for Issue #12

## Issue Summary

Implement SHACL-SPARQL **custom constraint components** (`sh:ConstraintComponent`) so shape authors can declare reusable, parameterized constraint components with their own SPARQL validators. The work must make the three currently-failing W3C SHACL-SPARQL component tests pass and reach the scoreboard target of **117/120** passing tests.

Key requirements derived from issue #12 and the W3C SHACL-SPARQL spec §6:

- Parse a component registry from the shapes graph: `sh:ConstraintComponent`, `sh:parameter` (with `sh:path` and `sh:optional`), and validators (`sh:validator`, `sh:nodeValidator`, `sh:propertyValidator`).
- Select validators per context: node shapes prefer `sh:nodeValidator`; property shapes prefer `sh:propertyValidator`; fall back to `sh:validator`.
- Support `sh:SPARQLAskValidator` (`sh:ask`) and `sh:SPARQLSelectValidator` (`sh:select`).
[...]

PR #144, opened 2026-07-18
fix(gts): make zstd encode total against huff0 encoder panics

paudley:
Refs #142

## Context
A downstream consumer (`gmeow-ontology`, generating `gmeow.gts`) hit a deterministic panic — `structured-zstd-0.0.40 huff0_encoder.rs "This is an internal error"` — inside PurRDF's GTS composer (`emit_gts`) during final zstd compression of a large multi-blob bundle. The workspace dep was already bumped `0.0.40 → 0.0.49` on main (commit 80aebbc), but (a) the fix was **unverified** and (b) the encode path had **zero fault tolerance**: `compress_to_vec` returns `Vec<u8>` (not `Result`), so an internal encoder invariant violation could only surface as a panic that unwinds through `emit_gts`.

The huff0 panic guard is **still compiled in 0.0.49** (moved to `huff0_encoder.rs:746`); the real fix is behavioral — upstream PR #424 ("skip FSE weight description for streams it cannot encode", released 0.0.41) — which changes *which* streams reach the guard.

## What this PR does
Makes GTS zstd encoding **total** — `encode : Bytes → Bytes` was a partial function (undefined on huff0-degenerate inputs); a `catch_unwind` guard closes it into a total function whose bottom element is a raw-literals zstd frame (a valid instance of the codec's own output space, so `decode ∘ encode = id` with no new wire concept).

- `crates/gts/src/codec.rs`: `guarded_encode` combinator; plain/rsyncable encode degrades to a `CompressionLevel::Uncompressed` (raw-literals) frame on an internal panic — decode-compatible and deterministic. The **dict path returns `CodecError::Failed`** rather than a dict-less frame, because the reader registers the dict and the existing mismatch test shows the decoder refuses dict mismatches (a dict-less fallback could be silently *undecodable* — strictly worse than the panic).
- `frame_block_kinds`/`BlockKind`: read a zstd frame's block types without decompressing (distinguishes native compression from a degraded fallback).
- `crates/gts/tests/encode_never_panics.rs`: standing class-of-bugs guarantee — `encode_chain` (zstd + zstd-rsyncable) never unwinds over a generated input spread (with decode round-trip byte-identity); `decode_chain` and `frame_block_kinds` never panic on arbitrary/truncated bytes (guards the untrusted-reader surface against a structured-zstd internal panic becoming a DoS).

## Verification
Clippy `--workspace --all-targets -D warnings`, **3693 tests / 139 suites, 0 failed**, all six hygiene checks, `make wasm`, and `make doc` — all green. Determinism preserved (`snapshot_content_id_is_order_independent`, `codec_recompression_preserves_refold_equivalence`, dict compaction round-trips).

## Honest limitation on requirement 3
The issue asked to *verify the bump against a bundle large enough to actually trip the 0.0.40 panic*. **This could not be independently reproduced.** The =0.0.40 trigger did NOT reproduce from 25 MB of diverse real repo bytes (198,184 windows, sizes 1 KiB–1 MiB, levels 3/12/19 + the DIST-profile rsyncable 64 KiB path) nor from theory-driven synthetics targeting PR #424's degenerate weight histograms (`max_count==1` / `max_count==len`). The real gmeow bundle was unavailable and the maintainer authorized skipping the reproduction. Verification of the 0.0.49 fix therefore rests on **upstream PR #424 + the generic `catch_unwind` guard added here**, not on a committed regression test that reproduces the specific input. Recorded in `.deficiencies`.

## Out of scope (maintainer-directed)
Version bump / changelog / `rust-v0.7.x` release tag are **batched across open tickets** and cut later — not in this PR.

## Summary by CodeRabbit

- **New Features**
  - Added frame inspection support to identify whether compressed data blocks are raw, run-length encoded, or compressed.
  - Added clear error reporting for malformed or truncated compression frames.

- **Bug Fixes**
  - Compression now avoids unexpected crashes and safely falls back when standard encoding encounters an internal failure.
  - Dictionary-based compression reports an error instead of producing unusable output.

- **Tests**
  - Expanded coverage for compression round trips, malformed input, truncated frames, and panic-free processing across varied data.

### DIFF
--- a/crates/gts/src/codec.rs
+++ b/crates/gts/src/codec.rs
@@ -11,8 +11,9 @@
 use std::borrow::Cow;
 use std::fmt;
 use std::io::{Read, Write};
+use std::panic::{AssertUnwindSafe, catch_unwind};
 
-use structured_zstd::decoding::{FrameDecoder, errors::FrameDecoderError};
+use structured_zstd::decoding::{FrameDecoder, errors::FrameDecoderError, read_frame_header_info};
 use structured_zstd::encoding::{CompressionLevel, StreamingEncoder, compress_to_vec};
 
 /// A catalog entry (§5, §8.5).
@@ -139,8 +140,32 @@ fn decode_one(codec: &Codec, data: &[u8]) -> Result<Vec<u8>, CodecError> {
 
 const RSYNCABLE_BLOCK_SIZE: usize = 65_536;
 
+/// Run a zstd encode that may panic on an internal encoder invariant
+/// (structured-zstd `huff0_encoder.rs` "internal error"); on unwind, degrade
+/// per the caller. Deterministic: a given input either always encodes or
+/// always degrades. Mirrors the panic-totalization pattern in
+/// `crates/rdf/src/native_codecs/parse.rs`.
+///
+/// `AssertUnwindSafe` is sound here: the input slice is untouched by the
+/// encode, and the poisoned encoder/buffers are dropped on unwind, so no
+/// shared mutable state crosses the catch boundary. On `wasm32` (panic=abort)
+/// `catch_unwind` compiles cleanly but does not catch — the guard is a
+/// compile-clean no-op there; the panic is a native-producer concern.
+fn guarded_encode<T>(f: impl FnOnce() -> T, on_panic: impl FnOnce() -> T) -> T {
+    match catch_unwind(AssertUnwindSafe(f)) {
+        Ok(v) => v,
+        Err(_) => on_panic(),
+    }
+}
+
 fn encode_zstd(data: &[u8], level: Option<i32>) -> Vec<u8> {
-    compress_to_vec(data, zstd_level(level))
+    // `Uncompressed` produces a valid Zstandard frame the existing
+    // `FrameDecoder` already round-trips, and it cannot itself build a
+    // Huffman table — so the fallback can never re-trip this guard.
+    guarded_encode(
+        || compress_to_vec(data, zstd_level(level)),
+        || compress_to_vec(data, CompressionLevel::Uncompressed),
+    )
 }
 
 fn encode_zstd_with_dict(
@@ -151,10 +176,27 @@ fn encode_zstd_with_dict(
     let mut enc = StreamingEncoder::new(Vec::<u8>::new(), zstd_level(level));
     enc.set_dictionary_from_bytes(dict)
         .map_err(|e| CodecError::Failed(format!("zstd dictionary load failed: {e}")))?;
-    Write::write_all(&mut enc, data)
-        .map_err(|e| CodecError::Failed(format!("zstd dict encode failed: {e}")))?;
-    enc.finish()
-        .map_err(|e| CodecError::Failed(format!("zstd dict encode failed: {e}")))
+    // Unlike `encode_zstd`, a dict-primed encode has no safe raw-literals
+    // degrade: the reader registers this dictionary against the frame, and
+    // the mismatch test (`zstd_dict_codec_rejects_a_mismatched_dictionary`)
+    // shows the decoder refuses a dict/frame mismatch outright — a
+    // dict-less fallback frame could be silently undecodable, strictly
+    // worse than the panic it would replace. Returning `Err` keeps encoding
+    // total (no panic escapes this function) without ever emitting
+    // undecodable bytes.
+    guarded_encode(
+        move || {
+            Write::write_all(&mut enc, data)
+                .map_err(|e| CodecError::Failed(format!("zstd dict encode failed: {e}")))?;
+            enc.finish()
+                .map_err(|e| CodecError::Failed(format!("zstd dict encode failed: {e}")))
+        },
+        || {
+            Err(CodecError::Failed(
+                "zstd dict encode aborted (huff0 internal error)".into(),
+            ))
+        },
+    )
 }
 
 fn encode_zstd_rsyncable(data: &[u8], level: Option<i32>) -> Vec<u8> {
@@ -266,6 +308,70 @@ pub fn decode_chain_with_decrypt(
         }
     }
     Ok(current.into_owned())
+}
+
+/// A zstd block's on-wire kind, read from its 3-byte `Block_Header` (RFC 8878
+/// §3.1.1.2) without decompressing.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BlockKind {
+    /// `Block_Type` 0: the block body is the literal, uncompressed bytes.
+    Raw,
+    /// `Block_Type` 1: the block body is a single byte repeated `Block_Size` times.
+    Rle,
+    /// `Block_Type` 2: the block body is Huffman/FSE-compressed (the normal path).
+    Compressed,
+}
+
+/// The block types in a zstd frame, read from block headers WITHOUT
+/// decompressing. Lets a caller distinguish a natively-compressed frame from a
+/// raw-literals degrade. A test/inspection utility (no production degradation
+/// logging is wired in this change).
+///
+/// Parses per RFC 8878: the frame header (magic + descriptor, read via
+/// `structured_zstd::decoding::read_frame_header_info` so the magic/flag
+/// decoding never drifts from the crate's own decoder), then a sequence of
+/// 3-byte `Block_Header`s (bit0 `Last_Block`, bits1-2 `Block_Type`, bits3-23
+/// `Block_Size`) until `Last_Block`. The trailing content checksum, if any,
+/// is not consumed — it is not needed to classify block kinds.
+///
+/// # Errors
+/// `CodecError::Failed` on a malformed/truncated frame header, a truncated
+/// block, or a block declaring the reserved `Block_Type`; never panics (every
+/// slice access is checked).
+pub fn frame_block_kinds(frame: &[u8]) -> Result<Vec<BlockKind>, CodecError> {
+    let info = read_frame_header_info(frame, false)
+        .map_err(|e| CodecError::Failed(format!("zstd frame header read failed: {e}")))?;
+    let mut offset = info.header_size;
+    let mut kinds = Vec::new();
+    loop {
+        // 3-byte Block_Header (RFC 8878 §3.1.1.2), little-endian: bit0 =
+        // Last_Block, bits1-2 = Block_Type, bits3-23 = Block_Size.
+        let header = frame
+            .get(offset..offset + 3)
+            .ok_or_else(|| CodecError::Failed("zstd frame block header truncated".into()))?;
+        let raw = u32::from(header[0]) | (u32::from(header[1]) << 8) | (u32::from(header[2]) << 16);
+        let last_block = raw & 1 != 0;
+        let block_size = (raw >> 3) as usize;
+        let (kind, on_disk) = match (raw >> 1) & 0b11 {
+            0 => (BlockKind::Raw, block_size),
+            1 => (BlockKind::Rle, 1),
+            2 => (BlockKind::Compressed, block_size),
+            _ => {
+                return Err(CodecError::Failed(
+                    "zstd frame block has reserved Block_Type".into(),
+                ));
+            }
+        };
+        kinds.push(kind);
+        offset = offset
+            .checked_add(3 + on_disk)
+            .filter(|end| *end <= frame.len())
+            .ok_or_else(|| CodecError::Failed("zstd frame block body truncated".into()))?;
+        if last_block {
+            break;
+        }
+    }
+    Ok(kinds)
 }
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


paudley:
# Implementation Plan for Issue #142

## Context

Downstream consumer `gmeow-ontology` (generating `gmeow.gts`) hits a
**deterministic panic** — `structured-zstd-0.0.40 huff0_encoder.rs:643
"This is an internal error"` — inside PurRDF's GTS bundle composer
(`emit_gts`) during final zstd compression of a large multi-blob bundle. It is
an internal invariant violation in the huff0 (literals/entropy) encoder
triggered by a specific literal byte distribution, **not** OOM.

The workspace dependency was already bumped `structured-zstd 0.0.40 → 0.0.49`
on `main` (commit `80aebbc`, `Cargo.toml:133`, resolved in `Cargo.lock`). Two
things the issue asked for are therefore **still open**:

1. The bump is **unverified** — critically, the panic guard is *still compiled*
   in 0.0.49 (moved to `huff0_encoder.rs:746`, byte-identical body). The real
   fix is behavioral: upstream PR #424 ("skip FSE weight description for streams
   it cannot encode", released in 0.0.41) changes *which* streams reach the
   guard. So a source grep proves nothing — only replaying a genuinely
   triggering input through the **raw, unguarded** encoder demonstrates the fix.
2. The encode path has **zero fault tolerance**: `compress_to_vec` returns
   `Vec<u8>` (not `Result`), so any internal encoder failure can *only* surface
   as a panic that unwinds through `emit_gts`. Decode is defensive; encode is
   not (`crates/gts/src/codec.rs:142`).

**Intended outcome:** make the GTS encode path *total* — `encode : Bytes → Bytes`
is currently a partial function (undefined on huff0-degenerate inputs);
`catch_unwind` closes it into a total function whose bottom element is a
raw-literals zstd frame (a valid instance of the codec's *own* output space, so
`decode ∘ encode = id` with no new wire concept). And land a permanent
regression test proving a real huff0-triggering bundle now composes **natively**
(not via fallback) under 0.0.49.

None — no `.baseline` file present in the repo.

None — no `.goals` file present. Standing repo constraints from `AGENTS.md`/
`CLAUDE.md` that this plan MUST satisfy:
- No new Cargo features; dependency pins live only in root `[workspace.dependencies]`.
- No nondeterminism in serializers / GTS writer — output stays byte-reproducible.
- Every release crate (incl. `purrdf-gts`) stays `wasm32-unknown-unknown`-clean (`make wasm`).
- Clippy pedantic/nursery warning-free; `make check` + `make doc` green.
- PurRDF mints no vocabulary IRIs; fixtures use `example.org`.

## Scope Decisions (captured user sign-off)
- **Version bump / changelog / `rust-v0.7.x` tag are OUT of this ticket.** The
  user directed (this conversation) that the release is **batched across all
  open tickets** and cut once, later. This is an explicit, authorized descope of
  the issue's "cut a tag" criterion — recorded here, not silently dropped.
  **Consequence (per completion-adversary F2):** issue #142 must **stay OPEN**
  until the batched `rust-v0.7.x` tag ships. The PR therefore references the
  issue with `Refs #142` / `Part of #142`, **never** `Closes #142`. The tag is
  the recorded open blocker; the fix + regression test in this PR is what makes
  the tag, when cut, genuinely close #142.
- **No repro bundle is available** (user confirmed). The triggering fixture is
  *constructed* and proven to panic under a scratch `=0.0.40` pin.
- **Defensive fallback is added now** as permanent hardening (user chose) —
  justified because the panic guard is still compiled in 0.0.49.

## Completeness Contract

Every requirement/criterion maps to a numbered task as an **executable check on a
production surface**. Nothing is deferred, stubbed, or left for a future PR,
except the release-tag criterion explicitly descoped above with user sign-off
(handled as an open blocker keeping #142 open, not a silent omission).

| Explicit 1 — bump structured-zstd past 0.0.40 (already on main) → assert lock resolves ≥ 0.0.49 | Task 3 |
| Explicit 3 — verify the bump against an input that trips the 0.0.40 panic; do NOT assume newer huff0 fixes it → **raw-encoder** replay, non-confounded | Task 2 |
| Explicit 3 (provenance) — the frozen fixture is byte-identical to the input in the recorded 0.0.40 backtrace (proof produced FROM the committed fixture, not a pre-minimization candidate) | Task 2 |
| Explicit 4 — defensive fallback when the entropy encoder bails (encode becomes total) | Task 1 |
| Explicit 2 — cut `rust-v0.7.1` tag | **DESCOPED (user sign-off, batched release); #142 stays OPEN** — see Scope Decisions |
| Implicit 1 — determinism preserved; decoded content-id byte-reproducible; dict conformance passes | Task 1 + Task 3 |
| Implicit 2 — `wasm32-unknown-unknown`-clean (`make wasm`) | Task 3 |
| Implicit 3 — no new features; pin only in `[workspace.dependencies]` | Task 3 |
| AC — previously-panicking bundle composes through `emit_gts` AND the fallback did NOT fire (proves native 0.0.49 encode, not masking) | Task 2 |
| AC — `make check`, `make wasm`, `make doc` green | Task 3 |

- **Extract a shared cross-crate `catch_unwind` totalization util** (parity with
  `crates/rdf/src/native_codecs/parse.rs`). DECLINED: would couple lean,
  pyo3-free, wasm-clean `purrdf-gts` to `purrdf-rdf`, and the two guards degrade
  to *different* types (a diagnostic token vs. a raw zstd frame). Keep one
  in-crate `guarded` combinator in `codec.rs` and cross-reference parse.rs in its
  doc comment.
- **Guard `dict.rs` FastCOVER training with the same harness.** DECLINED for this
  ticket: dictionary *training* is a distinct structured-zstd surface that
  already returns `Result<_, DictError>` and is **not** on the `emit_gts` payload
  path #142 reported. Hardening it is separate work with its own surface and
  tests; folding it here would broaden the blast radius of a targeted fix without
  evidence it is implicated. (Not a deferral of any #142 criterion — a distinct
  capability.)
- **Suppress the panic hook so caught panics print nothing to stderr.** DECLINED:
  suppressing the global hook is a process-wide side effect, unsound in a
  library, and `parse.rs` does not do it (parity). The one stderr line per caught
  panic is documented instead.

---

1. Resolve the top-level repo and derive paths:
   ```bash
   BRANCH="paudley/142-huff0-encode-fallback"
   WORKTREE="$REPO_ROOT/.worktrees/142-huff0-encode-fallback"
   ```
   ```bash
   ```
   ```bash
   cd "$WORKTREE"
   git branch --show-current   # MUST show paudley/142-... NOT main
   ```
4. Warm the build: `cargo check --workspace --locked` (gate run is Task 3).
5. Post this plan to the issue: `gh issue comment 142 --body-file <this-plan-file-path>`.

**⚠️ DO NOT proceed until the worktree exists and you are on the feature branch.**

---

## Task 1: Make the GTS encode path total (defensive fallback + degradation observability)

**Goal:** encoding never panics; on any internal encoder unwind it degrades to a
valid, decode-compatible, deterministic outcome; and *whether* it degraded is
observable from the emitted bytes (silent degradation in a committed,
content-addressed artifact is itself a defect). Mirror the established pattern in
`crates/rdf/src/native_codecs/parse.rs` (`catch_unwind` → named degradation).

**File:** `crates/gts/src/codec.rs`.

1. **Guarded combinator** (single funnel; doc-comment cross-refs parse.rs and
   justifies `AssertUnwindSafe`: the input slice is untouched and the poisoned
   `StreamingEncoder`/buffers are dropped, so no shared mutable state crosses the
   catch boundary):
   ```rust
   /// Run a zstd encode that may panic on an internal encoder invariant
   /// (structured-zstd huff0_encoder.rs "internal error"); on unwind, degrade
   /// per the caller. Deterministic: an input either always encodes or always
   /// degrades. Mirrors crates/rdf/src/native_codecs/parse.rs. On wasm32
   /// (panic=abort) catch_unwind is a compile-clean no-op — see wasm note.
   fn guarded_encode<T>(f: impl FnOnce() -> T, on_panic: impl FnOnce() -> T) -> T {
       match std::panic::catch_unwind(std::panic::AssertUnwindSafe(f)) {
           Ok(v) => v,
           Err(_) => on_panic(),
       }
   }
   ```
2. **Plain / rsyncable paths → raw-literals floor (decode-safe, no dict):**
   - `encode_zstd` (`codec.rs:142`): `guarded_encode(|| compress_to_vec(data, level), || compress_to_vec(data, CompressionLevel::Uncompressed))`.
     `Uncompressed` wraps bytes in a valid Zstandard frame the existing
     `FrameDecoder` round-trips (`codec.rs:105`, unit-tested `codec.rs:306`) and
     cannot itself build a Huffman table, so it can never re-trip the guard.
   - `encode_zstd_rsyncable` (`codec.rs:160`): per-block through the guard.
3. **Dict path → `Err`, NOT a silent degrade (fixes latent corruption hole,
   enhancement #9):** `encode_zstd_with_dict` (`codec.rs:146`) wraps the
   `write_all`/`finish` region in the guard, but on panic returns
   `CodecError::Failed("zstd dict encode aborted (huff0 internal error)")` rather
   than a dict-less frame. Rationale: the GTS header records the dict, the reader
   registers it via `add_dict_from_bytes`, and the existing mismatch test
   (`codec.rs:421`) shows the decoder *refuses* dict mismatches — a dict-less
   fallback frame could be **undecodable**, strictly worse than the panic. `Err`
   is panic-free (totality holds: no unwind escapes) and never emits
   undecodable bytes. (This path is not what #142 hit — large bundles use
   `zstd-rsyncable`, which carries no dict.)
4. **Frame-kind inspector — pure read-side function over bytes** (the mechanism
   the Task 2 non-confound check uses to prove native encode vs raw floor): add
   `pub fn frame_block_kinds(frame: &[u8]) -> Result<Vec<BlockKind>, CodecError>`
   that parses the zstd frame header + block headers (2-bit block type:
   `Raw`/`RLE`/`Compressed`) **without decompressing**. Deterministic, no global
   state, wasm-clean, no wire-format change. Lets a test assert a frame was
   *compressed* vs *raw-degraded*. (Scoped as a test/inspection utility — this
   ticket does NOT wire production-side degradation logging; the encode path
   still degrades without a runtime signal, which is out of #142's scope.)
5. **Unit tests** (`codec.rs` `#[cfg(test)]`), leaning on the existing raw-frame
   round-trip (`codec.rs:306`) and decode grow-loop (`codec.rs:112`):
   - `guarded_encode_falls_back_to_raw_on_injected_panic`: closure panics →
     output decodes (`decode_one`) byte-identically to input; `frame_block_kinds`
     reports `Raw`.
   - `guarded_encode_is_deterministic`: same input twice → identical bytes, both
     success and forced-fallback paths.
   - `guarded_rsyncable_falls_back_per_block`: multi-block payload, one block's
     inner encode panics → whole payload still decodes to the original.
   - `dict_encode_aborts_to_err_not_undecodable_frame`: dict encode with an
     injected panic → `Err(CodecError::Failed)`, and (positive control) a normal
     dict encode + decode-with-dict round-trips.

**Wasm note (honest, per enhancement #10):** `wasm32-unknown-unknown` defaults to
`panic=abort`, so `catch_unwind` **compiles cleanly but does not catch** there —
the guard is a compile-clean no-op on wasm. That is acceptable because the panic
is a **native producer** concern (`emit_gts` composition runs native; verified:
no `panic=abort` in any native profile, so the guard is effective there). `make
wasm` (Task 3) proves compilation only — the plan does **not** claim runtime
catching on wasm.

**Determinism note (per enhancement #12):** the fallback preserves *per-toolchain*
determinism (same input → same panic → same raw frame). The invariant the GTS
writer must keep across dep bumps is the **decoded** snapshot content-id
(`snapshot_content_id`, computed pre-compression), not the compressed bytes
(already version-dependent before this change). Task 3 asserts at that layer.

### After Task 1: commit (github-commit-agent) → push (github-push-agent) →
wiring-verifier → post update to issue #142.

---

## Task 2: Reproduce a real huff0 trigger; verify 0.0.49 fixes it **natively** (non-confounded)

**Goal (issue req 3):** an executable check demonstrating a genuinely-triggering
input (a) panics under `=0.0.40`, and (b) under 0.0.49 encodes **natively without
the guard firing** — so the test would go **red** if 0.0.49 were still broken.
This is the completion-adversary F1 fix: never assert only "`emit_gts` returned
`Ok`", which the Task 1 guard makes tautological.

1. **Construct a genuine trigger, derived from the fault precondition** (enhancement
   #2, de-risks the search vs. a blind sweep). The 0.0.40 guard fires when the
   huff0 literal-weight sum is not a power of two (`huff0_encoder.rs:643`, the
   degeneracy PR #424 stopped feeding to the guard). First check whether
   upstream PR #424 / the structured-zstd test suite ships its own triggering
   input and lift it; otherwise construct from the precondition and, as a
   backstop, sweep the in-repo large corpora `large_non_symmetric_source`
   (`crates/rdf/tests/gts_certify.rs:188`) and `size_comparison_source` (300
   blobs, `crates/rdf/tests/dict_vectors.rs:70`) plus mixed-entropy blocks sized
   past `rsyncable_threshold` (`choose_transform`, `gts_compose.rs:459`).
2. **Freeze/minimize the reproducer FIRST** (completion-adversary F4 — order is
   load-bearing). Reduce the candidate to a small deterministic in-repo fixture (a
   `const`/builder or a byte fixture under the crate's own `tests/` dir — NOT
   under frozen `vectors/`; corpus-frozen-clean; `example.org`-only if it carries
   IRIs). This byte sequence is the artifact the shipped tests run.
3. **RECORD the demonstration on the FINAL frozen fixture (F3 + F4).** In a
   **scratch** throwaway pinned to `structured-zstd = "=0.0.40"`, run **the
   committed frozen fixture itself** (not a pre-minimization candidate) through
   the raw `compress_to_vec` at `DIST_ZSTD_LEVEL = 12` and the rsyncable path.
   **Capture the actual panic output (`huff0_encoder.rs:643 "internal error"`
   backtrace) — which MUST be the backtrace produced by the committed fixture —**
   and paste it plus the exact fixture constructor/seed verbatim into the Task-2
   commit message and PR body. If any minimization step changed a byte, the proof
   is re-run on the final bytes; a backtrace from a superseded candidate is not
   accepted. The scratch pin is then removed (certification, not shipped code);
   the evidence it produced is committed prose+bytes tied to the exact shipped
   fixture, not a discarded claim.
4. **Committed non-confounded verification tests** on the 0.0.49 tree:
   - **(req-3, raw/unguarded — the decisive check)** in `crates/gts/tests/`
     (crate already deps structured-zstd): call `structured_zstd::encoding::
     compress_to_vec(&reproducer, CompressionLevel::Level(12))` and the rsyncable
     block path **directly** (bypassing the guard). No panic ⇒ 0.0.49 fixed it; a
     residual 0.0.49 panic makes the test **red** (not silently caught).
   - **(guard-did-not-fire)** compose/encode the reproducer through the
     production `encode_chain(["zstd-rsyncable"], …)` and assert via
     `frame_block_kinds` that blocks are `Compressed` (proves the native encoder
     succeeded, not the raw floor), then decode back to the original bytes.
   - **(integration, production surface)** in `crates/rdf/tests/`: build a GTS
     bundle whose blob content is/contains the reproducer, compose via `emit_gts`
     with `profile = "dist"`, assert `Ok`, assert its zstd frames are
     `Compressed` (not silently degraded), and **round-trip decode to the
     original graph/blobs** (correctness, not just no-panic).
5. **Class-of-bugs invariant tests** (enhancement #1/#7), converting the one-off
   into a standing guarantee across **both directions**:
   - `encode_chain(["zstd"|"zstd-rsyncable"], x)` **never panics** over a
     generated spread of inputs (encode totality).
   - `decode_chain` **never panics** over arbitrary/truncated bytes — the reader
     is the untrusted-input surface where a structured-zstd internal panic would
     be a real DoS.

**Honesty gate (STOP-and-ask, not an escape hatch):** if the derived-plus-sweep
search does **not** surface a natural `=0.0.40` panic, STOP and report to the
user before opening the PR. Do NOT substitute the injected-panic unit test (which
proves the *fallback contract*, not req-3) for the raw-encoder replay. Options to
put to the user: keep searching / obtain the real gmeow bundle / escalate req-3.
The PR is not cut with req-3 unmet.

### After Task 2: commit (github-commit-agent, message carries the recorded 0.0.40
panic proof) → push (github-push-agent) → wiring-verifier → post update to #142.

---

## Task 3: Verify chokepoint, dependency floor, determinism, and all gates

**Goal:** prove the remaining criteria as executable checks.

1. **Sole-chokepoint (enhancement #5):** an executable grep proving `codec.rs` is
   the only compression site — no `compress_to_vec`/`StreamingEncoder` call
   outside `crates/gts/src/codec.rs` in `crates/gts` (so a future codec added
   straight in `writer.rs` can't bypass the guard). Encode:
   `! grep -rn 'compress_to_vec\|StreamingEncoder' crates/gts/src --include='*.rs' | grep -v 'src/codec.rs'`.
2. **Dependency floor (Explicit 1):** assert `structured-zstd` resolves
   `>= 0.0.49` (`cargo tree -p structured-zstd` / `Cargo.lock` grep); confirm the
   pin lives only in root `[workspace.dependencies]` and no new feature
   (`scripts/check-no-features.py`, `check-versions.py` — both in `make check`).
3. **Determinism (Implicit 1, at the correct layer):** run the GTS codec + dict
   conformance/round-trip suites (`crates/gts/tests/`,
   `crates/rdf/tests/dict_vectors.rs`, `gts_certify.rs`); compose the Task 2
   reproducer twice and assert identical bytes AND identical decoded
   `snapshot_content_id`; confirm FastCOVER seeding still deterministic under
   0.0.49 (`dict.rs` BLAKE3-derived seed, unchanged).
4. **Full gates:** `make check` (fmt + clippy pedantic/nursery + build + tests +
   hygiene: no-features, licenses, corpus-frozen, generated, issue-refs,
   versions + `make wasm`), then `make doc` (rustdoc `-D warnings`, separate from
   `make check`). All green.

### After Task 3: commit any test-only additions (github-commit-agent) → push
(github-push-agent) → post update to issue #142.

---

1. Verify all committed/pushed; run the Mechanical Deferral Scan from
   `_no-deferrals.md` against the branch diff; resolve every hit.
2. Create PR (**`Refs #142`, NOT `Closes` — F2**):
   ```bash
   gh pr create --title "fix(142): make GTS zstd encode total; regression-test huff0 panic fix" --body "Refs #142 (stays OPEN until the batched rust-v0.7.x release tag ships)

   ## Summary
   structured-zstd was already bumped 0.0.40 → 0.0.49 on main, but the huff0
   panic fix was unverified and the encode path had no fault tolerance. This PR
   makes GTS zstd encoding total (catch_unwind → raw-literals floor; dict path
   returns Err rather than an undecodable frame) and adds a NON-CONFOUNDED
   regression test: a fixture proven to panic under =0.0.40 (recorded below)
   now encodes NATIVELY under 0.0.49 (guard does not fire).

   ## Recorded 0.0.40 reproduction
   <pasted panic backtrace from the scratch =0.0.40 run + exact fixture constructor/seed>

   ## Changes
   - codec.rs: guarded_encode on plain/rsyncable (raw floor) + dict path -> Err
   - codec.rs: frame_block_kinds (degradation observable from bytes)
   - regression tests: raw-encoder replay (req-3), guard-did-not-fire, emit_gts
     round-trip; encode+decode never-panic invariant tests

   ## Out of scope (signed off)
   Version bump / changelog / rust-v0.7.x tag are batched across open tickets;
   #142 stays OPEN until that tag ships.
   "
   ```
3. Post plan to PR:
   ```bash
   ```
4. Post to the PR and issue #142 the answers to:
5. Report: "Stage 1 complete. PR #<num> created for issue #142. Ready for /stage2 142".

## Execution Rules
1. Task 0 (Worktree Setup) is MANDATORY — never work on `main`.
2. Commit after EACH task (github-commit-agent) — never batch.
3. Push after EACH commit (github-push-agent).
4. Task N (PR) is MANDATORY and last; PR uses `Refs #142`, never `Closes`.
5. Use sub-agents for all commits/pushes; never run git directly in main context.
6. NO deferrals beyond the release-tag descope signed off above (which keeps #142 OPEN).

paudley:
## Least confident / what you should know

**What I am least confident about:** whether `0.0.49` actually fixes the *specific* huff0 panic `gmeow` hit. I could not reproduce the `=0.0.40` trigger from 25 MB of real repo data (198K windows) or from theory-driven synthetics, so the "fixed" claim rests on upstream PR #424 + the generic `catch_unwind` guard — not an independent reproduction. If the real failing distribution is somehow *also* unfixed in 0.0.49, my guard would silently degrade that frame to raw-literals (correct output, larger) rather than surface it. `frame_block_kinds` exists so a future test with the real bundle can assert native-vs-degraded, but no production code logs degradation, so a silent fallback in the field would be invisible without inspecting the bytes.

**What you may not know but should:**
1. **The panic guard is still compiled into 0.0.49** (`huff0_encoder.rs:746`) — the bump is a *behavioral* fix (which streams reach the guard), not a removal. So the defensive fallback is not redundant belt-and-suspenders; it is the only thing standing between a future edge input and an aborted `emit_gts`.
2. **`catch_unwind` is a no-op on wasm** (`wasm32-unknown-unknown` defaults to `panic=abort`) — the guard only protects the native producer path. That is where `emit_gts` runs, so it is acceptable, but a wasm caller composing bundles would still abort on such a panic.
3. **The dict path degrades differently** — it returns `Err`, not a raw frame, to avoid emitting bytes a dict-registered reader would refuse. So a huff0 panic on a dict-primed frame turns a large-bundle compose into a surfaced error rather than a silent size regression.
4. **The reproduction gap is logged in `.deficiencies`** (git-excluded, local) — if you later obtain the `gmeow` bundle, dropping it through a `=0.0.40` scratch pin would let you freeze the faithful regression test this PR could not build.

paudley:
Closing without merging. The underlying cause (structured-zstd 0.0.40 huff0 encoder panic) is an upstream bug already fixed in 0.0.41 (PR #424); main already pins 0.0.49 (commit 80aebbc). The defensive fallback + new frame_block_kinds API in this branch are added surface for no gain now that the cause is fixed at the source, so they are discarded. Superseded by the repin already on main.

PR #155, opened 2026-07-20
Optimize realtime and whole-bundle SHACL validation

### LINKED ISSUE #153: SHACL engine: parallelize per-focus constraint evaluation for whole-bundle validation
## Problem

`purrdf::shapes::engine::validate_with[_focus_filter]` (`crates/shapes/src/engine.rs`) evaluates
constraints **per focus node serially**. On whole-bundle SHACL — validating every focus node of a
large flattened dataset against the full shape union — this is single-threaded and dominates wall
time. Observed downstream: gmeow-ontology's `bundle-validate` CI job (`make validate-gts` →
`ValidationRun::run` → `validate_dataset` over a ~23M-triple flattened `gmeow.gts`) runs **~50+
minutes**, pegged at one core throughout.

## Why it's serial today

engine.rs (~line 241) documents the deliberate choice:

> Per-focus constraint evaluation stays SERIAL. A rayon `par_iter` over the focus loop was measured
> on `shacl_validate/large_hierarchy` (3000 focus nodes) and REGRESSED ~9% … per-focus work (~5 µs)
> is dwarfed by thread-pool dispatch and shared-`Store` read contention … the parallel path waits on
> the re-entry condition: per-focus cost >50–100 µs.

That benchmark (3000 nodes @ ~5 µs) is not representative of whole-bundle validation: the real
workload is **millions** of focus nodes, and per-focus cost on real shapes (`sh:class` rdfs:subClassOf
BFS + `sh:pattern` regex, and any SHACL-SPARQL) is well above the stated 50–100 µs re-entry threshold.
The code already notes the seam is ready ("The frozen `RdfDataset` is `Sync`, so the seam stays ready").

## Proposal (measure-first)

- Re-benchmark the focus loop parallelization on a **whole-bundle-scale** input (or a synthetic
  million-focus-node dataset), not the 3000-node micro-bench.
- If it wins there (expected), rayon-`par_iter` the per-focus (and/or per-shape) evaluation, keeping:
  - deterministic report output (the existing full-identity `sort_by_cached_key` already normalizes
    order — parallel accumulation + final sort preserves byte-stable reports);
  - the per-run subclass-closure memo correct under concurrency (per-thread memo or a concurrent map);
  - `enter_function_scope` (SHACL-AF function table) valid on worker threads.
- Consider a size threshold so small validations stay serial (avoid the measured 3000-node regression)
  and only large focus sets go parallel.

## Impact

Unblocks bringing gmeow-ontology's whole-bundle conformance gate from ~50 min back to minutes on
multi-core CI, with **no loss of coverage**.

paudley:
Closes #153.

## Summary

- benchmark whole-bundle SHACL execution through the production projected-dataset path at up to one million focus nodes
- compile dataset-bound validation invariants once, including batched asserted-subclass closures
- use a deterministic adaptive Rayon scheduler for sufficiently large focus sets while retaining the serial path for small work
- add `PreparedValidator` for repeated realtime validation against one immutable snapshot, with bounded `Term` and dataset-local `TermId` entry points
- retain the builder's compact store-once term index in frozen datasets so cold preparation does not rebuild a million-term reverse map

Core target membership in bounded calls is answered by direct IR index probes; it does not enumerate the dataset's full target sets. Generic SHACL-SPARQL targets are evaluated once during preparation and retained exactly, avoiding query rewrites that could change `ORDER`, `LIMIT`, aggregate, or subquery semantics. All result paths end in the same total canonical sort, so parallel completion order cannot affect report bytes.

The retained term index is derived and non-serialized, stores only dense `u32` entries, and is never read by byte-producing paths. There are no new Cargo features, optional dependencies, vocabulary defaults, or generated/golden changes.

## Performance

Measured in optimized Criterion runs with 32 Rayon threads:

| Workload | Before | After | Improvement |
|---|---:|---:|---:|
| Whole Core, 1,000,000 focus nodes | 14.143 s | 416.83 ms | 33.9x |
| Whole Core, 100,000 focus nodes | 1.2177 s | 31.24 ms | 39.0x |
| Whole SHACL-SPARQL, 4,096 focus nodes | 1.8923 s | 79.05 ms | 23.9x |
| Realtime compatibility filter, 1 of 1,000,000 | 266.93 ms | 1.15 us prepared | about 232,000x |

On the 1,000,000-focus / 4,000,079-quad realtime fixture, cold preparation fell from about 1.205 s and 3,000,129 allocations / 210 MB transient allocation to about 1.70 ms and 119 allocations / 14.9 KB. Prepared bounded validation measured 10.1 us for 8 focus nodes, 83.9 us for 64, 683.5 us for 512, and 1.55 ms for 4,096.

## Correctness

- bounded term/ID reports match the compatibility-filter report byte-for-byte across `sh:targetClass`, `sh:targetSubjectsOf`, `sh:targetObjectsOf`, `sh:targetNode`, and `sh:SPARQLTarget`
- prepared whole-bundle output matches the existing whole-validation report byte-for-byte
- duplicate candidates, foreign explicit targets, invalid IDs, deterministic parallel output, and worker-thread SHACL-AF function scope are covered
- the prepared value is compile-checked as `Send + Sync`

## Validation

- `make check`
- `make conformance` — 2,876 pass, 10 strictly ledgered xfail/skip, 0 failures
- `make doc` with rustdoc warnings denied
- `make wasm-pkg-size` — 7,771,699 / 8,030,000 bytes; 52,018 SIMD opcodes verified; budget unchanged
- `make wasm-pkg-test` — TypeScript, 60 Node tests, and package/tarball smoke green
- focused 460 `purrdf-core` and 518 `purrdf-shapes` unit tests plus warning-free clippy

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1444,6 +1444,7 @@ dependencies = [
  "purrdf-sparql-algebra",
  "purrdf-sparql-eval",
  "purrdf-xsd",
+ "rayon",
  "regex",
  "serde",
  "serde_json",
--- a/crates/rdf-core/src/dataset_view.rs
+++ b/crates/rdf-core/src/dataset_view.rs
@@ -627,7 +627,7 @@ impl DatasetView for RdfDataset {
 
     #[inline]
     fn term_id_by_value(&self, value: &TermValue) -> Option<TermId> {
-        // Delegate to the inherent lazy reverse value index (no minting).
+        // Delegate to the retained store-once term index (no minting).
         Self::term_id_by_value(self, value)
     }
 
--- a/crates/rdf-core/src/ir/builder.rs
+++ b/crates/rdf-core/src/ir/builder.rs
@@ -976,6 +976,7 @@ impl RdfDatasetBuilder {
             locations.into_boxed_slice(),
             caps,
             named_graphs.into_boxed_slice(),
+            interner.index,
             interner.content_ids,
             derivation_predicate,
         )
--- a/crates/rdf-core/src/ir/dataset.rs
+++ b/crates/rdf-core/src/ir/dataset.rs
@@ -28,6 +28,8 @@ use std::collections::HashMap;
 use std::hash::{Hash, Hasher};
 use std::sync::{Arc, OnceLock};
 
+use hashbrown::HashTable;
+
 use crate::content_id::Blake3ContentId;
 use crate::dataset_view::GraphMatch;
 // Re-exported `pub(crate)` so the sibling `super::dataset::FastHasher` path used by
@@ -45,7 +47,6 @@ use super::term::{BlankScope, InternedTerm, TermId, TermValue, arena_str};
 /// as virtual triples in [`RdfDataset::reifier_quads`].
 const RDF_REIFIES: &str = "http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies";
 
-type ValueIndex = HashMap<u64, Vec<TermId>, FastHasher>;
 /// Lazy successor→predecessors reverse index for
 /// [`RdfDataset::predecessors`]: each successor `TermId` maps to its
 /// predecessor `TermId`s (decoded from the PREDECESSOR-LINK annotation rows),
@@ -218,13 +219,14 @@ pub struct RdfDataset {
     locations: Box<[(QuadHandle, RdfLocation)]>,
     /// Capability flags, computed ONCE at freeze.
     caps: RdfStoreCapabilities,
-    /// Lazy value→id reverse index for [`RdfDataset::term_id_by_value`].
-    /// Keyed by a canonical **hash** of each term's dataset-independent value (with
-    /// `Vec<TermId>` buckets to resolve the rare collision), NOT by full
-    /// [`TermValue`] copies — so building it duplicates **no** term strings (~10×
-    /// leaner than an owned-key map). Built lazily (the builder's interner index is
-    /// dropped at freeze); `OnceLock` keeps the frozen dataset `Send + Sync`.
-    value_index: OnceLock<ValueIndex>,
+    /// Store-once term hash→id index moved intact from the builder at freeze.
+    ///
+    /// Values are only dense `u32` indices into `terms`; equality resolves through
+    /// the arena, so no term strings are duplicated. Retaining the already-built
+    /// table makes every reverse term lookup immediately available without an
+    /// O(term-count) lazy rebuild on a freshly published immutable snapshot.
+    /// Derived, non-serialized, and never consulted by byte-producing paths.
+    term_index: HashTable<u32>,
     /// Lazy permutation quad indexes for indexed
     /// [`quads_for_pattern`](RdfDataset::quads_for_pattern_indexed) (P4b). SPOG
     /// is free (the `quads` table is already freeze-sorted by `(s, p, o, g)`); the
@@ -267,7 +269,7 @@ pub struct RdfDataset {
     /// [`RdfDataset::predecessors`]/[`RdfDataset::predecessor_chain`]. DERIVED,
     /// NON-SERIALIZED: a pure function of the frozen `annotations` table plus
     /// the configured `derivation_predicate`, identical determinism-safety
-    /// rationale to `value_index`/`indexes` — built lazily (`OnceLock` keeps
+    /// rationale to `term_index`/`indexes` — built lazily (`OnceLock` keeps
     /// the frozen dataset `Send + Sync`), never persisted, and read only by
     /// dataset-local `TermId`s (C0.8: never serialized, never read by a
     /// writer).
@@ -535,6 +537,7 @@ impl RdfDataset {
         locations: Box<[(QuadHandle, RdfLocation)]>,
         caps: RdfStoreCapabilities,
         named_graphs: Box<[TermId]>,
+        term_index: HashTable<u32>,
         content_ids: HashMap<TermId, Blake3ContentId, FastHasher>,
         derivation_predicate: Option<TermId>,
     ) -> Self {
@@ -546,7 +549,7 @@ impl RdfDataset {
             annotations,
             locations,
             caps,
-            value_index: OnceLock::new(),
+            term_index,
             indexes: QuadIndexes::default(),
             named_graphs,
             content_ids,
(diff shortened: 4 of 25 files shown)


paudley:
## Stage 2 gap analysis and remediation plan

The first implementation is functionally complete and the broad local gates are green, but adversarial review found remaining gaps that matter for the stated realtime-validation workload. I am treating these as required PR work, not follow-up items.

### Gaps found

1. **Canonical focus ordering still allocates per term.** Four focus-node sort sites render every term into an owned `String`. At million-focus scale this adds avoidable allocation and memory traffic. The replacement must preserve the existing display-byte ordering exactly; changing to enum-derived structural order would alter deterministic output and error ordering.
2. **Recursive constraints discard known interned IDs.** Several recursive branches materialize a `Term` and immediately reverse-resolve it to a `TermId`, adding repeated hash probes on the hot path. They should carry the already-known ID through recursion whenever one exists.
3. **Parallel determinism coverage is narrower than the contract.** The scheduler has ordered reduction, but validation needs forced serial/parallel equivalence over the first-party corpus, multiple worker/chunk geometries, exact error equality, and evidence that the parallel path actually uses more than one worker.
4. **Realtime usage and performance evidence are under-documented.** `PreparedValidator` needs an executable usage example, and the benchmark documentation needs fixture cardinalities, invocation, prepared-vs-compatibility interpretation, and the deterministic scheduling contract.
5. **Touched validation code still contains process-only issue-tracker comments.** These will be replaced with durable technical explanations.

### Remediation sequence

1. Add an allocation-free canonical `Term` comparator that is proven pairwise-equivalent to the current rendered-string order, use it at all four reviewed sort sites, and measure the million-focus path.
2. Reuse `ValueNode` IDs through recursive constraints and validate equivalence with focused engine tests and benchmarks.
3. Add a test-only scheduler control for worker/chunk geometry, then compare exact reports across serial and parallel execution over the complete first-party corpus; test exact error selection and genuine multi-worker participation.
4. Add the prepared-validator example and benchmark/scheduler documentation, and replace process references with durable rationale.
5. Run focused tests and clippy after each logical change, then rerun `make check`, conformance, Python bindings, docs, WASM package gates, and the complete validation benchmark target. Resolve review threads only after their fixes are pushed and verified.

Each logical remediation will be committed, pushed, and reported independently so its evidence remains reviewable.

paudley:
Stage 2 gap 1 is remediated in `4a9ab557` (`Eliminate SHACL canonical sort key allocations`).

- Replaced per-term rendered `String` sort keys with a streaming comparator proven pairwise-equivalent to the existing display-byte contract across IRIs (including prefix edges), blanks, escaped/directional literals, and nested RDF 1.2 triple terms.
- Applied the comparator to all native `Term` canonical sorts, including the four reviewed focus-node sites.
- Added a direct byte-slice IRI fast path and deterministic stable parallel sorting for large focus sets; small realtime requests remain serial to avoid scheduler overhead.
- Optimized 32-thread probes improved the million-focus Core fixture from the prior 416.83 ms to 339.85 ms (about 18.5%) and the prepared 4,096-ID request from 1.551 ms to 1.352 ms (about 12.8%).
- Validation: all 518 library tests, every `purrdf-shapes --all-targets` integration/bench smoke test, the focused comparator suite, formatting, and warning-free all-target clippy pass.

I am resolving the four allocation review threads against this pushed commit, then moving to recursive interned-ID preservation.

paudley:
Stage 2 gap 2 is remediated in `a66d3bb2` (`Preserve interned IDs through recursive SHACL checks`).

- `sh:not`, `sh:and`, `sh:or`, `sh:xone`, `sh:node`, and qualified-value-shape recursion now carry the `ValueNode`'s known `TermId` across materialization.
- Member and sibling loops resolve identity once per value node rather than reverse-hashing the same owned term once per recursive shape.
- Foreign SHACL-AF values retain the existing lookup fallback, so behavior is unchanged when no interned ID exists.
- Validation: all 87 constraint tests, all 21 engine tests, formatting, diff hygiene, and warning-free all-target clippy pass.

Next I am strengthening the parallelism contract across corpus cases, worker counts, chunk geometries, and exact hard-error selection.

paudley:
Stage 2 gap 3 is remediated in `da59adbb` (`Prove deterministic SHACL parallel execution`).

- Added a scoped test seam that forces both serial/parallel selection and exact chunk geometry without changing production behavior.
- Proved byte-identical reports across all 69 first-party corpus cases for serial execution and 2-/4-worker pools with chunk sizes 1, 7, and 64.
- Proved the worker-local SHACL-AF function registry remains byte-identical under 2-/4-worker pools and three chunk geometries.
- Proved the exact earliest hard-error string remains stable across 2-/4-worker pools and chunk sizes 1, 2, 3, 8, and 64.
- Added a barrier-backed test that requires all four configured Rayon workers to participate, preventing a false “parallel” test that only exercises one worker.
- Validation: all 521 library tests pass, including the complete corpus matrix; formatting, diff hygiene, and warning-free all-target clippy pass.

Next I am completing the long-lived validator example and performance/scheduler documentation, while replacing process-only comments with durable technical rationale.

paudley:
Stage 2 documentation/process gap is remediated in `ca7942c3` (`Document realtime SHACL validation operations`).

- Added a compiling `PreparedValidator` example that builds one projected snapshot, prepares once, and validates dataset-local IDs repeatedly.
- Documented all four SHACL benchmark groups, exact cardinalities, targeted commands, prepared-vs-compatibility semantics, and representative realtime/bulk observations.
- Documented the production serial threshold, chunk geometry, ordered reduction, stable large-set sort, and the corpus/thread/chunk/error proof.
- Replaced issue-number/process comments with durable behavior contracts and corrected legacy prose that described PurRDF itself as an ontology.
- Validation: the new doctest passes; all 521 library tests pass; rustdoc with warnings denied, formatting, diff hygiene, process-reference scan, and warning-free all-target clippy pass. The optimized complete `shacl_focus` benchmark smoke target also passes through the 1,000,000-focus fixture.

All identified Stage 2 implementation gaps are now closed. I am moving into the full repository gate, final review-thread/deferral recheck, and hosted-check observation.

paudley:
Optimize realtime and whole-bundle SHACL validation

## Summary

Replace the serial, allocation-heavy SHACL focus loop with a measured validation engine for both latency-sensitive bounded requests and deterministic million-focus whole-bundle work. Frozen datasets now retain a store-once reverse term index, validation preserves interned identities through recursive constraints, large focus sets use bounded Rayon scheduling, and callers can prepare immutable validator state once for repeated realtime requests.

## Requirements satisfied

- Benchmark the actual focus-node execution path at 512 through 1,000,000 focus nodes, including representative SHACL Core and caller-supplied SHACL-SPARQL function work.
- Parallelize sufficiently large focus sets while retaining a serial threshold for small requests and deterministic report/error ordering across worker counts and chunk geometries.
- Keep subclass closure, SHACL-AF function scope, diagnostics, and exact report behavior correct on worker threads.
- Add a `PreparedValidator` surface that resolves selected focus terms to interned IDs once and validates bounded realtime requests without rescanning or rebuilding immutable dataset invariants.
- Remove canonical sort-key strings and recursive reverse term lookups from the hot path while preserving the exact pre-existing lexical report order, including RDF 1.2 quoted triples and directional literals.
- Prove the scheduler uses multiple workers and produces identical reports across the complete first-party SHACL corpus and user-function cases.
- Document workload cardinalities, scheduler semantics, preparation lifetime, realtime usage, allocation counters, and reproducible benchmark commands.

## Files and purpose

- `crates/rdf-core/src/ir/{builder,dataset}.rs` and `dataset_view.rs` build and retain the immutable reverse term index at freeze time and expose constant-time identity lookup without lazy synchronization.
- `crates/shapes/src/parallel.rs` adds deterministic, bounded focus scheduling with a serial small-workload path and test-controlled worker/chunk geometry.
- `crates/shapes/src/engine.rs` adds prepared validation, interned focus execution, streaming canonical comparison, deterministic result/error reduction, worker-scoped SHACL-AF functions, and scheduler/corpus proofs.
- `crates/shapes/src/term.rs` implements allocation-free comparison in the exact canonical term-rendering order and exhaustive pairwise parity tests for IRIs, blank nodes, literals, directionality, escaping, and nested quoted triples.
- `crates/shapes/src/{constraints,components,expression,instance,path,rules,schema_import,shapes,sparql,text_ingest}.rs` and parser modules preserve `TermId` values throughout constraint recursion and adapt all SHACL execution paths to the indexed identity model.
- `crates/shapes/benches/validate.rs` adds fixed-cardinality whole-bundle, SPARQL-function, preparation, and bounded-realtime measurements with allocation/byte reporting.
- `crates/shapes/Cargo.toml`, `crates/shapes/src/lib.rs`, and `Cargo.lock` wire the always-on native parallel implementation and public surface without semantic features or optional behavior.
- `docs/BENCHMARKS.md` records the benchmark contract, commands, representative observations, threshold policy, preparation semantics, and operational guidance for realtime consumers.
- No generated file or frozen conformance vector was edited, and no file was deleted.

## Performance evidence

- The one-million-focus optimized Core probe improved from the original approximately 14.143 seconds to a best observed 339.85 milliseconds; the final complete local smoke remained green under concurrent load at 583.95 milliseconds.
- Preparing the one-million-focus immutable validator snapshot takes approximately 1.35-1.93 milliseconds in observed runs.
- Prepared bounded requests measured approximately 1.88 microseconds for one focus ID, 12.07 microseconds for eight, 87.54 microseconds for 64, 692.45 microseconds for 512, and 2.21 milliseconds for 4,096 in the final complete smoke.
- The 4,096-focus path is above the parallel threshold; smaller realtime requests avoid scheduler overhead and stay serial.
- Criterion remains report-only: timing values are observations, while semantic equality, exact ordering, worker participation, allocation counts, and fixed cardinalities are asserted by tests and harness setup.

## Validation

- `UV_CACHE_DIR=/tmp/uv-cache make check`: green, including formatting, warning-free workspace Clippy, workspace build/tests/doctests, generated and frozen-artifact checks, feature/dependency/license/version hygiene, kernel ring fences, C smoke, and all release crates for `wasm32-unknown-unknown`.
- `UV_CACHE_DIR=/tmp/uv-cache make conformance`: 2,876 passed, 10 ledgered, zero failures.
- `UV_CACHE_DIR=/tmp/uv-cache make pytest`: 390 passed, 1 skipped, 4 strict expected failures.
- `make doc`: green with rustdoc warnings denied; the public `PreparedValidator` example compiles.
- `RAYON_NUM_THREADS=32 cargo bench -p purrdf-shapes --bench validate --locked -- --test`: every optimized SHACL and LinkML benchmark group passed, including the million-focus fixture.
- `make wasm-pkg-size`: 7,794,851 optimized bytes of the unchanged 8,030,000-byte budget; 52,366 SIMD opcodes verified.
- `make wasm-pkg-test`: TypeScript checking, all 60 Node runtime tests, SIMD verification, and packed-install smoke passed; npm tarball and unpacked sizes remain within budget.
- Exact serial/parallel report equality passed for all 69 first-party SHACL corpus cases over 2- and 4-worker pools with chunk sizes 1, 7, and 64, plus SHACL-AF user functions and earliest-error selection.
- GitHub final-head correctness and release checks passed for workspace, full tests, Python, conformance, C API, WebAssembly, docs, MSRV, projection oracles, CodeQL, and CodeRabbit. The explicitly report-only benchmark job remained in progress compiling its harnesses; the complete affected optimized SHACL benchmark target passed locally.
- All four actionable review threads were fixed and resolved; pagination reports no additional review threads. The worktree is clean, its pushed head matches the PR head, and it is zero commits behind `origin/main` before final synchronization.

## Standing constraints

- `.goals` is satisfied: the canonical engine remains Rust-first, RDF 1.2/RDF-star terms are first-class, invalid requests hard-fail, and the public Python/foreign-host model remains downstream of the same implementation.
- Parallel execution is always part of the native SHACL engine rather than a semantic Cargo feature; no optional dependency or caller-dependent semantic build was introduced.
- `purrdf-core` remains free of Oxigraph and PyO3, foundation ring fences remain intact, and all published crates remain wasm-buildable.
- Report bytes and diagnostic selection remain deterministic; no random hasher, clock, RNG, or iteration-order dependency enters an output path.
- No vocabulary IRI or semantic default was minted, no legacy ontology model was retained, and no generated projection was hand-edited.
- The design targets maximal utility, performance, and portability for both realtime validation and large consumer bundles.

## Deferral gate

- The strict scan of all added source/documentation lines and branch commit messages has no TODO, FIXME, stub, placeholder, not-implemented, follow-up, or future-work marker.
- Historical issue and PR comments contain an initial Stage 2 statement naming remaining gaps; the later completion report and four signed remediation commits close every named gap.
- CodeRabbit's phrase `temporary PR review limit` describes its external service quota, not PurRDF implementation work.
- No incomplete implementation or deferred requirement remains.

## Implementation notes

- Canonical comparison streams the same syntax the prior `Term::to_string()` sort keys emitted, avoiding per-term heap strings without silently substituting Rust enum ordering.
- The frozen reverse index is moved from the builder into `RdfDataset`, so repeated focus and recursive constraint checks reuse interned IDs rather than rebuilding or reverse-hashing materialized terms.
- Parallel workers write independent ordered partitions; deterministic reduction restores canonical focus and result order and returns the same earliest error as serial execution.
- SHACL-AF function scope is explicitly entered on every worker, and tests exercise caller-defined functions across scheduler geometries.
- Prepared validation borrows one immutable data/shapes snapshot and is intended for repeated bounded realtime calls; mutation requires constructing a new snapshot.

## Base synchronization and conflicts

Immediately before finalization, `origin/main` was fetched and merged into the issue branch. Git reported `Already up to date`; the branch remained zero commits behind and eight commits ahead, so no merge commit or conflict resolution was required.

PR #165, opened 2026-07-26
feat(gts): support multi-dictionary rsyncable packs

paudley:
## Summary

- support multiple named in-band GTS dictionaries with deterministic per-frame selection
- prime every independent `zstd-rsyncable` block with the selected dictionary while preserving cut points
- expose declared zstd levels, append-to-segment authoring, and deterministic dictionary training
- carry the plan through compaction, composition, Python bindings, and the agent-memory example
- extend the conformance/spec documentation and add dictionary, append, authorship-census, and fold-oracle coverage
- harden append integrity after review: reject ambiguous catalogs and levels, fail unknown revisions closed, and cache append state and segment terms

## Why

The reader already understood the GTS header many-dictionary map, but the writer only accepted one dictionary and could not represent multiple same-codec catalog entries. Rsyncable frames also rejected dictionaries, compaction did not apply one declared transform profile uniformly, and append-oriented callers repeatedly minted headers instead of continuing a segment.

## Impact

Callers can pin several dictionaries in one pack, select one by name for each frame, verify rsyncable block layout and dictionary IDs without decompression, declare the zstd authoring level on the wire, and continue an existing segment without re-emitting its header or dictionaries. Invalid, ambiguous, or contradictory dictionary selections and catalog declarations fail closed.

Vectors 30–33 each carry deterministic `.expected.json` fold oracles derived through the production GTS-to-dataset bridge and N-Quads serializer. The generator and drift guard share one implementation, so corpus generation and validation cannot silently diverge. The release owner explicitly directed that no PR be skipped, so this complete vector change set is retained for the 0.8.5 release.

## Review remediation

- reject zero/oversized zstd frame spans, duplicate catalog names/ids, out-of-range levels, and conflicting append levels
- prove dictionary selection reaches emitted payload frame catalog chains
- reject revisions of absent claims without appending bytes
- cache segment terms and recovered append state across agent-memory writes, refreshing on file-length changes
- document the exclusive single-writer boundary
- add populated `MediumPlan`, malformed-catalog, duplicate-census, and trailing-comment regression coverage
- retain the unconditional vector module because repository policy forbids Cargo feature-gated semantics

## Validation

- `UV_CACHE_DIR=/tmp/uv-cache make check`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p purrdf-gts -p purrdf-rdf --no-deps --locked`
- `cargo test -p purrdf-gts --test multi_dict_writer --test agent_memory_append --locked`
[...]

PR #168, opened 2026-07-28
feat(gts)!: pin caller-supplied in-band dictionary bytes in a compaction plan

paudley:
## What this fixes

`DictPlan` could only name dictionaries the compactor **derived** from the pack's own
content-blob corpus (`DictStrategy::Trained` / `DictStrategy::RawContent`). A consumer that
ships named dictionaries and pins them in a bundle header therefore could not get what it
asked for: compacting "under dictionary X" produced a pack primed with a pack-local
derivation carrying X's **name** but not X's **bytes**.

That is a defect in the id→bytes function, not a missing preference — one dictionary id
resolved to many byte sequences and stopped identifying a decodable dictionary at all.

## The change

`DictStrategy::Pinned(Vec<u8>)` — caller-supplied bytes used verbatim: no training, no corpus
derivation, no truncation. They ride the header `"dct"` map under the plan's name exactly like
`WriterOptions.dicts`, so a reader resolves them the normal way and every primed frame's zstd
header declares that dictionary's own finalized `Dictionary_ID`.

A **variant** rather than a second `dicts_bytes` field on `DictPlan`, so there stays **one**
list of named dictionaries: one name-uniqueness check, one selection-resolution rule, one loop
in `build_pack_dicts`, and mixed plans expressible without a second ordering to reconcile.

### Derivation, not pinning, is what needs a corpus

The blob-less refusal is now a consequence of **derivation**: the corpus is collected only when
some entry derives from it, so a wholly-pinned plan compacts a terms/quads-only log (the
agent-memory shape) with no corpus in sight. `Trained`/`RawContent` plans keep the existing
refusal verbatim.

**Mixed plans (some pinned, some derived) are supported, not refused.** Pinning and deriving are
per-entry questions and neither constrains the other; a mixed plan over a blob-less log still
hits the derived half's corpus refusal.

### Refuse-don't-trust

Every other §10.1 gate is untouched, and one is added for the only input the compactor does not
produce itself: pinned bytes must parse as a finalized zstd dictionary with a non-zero
`Dictionary_ID`, refused **by name** rather than surfacing as a deep codec failure.

## Determinism and frozen vectors

Pinned bytes are an input, and catalog id assignment remains a pure function of the sorted
(codec, dict-name) set. **No frozen vector changed** — `vectors/25b`, `27`, `30`–`33` and their
byte-identity regeneration tests all pass untouched, because the derived path is bit-for-bit as
it was.

## Breaking change

`DictStrategy` loses `Copy` (it now owns bytes) and gains a variant. Every in-tree call site
passes a fresh literal, so nothing else changed here — but this is source-breaking for
downstream exhaustive `match` arms and any use-after-move, and is released as a **minor** bump
per the pre-1.0 policy in `CHANGELOG.md`.

## Coverage

New: `crates/gts/tests/pinned_dict_compaction.rs`, `crates/rdf/tests/pinned_dict_compaction.rs`

- the header `"dct"` entry asserted **byte-for-byte** against the caller's bytes, with a
  derived-plan anti-tautology so the assertion cannot pass vacuously
- every primed frame resolving to the pinned catalog entry and declaring its `Dictionary_ID`
- a blob-less log compacting under a pinned plan and still refusing a derived one
- byte-identical output across two in-process runs
- round-trip at RDFC-1.0 canonical-graph identity plus full `verify_compaction`
- the mixed-plan ruling in both directions
- unusable pinned bytes refused

## Also in this PR

`docs(gts):` — `frozen_vector_pins_no_pack_dictionary` attributed the 25b vector's absent `"dct"`
header to `DictStrategy::None` in three places (a comment and two assertion messages). No such
variant has ever existed; the generator compacts with `DictPlan::undicted()`. Prose only, no
assertion logic touched.

## Summary by CodeRabbit

* **New Features**
  * Added support for pinned compression dictionaries, preserving caller-provided dictionary bytes verbatim in compacted pack headers.
  * Supports plans mixing pinned and derived dictionaries, including blob-less inputs.
* **Bug Fixes**
  * Validates pinned dictionary bytes as finalized zstd dictionaries; rejects empty/non-dictionary/non-finalized content and enforces clearer failure when derived content blobs are required but missing.
* **Documentation**
  * Updated compact streamable guidance with the new derived-vs-pinned sourcing and failure modes.
* **Tests**
  * Added/extended integration tests for pinned and mixed dictionary compaction across GTS and RDF, including fold/digest and byte-preservation checks.
* **Chores**
  * Adjusted changelog templating to reliably preserve breaking-change markers.

### DIFF
--- a/cliff.toml
+++ b/cliff.toml
@@ -11,6 +11,11 @@
 # Exact rendered scope/message pairs are emitted once per release group. Multiple
 # implementation commits can legitimately share a conventional subject, but one
 # repeated user-facing release-note bullet carries no additional information.
+#
+# A conventional `type(scope)!:` subject or a `BREAKING CHANGE:` footer renders a
+# leading **BREAKING** marker on its bullet. Pre-1.0 the suite ships breaking
+# changes under a minor bump, so the changelog is the only place a consumer can
+# see that a bump carries one — it must never be silent.
 
 [changelog]
 header = """
@@ -37,7 +42,7 @@ body = """
 {% set entry_key = entry_scope ~ "::" ~ entry_message -%}
 {% if entry_key not in rendered_entries -%}
 {% set_global rendered_entries = rendered_entries | concat(with=entry_key) -%}
-- {% if commit.scope %}**{{ commit.scope }}:** {% endif %}{{ commit.message | split(pat="\n") | first | upper_first | trim }}
+- {% if commit.breaking %}**BREAKING** {% endif %}{% if commit.scope %}**{{ commit.scope }}:** {% endif %}{{ commit.message | split(pat="\n") | first | upper_first | trim }}
 {% endif -%}
 {% endfor %}
 
@@ -80,7 +85,9 @@ commit_parsers = [
   { message = ".*", group = "Other" },
 ]
 
-protect_breaking_commits = false
+# A breaking commit is never dropped by a `skip = true` parser: pre-1.0 a minor
+# bump MAY carry breaking changes, so the marker must always reach the changelog.
+protect_breaking_commits = true
 filter_commits = false
 # Only the canonical crates.io release tags mark releases; the parallel
 # `py-v*` / `npm-v*` tags carry the same version and are not release boundaries.
--- a/crates/gts/src/compact.rs
+++ b/crates/gts/src/compact.rs
@@ -36,25 +36,54 @@ const XSD_DATETIME: &str = "http://www.w3.org/2001/XMLSchema#dateTime";
 
 /// The conventional in-band pack dictionary name for a single-dictionary plan.
 pub const DEFAULT_DICT_NAME: &str = "pack";
-/// Target size of a pinned pack dictionary. FastCOVER/raw-content truncate the
-/// content to fit; a fixed value keeps compaction byte-deterministic.
+/// Target size of a DERIVED in-band pack dictionary. FastCOVER/raw-content
+/// truncate the corpus to fit; a fixed value keeps compaction
+/// byte-deterministic. [`DictStrategy::Pinned`] bytes are never truncated — they
+/// are the caller's, verbatim.
 const DICT_TARGET_LEN: usize = 16 * 1024;
 
-/// How one named in-band dictionary's bytes are derived from the content-blob
-/// corpus (§8.5 `dct`).
+/// Where one named in-band dictionary's bytes come from (§8.5 `dct`).
 ///
-/// Both are byte-deterministic; the trained strategy is the production default
-/// (Req 3 headline is "trained"), and the raw-content strategy is a named
-/// alternate. The choice affects only the pinned dictionary bytes, never the
-/// fold — a reader decodes the in-band dictionary either way. "No dictionary"
-/// is NOT a strategy: it is an empty [`DictPlan::dicts`], so a plan can never
-/// name a dictionary that produces nothing.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+/// [`Self::Trained`] and [`Self::RawContent`] DERIVE the bytes from the pack's
+/// own content-blob corpus. [`Self::Pinned`] carries caller-supplied bytes used
+/// VERBATIM. All three end up pinned in the header `"dct"` map — the variant
+/// names where the bytes come from, not whether they are stored in-band.
+///
+/// The derived strategies are byte-deterministic functions of the corpus: the
+/// trained one is the production default (Req 3 headline is "trained"), the
+/// raw-content one a named alternate. [`Self::Pinned`] exists because a derived
+/// dictionary cannot honour an EXTERNAL dictionary id: a consumer that ships
+/// named dictionaries and asks for a pack "under dictionary X" must get X's
+/// actual bytes, not a pack-local re-derivation wearing X's name — otherwise one
+/// id resolves to many byte sequences and stops identifying a decodable
+/// dictionary at all.
+///
+/// The choice affects only the dictionary bytes, never the fold — a reader
+/// decodes the in-band dictionary either way. "No dictionary" is NOT a strategy:
+/// it is an empty [`DictPlan::dicts`], so a plan can never name a dictionary
+/// that produces nothing.
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub enum DictStrategy {
-    /// FastCOVER-trained dictionary (the production default).
+    /// FastCOVER-trained over the content-blob corpus (the production default).
     Trained,
-    /// Raw-content dictionary (canonical trailing window of the corpus).
+    /// Raw-content over the content-blob corpus (its canonical trailing window).
     RawContent,
+    /// Caller-supplied finalized zstd dictionary bytes, used verbatim: no
+    /// training, no corpus derivation, no truncation. These exact bytes ride the
+    /// header `"dct"` map under the plan's name, so the pinned name resolves to
+    /// exactly the dictionary the caller shipped.
+    Pinned(Vec<u8>),
+}
+
+impl DictStrategy {
+    /// Whether this strategy DERIVES its bytes from the pack's content-blob
+    /// corpus — and therefore requires the pack to have one.
+    fn derives_from_corpus(&self) -> bool {
+        match self {
+            Self::Trained | Self::RawContent => true,
+            Self::Pinned(_) => false,
+        }
+    }
 }
 
 /// Which in-band dictionary primes a frame — TOTAL, never `Option`.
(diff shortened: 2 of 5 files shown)


paudley:
Pin caller-supplied in-band dictionary bytes in a GTS compaction plan, so a
consumer that ships named dictionaries gets ITS bytes under ITS name instead of
a pack-local re-derivation wearing that name.

BREAKING CHANGE: `purrdf_gts::compact::DictStrategy` gains a `Pinned(Vec<u8>)`
variant and no longer derives `Copy`. Downstream exhaustive `match` arms over
`DictStrategy` and any use-after-move of a `DictStrategy` value must be updated.
The enum is deliberately not `#[non_exhaustive]`, matching the workspace's
documented convention for exhaustive public enums.

## What this fixes

`DictPlan` could only name dictionaries the compactor DERIVED from the pack's
own content-blob corpus (`Trained`/`RawContent`). Compacting "under dictionary
X" therefore produced a pack primed with a derivation carrying X's NAME but not
X's BYTES — one dictionary id resolving to many byte sequences, which stops
identifying a decodable dictionary. That is a defect in the id->bytes function,
not a missing preference.

`DictStrategy::Pinned(Vec<u8>)` uses the caller's bytes verbatim: no training,
no corpus derivation, no truncation. They ride the header `"dct"` map under the
plan's name exactly like `WriterOptions.dicts`, so a reader resolves them the
normal way and every primed frame declares that dictionary's finalized
`Dictionary_ID`.

A variant rather than a second `dicts_bytes` field keeps ONE list of named
dictionaries: one name-uniqueness check, one selection-resolution rule, one loop
in `build_pack_dicts`, and mixed plans expressible without a second ordering to
reconcile.

Derivation, not pinning, is what needs a corpus. The corpus is collected only
when some entry derives from it, so a wholly-pinned plan compacts a
terms/quads-only log (the agent-memory shape) with no corpus in sight.
`Trained`/`RawContent` keep the existing refusal verbatim. Mixed plans are
SUPPORTED, not refused; a mixed plan over a blob-less log still refuses for its
derived half.

## Files

- `crates/gts/src/compact.rs` — the `Pinned` variant, `derives_from_corpus`,
  corpus collection made conditional, and a refuse-don't-trust parse gate on the
  one input the compactor does not produce itself.
- `crates/gts/tests/pinned_dict_compaction.rs` (new) — 10 tests.
- `crates/rdf/tests/pinned_dict_compaction.rs` (new) — 3 tests.
- `crates/rdf/tests/streamable_vectors.rs` — prose fix: three references named
  `DictStrategy::None`, a variant that has never existed. The generator compacts
  with `DictPlan::undicted()`. Assertion logic untouched.
- `cliff.toml` — breaking changes now render a `**BREAKING**` marker.

## Standing constraints (.goals / AGENTS.md) — respected

- No Cargo features added. No dependency pinned in a member manifest.
- No frozen vector changed: `git diff origin/main...HEAD -- vectors/` is empty;
  `check-corpus-frozen.py` OK; `vectors/25b, 27, 30-33` byte-identity
  regeneration tests pass untouched, because the derived path is bit-for-bit as
  it was.
- Byte determinism preserved — pinned bytes are an input; catalog id assignment
  remains a pure function of the sorted (codec, dict-name) set. Verified across
  separate processes and across plan orderings.
- wasm32 clean (`make wasm` inside `make check`). Kernel ring-fence intact.
- Test fixtures use `example.org`; no vocabulary IRI is minted.

## Completion-Adversary: PASS (pass 2/3)

Pass 1 returned FINDINGS-OPEN with three findings. Each was independently
confirmed rather than accepted on report, and each was fixed by real work — none
was discharged by rewording, by filing a follow-up, or by arguing scope.

[F1] BRAND/FIXTURE VIOLATION — the new fixtures minted `gmeow.example` IRIs and
a `"gmeow-bundle-v1"` dictionary name, against the AGENTS.md rule that fixtures
use `example.org` and that the dependency arrow never points from purrdf to the
GMEOW ontology. Demonstrated novel: zero gmeow references existed in
`crates/gts/tests` on origin/main.
  DISCHARGED: `git diff origin/main...HEAD | grep -i gmeow` returns nothing; all
  12 fixture IRIs are `example.org`.

  Collateral, and the most valuable thing this audit produced: the rename BROKE
  `a_mixed_plan_pins_the_callers_bytes_and_derives_the_rest`, which asserted the
  header keys as `["gmeow-bundle-v1", "pack-local"]` and passed only because `g`
  sorts before `p`. The assertion now sorts its expectation and says plainly
  what it can observe, and a new test
  `the_order_dictionaries_appear_in_a_plan_does_not_change_the_bytes` asserts
  the real invariant byte-exactly. The adversary MUTATION-TESTED that new test:
  making catalog-id assignment plan-order-dependent (`&BTreeSet<&str>` -> `&[&str]`
  fed from insertion order) fails it while all nine other tests in the file still
  pass — so it is the only guard on the invariant, i.e. genuine new coverage.
  Mutation reverted; worktree verified clean.

[F2] DARK / UNREACHABLE GUARD — `DictPlan::validate` refused a pinned dictionary
declaring `Dictionary_ID` 0 with its own message, but that branch can never
execute: `dict::dictionary_id` decodes through `Dictionary::decode_dict`, which
already rejects a zero id unconditionally (`structured-zstd` dictionary.rs:157),
so it never returns `Ok(0)`.
  DISCHARGED: branch deleted, non-zero guarantee attributed to the parse where it
  actually lives. Refusal behaviour re-demonstrated unchanged on the production
  API and still names the offending dictionary — patching a real dictionary's id
  bytes to zero yields "Dictionary id must be non-zero" from the parse.

[F3] BREAKING CHANGE WAS INVISIBLE — the change is source-breaking but nothing
could have said so: no `!` subject, no `BREAKING CHANGE:` footer, and
`cliff.toml` had NO breaking-change rendering at all, so a `!` alone would have
produced a byte-identical changelog and a satisfied-looking checkbox.
  DISCHARGED, all parts verified by execution: PR retitled `feat(gts)!:` (ghprsq
  builds the squash subject from the PR title, so the marker reaches the commit);
  the bullet template emits `**BREAKING**` from `commit.breaking`;
  `protect_breaking_commits = true` so a breaking commit can never be dropped by
  a skip parser. Proven inert for published releases: the FULL repository history
  rendered at `--tag rust-v0.8.5` with old and new `cliff.toml` diffs to nothing.

Pass 2 verdict: PASS. No new findings against the remediation's own text. No
STOP-AND-ASK items. All pass-1 production-surface demonstrations re-run against
the new head and reproduced unchanged (verbatim bytes at 8 KiB / 32 KiB /
200 KiB with no truncation past `DICT_TARGET_LEN`; all 66 transformed frames
resolving to the pinned catalog entry and declaring its `Dictionary_ID`;
`verify_compaction` all_ok; blob-less pinned compaction alongside the unchanged
derived refusal; cross-process byte determinism).

One post-PASS change was made and is disclosed rather than hidden: the adversary
noted the mixed-plan assertion MESSAGE said "ordered by name", a property that
assertion structurally cannot observe because `header_dct` funnels entries
through a `BTreeMap`. It adjudicated this "a message imprecision, not a coverage
gap. No work is owed." It was fixed anyway, per the standing rule that a defect
you can see is a defect you own. The change is a comment and an assert message
only — no assertion logic, no behaviour — and `make check` and `make doc` were
re-run green on the final tree.

## Deferral gate: PASS

Mechanical deferral scan over the full branch diff: CLEAN (zero hits) on every
pass. Boundary-laundering scan: CLEAN. No follow-up issues were filed. No
`.deficiencies` entry was added — the file's mtime remains 2026-07-14, fourteen
days before this session, verified across both audit passes.

## Human decisions on the record

The version for the release that carries this break was put to the maintainer as
a blocking question BEFORE implementation, since crates.io/PyPI/npm publishes are
irreversible and CHANGELOG.md states "pre-1.0, a minor bump may carry breaking
changes and a patch bump is bugfix-only".

  Question: "DictStrategy gains a Pinned(Vec<u8>) variant and loses Copy — a
  breaking change to a public enum in purrdf-gts. Which version should I cut?"
  Decided-by: maintainer, verbatim selection "0.9.0 — minor (Recommended)",
  chosen over the offered alternative "0.8.6 — patch".

The maintainer also approved folding the `DictStrategy::None` prose fix into this
PR ("Fix it in this PR (Recommended)") and the branch rename off
`paudley/gts-multi-dict-rsyncable`, which was still bound to already-merged PR
#165 and made `gh pr view` resolution ambiguous.

## Gates

`cargo fmt --all` clean; `make check` exit 0 (fmt, clippy -D warnings across the
workspace, build, 150 test suites with 0 failures, no-features, licenses,
corpus-frozen, generated-drift, issue-refs, versions, rdf-core hygiene, wasm);
`make doc` exit 0. No merge conflicts — the branch was never behind origin/main.

## Notes for future work

GTS compaction is not exposed through the CLI, Python, or wasm surfaces. That is
pre-existing and symmetric — `Trained`/`RawContent` are equally unexposed — and
is recorded here as an observation about the crate's reach, not as anything this
change altered or owes.

PR #177, opened 2026-08-02
perf(rdf): cut allocations on the hot paths; fix(canon)!: collision-safe RDF 1.2 canonicalization profile

paudley:
Eleven commits. Nine on allocations in the hot paths, API evolution and the serializer's memory
shape; two closing **issue 176** — a canonicalization identity-collision that blocks a downstream
consumer's semantic identity layer.

## Allocations

**The join's binding frame.** `SlotSolution::bindings` was a `Vec`, and `extend_atom` clones the whole
solution once per surviving candidate row — an allocation per intermediate solution, paid
unconditionally, on the crate that never received the structural pass `purrdf-sparql-eval` gave its
own rows. The inline capacity is **measured, not assumed**: instrumenting across this workspace's
suites and the whole vendored OWL 2 RL corpus put the widest frame any rule produces at **eight**
variables. `Id` is a `NonZeroU32`, so `Option<TermId>` is 4 bytes through the niche and the frame is
32 inline bytes — less than a `Vec` header spends before addressing any element.

`sources` is deliberately left a `Vec`, and that is a measurement too: `SourceRow` is 32 bytes, the
body-length distribution over **36,014 compiled rules** is 72% at two atoms or fewer, and fan-out is
worst for the short bodies. Inlining eight source rows would put 256 bytes on every intermediate of
exactly the rules that make the most of them.

**Rejecting before copying.** `bind_row` copied the solution row and *then* asked whether the
candidate binds. The two rejections needing no state now happen first, as a strict subset of what the
binding loop rejects. It earns its place on the RDF 1.2 reification path, which this function's own
docs note is not index-pre-filtered.

## API evolution

**`#[non_exhaustive]` on 41 enums.** 0.11.0 had to be a minor bump *specifically* because four public
enums gained variants. Those four plus 37 public `*Error` types now carry it. Closed vocabularies
deliberately do not — `Regime`'s seven inhabitants are fixed by specifications, and a downstream
matching them exhaustively *should* stop compiling if a spec adds an eighth.

The workspace's own compile-time discipline is untouched: `#[non_exhaustive]` binds every crate except
the defining one, and `UndecidedReason`'s `Display` is ten arms with no wildcard in its own crate.
Total blast radius across 20 crates was **five** wildcard arms, each taking the outcome true by
construction — the XSD classifier answers `Unmodelled` rather than `IllTyped` because only the former
is a claim about *this crate* rather than about the caller's ontology.

**Error chains.** 52 `impl Error` bodies were empty. Six types that demonstrably wrap a cause now
chain, so `{:#}`, `anyhow` and any `source()` loop reach the real failure instead of stopping at the
wrapper. Non-wrapping variants return `None` deliberately — `EntailError::Inconsistent` carries a
*witness*, not a failure, produced by a run that succeeded.

## The serializer's memory shape

The text family allocated **~6 `String`s per quad**, held them all in a `Vec<String>`, then `join`ed
(copying the whole document) and `format!`ed (copying it again). Turtle wrapped that in a third copy.

Terms now append through `write_term`, so the `Cow` that `escape_iri`/`escape_literal` already returned
reaches the buffer with no intermediate. `RdfCodec::serialize_into(&SerGraph, &mut String)` is the
required method and `serialize` is provided from it, so the two can never disagree. The four text
formats reach the sink with nothing in between; the other four append a document they still build
whole, and each says why — RDF/XML, TriX, JSON-LD and YAML-LD assemble a *tree*, so their writers
cannot emit a prefix before knowing what follows.

`sort_canonical` no longer builds a `Vec<String>` key per row. It sorts through a comparator rendering
into two reused buffers: `O(n log n)` renders against `O(1)` scratch instead of `O(n)` keys.

## Determinism

The emitted order is a published property — it is what makes a document byte-identical across
`DatasetView` backends. The comparator is asserted **equivalent to the historical key construction**,
not merely valid, on a graph exercising rows that differ only in the graph slot. Mutation-tested:
rendering `None` as anything non-empty fails it.

That test also caught a wrong claim of mine on its first run — I had asserted the default graph leads
the *document*, when the graph slot is the last key component and so separates rows only once
subject, predicate and object have tied.

## Verification

250/250 W3C syntax round-trip vectors · RDFC-1.0 green over 65 vectors · SPARQL 800/5 · OWL 2 RL 50/50
· conformance TOTAL 3308/14/14/0 · `make check` · `make doc` · workspace clippy at `-D warnings`.

`purrdf-datalog` also gains its first bench: the evaluator had none, and the nearest measured it
through a regime's rule table. Three groups separate a join's fan-out cost from its per-solution cost,
because one figure mixing them cannot say which half a change moved. Report-only — no speedup is
asserted anywhere in this branch.

## Breaking changes

* 41 enums gain `#[non_exhaustive]`; downstream matches need a wildcard arm. Paid once, so no variant
  added after it is breaking again.
* `RdfCodec` implementors provide `serialize_into` rather than `serialize` (crate-internal trait).

---

# Canonicalization: collision-safe sentinels, a named profile, a frozen corpus

Closes the reported identity-collision in RDF 1.2 canonicalization.

## The defect

The canonicalizer lowers the RDF 1.2 reifier and annotation overlays onto RDFC-1.0 using sentinel
IRIs, and the module asserted these were IRIs "that no real dataset IRI can occupy". **Nothing made
that true.** `urn:purrdf:rdfc:reifies` is a legal IRI a dataset may assert as an ordinary predicate,
and a dataset that did so canonicalized to the *same bytes* as a genuine reifier structure.

For a consumer minting identity from canonical bytes, two structurally different datasets sharing a
digest is an identity-forgery primitive. The disjointness the overlay's losslessness rests on had to
hold by construction.

## The fix: refusal, at the namespace

Any IRI in `urn:purrdf:rdfc:`, in **any** position, is now refused.

Stated over the **namespace** rather than the two sentinel spellings, and that is load-bearing: an
enumeration must be re-audited every time the overlay grows a row, and that audit is exactly the step
that gets skipped. The sweep covers nested triple terms and literal datatype slots — the datatype
slot is safe today, and a rule with a carve-out for whichever position happens to be harmless is one
nobody can audit.

Refusal over injective escaping (as the consumer requested): the property to audit becomes a total
rule over the input rather than a proof about an escaping function.

Two things had to be made **total**, not merely correct:

* **The diagnostic.** The reported violation is the least `(position, iri)`, not the first
  encountered — first-encountered means interning order, which differs between backends holding the
  same dataset.
* **The typing, all the way up.** Three call sites did `map_err(|_| …BudgetExceeded)`, which compiled
  fine and would have reported a forgery attempt as a graph that ran out of budget. That is the
  subtler half of this bug: discarding an error's value compiles forever and silently re-labels every
  *new* failure mode as an old one.

## The profile

`docs/RDF12-CANON-PROFILE.md` specifies the algorithm under the stable name **`purrdf-rdfc12` v1**.
Writing the spec surfaced something the code had never stated: a named-graph annotation emits **five**
tokens before the `.`, so the output is a canonical **byte string**, not re-parseable N-Quads. A
consumer feeding it back to a parser would have found that out the hard way.

`CANON_PROFILE_ID` / `CANON_PROFILE_VERSION` / `RDFC_CALL_LIMIT` are public so a pin can be checked
against the build actually linked, not just against a document.

## The corpus

`vectors/rdf12-canon/` — 17 goldens and 9 refusals, self-describing via `manifest.tsv` and
`pairs.tsv` so the other side can run it without reading our test code. Refusal expectations pin the
**position**, not just the fact of refusal.

`CANON_CORPUS_DIGEST` is the SHA-256 of the corpus freeze manifest, reproducible with one
`sha256sum` and none of our code — a digest only its author can compute is not one anybody can check.

Goldens are generated from the implementation, so they are evidence of **stability, not
correctness** — the README and harness say so plainly. Correctness evidence is the W3C suite for the
RDF 1.1 subset and the property relations for the overlay.

## Verified falsifiable

| Mutation | Caught by |
|---|---|
| sweep disabled | 6 tests |
| namespace narrowed to the two sentinel names | 2 tests |
| one line appended to a golden | corpus harness **and** freeze gate, independently |

## Breaking changes

* `try_canonicalize`/`try_canonicalize_with` return `CanonError`, not `BudgetExceeded`.
* `PackError::CanonBudgetExceeded` / `CertifyError::CanonBudgetExceeded` → `CanonRefused(CanonError)`.
* `PackError` is no longer `Copy` (a refusal that names an IRI cannot be). Ripple: zero call sites.
* `canonicalize`/`canonicalize_with` panic on reserved-vocabulary input as they already did on budget
  exhaustion; both now carry a `# Panics` contract directing untrusted input to `try_canonicalize`.

## Summary by CodeRabbit

* **New Features**
  * Added the RDF 1.2 canonicalization profile, including support for annotations, reifiers, and nested triple terms.
  * Added reserved-vocabulary protection, typed refusal diagnostics, profile metadata, and expanded public access to canonicalization APIs.
  * Added append-based RDF serialization for more efficient output generation.

* **Bug Fixes**
  * Improved propagation of canonicalization and underlying error details.
  * Prevented unknown future error conditions from being silently misclassified.

* **Documentation & Tests**
  * Added profile documentation, conformance coverage, and frozen validation vectors.


paudley:
Implemented on PR #177, landing in **0.12.0**. All four acceptance criteria are closed; the pin target is the tagged `v0.12.0` release.

## 1. Injection rejection — done, and wider than asked

You asked for the two sentinels. The rule is over the **whole namespace** `urn:purrdf:rdfc:` instead, exported as `purrdf_core::RESERVED_NAMESPACE`. An enumeration has to be re-audited every time the overlay grows a row, and that audit is exactly the step that gets skipped; a namespace rule is one sentence checkable against the whole module, and it costs nothing to widen because no vocabulary is published there. Your parenthetical — "the same question applies to any other reserved vocabulary the overlay introduces" — is answered structurally rather than by us remembering.

"Any position" covers subject, predicate, object, graph, **nested inside a triple term at any depth**, and **a literal's datatype IRI**. The datatype slot is safe today; it is swept anyway, because a rule with a carve-out for whichever position happens to be harmless is one you cannot audit.

Refusal, not escaping — your reasoning, adopted: the property you audit becomes a total rule over the input rather than a proof about an escaping function.

Two things we had to make total rather than merely correct:

- **The diagnostic.** The violation reported is the least `(position, iri)`, not the first encountered. First-encountered means statement order, statement order is interning order, and two backends holding the same dataset need not agree on it. Two conforming implementations must now reject the same datasets *and* name the same violation — otherwise your refusal receipts aren't comparable.
- **The typing, all the way up.** Three call sites did `map_err(|_| …BudgetExceeded)`, which compiled fine and would have reported a forgery attempt as a graph that ran out of budget. `PackError` and `CertifyError` now carry the typed `CanonError`.

`purrdf_core::check_admissible` applies the predicate alone, so you can screen at **admission** rather than only when identity is minted — the "and ideally dataset admission" half of your ask.

## 2. Named, versioned profile — `purrdf-rdfc12` v1

`docs/RDF12-CANON-PROFILE.md`. Algorithm, overlay lowering rules row by row, blank-node handling across the overlay, reserved vocabulary, refusal rule and its determinism, bounds.

`CANON_PROFILE_ID`, `CANON_PROFILE_VERSION` and `RDFC_CALL_LIMIT` are **public constants**, so you can assert the build you linked is the profile you pinned rather than trusting a document. §7 states what increments the version and — as importantly — what does not.

**One thing you will want to know before you bind this.** Writing the spec surfaced something the code had never stated: a named-graph annotation emits **five** tokens before the terminating `.`. That is deliberate (a genuine quad never carries two graph tokens, which is what keeps the row from colliding with one), but it means the canonical form is **an opaque byte string to compare or digest, not a document to feed back to an N-Quads parser**. On the RDF 1.1 subset the output *is* valid canonical N-Quads. Profile §3.1.

Also worth binding explicitly: **the hash algorithm is part of the identity.** SHA-384 output is not comparable to SHA-256 output, so a pin naming the profile but not the algorithm is incomplete.

## 3. Explicitly not RDFC-1.0 — stated in both directions

Profile §1. It accepts *less* (reserved vocabulary refused) and emits *different bytes* (RDFC-1.0 has no reifiers to emit). On the RDF 1.1 subset the two agree byte for byte, and that agreement is gated by the vendored W3C `rdf-canon` suite rather than claimed.

## 4. Bounded fallible entry points — kept, and the bound is now contractual

`RDFC_CALL_LIMIT` = 1 000 000, public. No knob, fixed by the build, so two parties on the same profile version refuse the same datasets. The §5 sweep runs **before** any hashing: a dataset that is both inadmissible and pathologically symmetric is refused for the reason that makes it dangerous, without spending the budget to discover it.

## Vector corpus

`vectors/rdf12-canon/` — 17 goldens, 9 refusals. Self-describing via `manifest.tsv` (input, kind, expectation) and `pairs.tsv` (relations *between* cases), so you can run it without reading our test code. Frozen: every payload byte is covered by a SHA-256 manifest checked on every build.

Refusal expectations pin the **position**, not just the fact of refusal — §5.3 makes the diagnostic deterministic, and an untested guarantee is not one.

`poison-forgery` is your attack, built end to end: it asserts as an ordinary quad exactly the row `reifier-simple` lowers to. Its test checks **both** halves — that the genuine structure still produces that row, and that the literal assertion is refused — so it cannot keep passing by quietly ceasing to be a forgery.

**Corpus identity:** `purrdf_core::CANON_CORPUS_DIGEST` is the SHA-256 of the corpus freeze manifest, reproducible with `sha256sum scripts/conformance-frozen/vectors-rdf12-canon.sha256` and none of our code — a digest only its author can compute is not one you can check. The harness asserts the constant against the file, so the corpus cannot move without the pin moving in the same commit.

**One caveat, stated plainly rather than left for you to infer.** The goldens are *generated from the implementation*. They are evidence of **stability, not correctness** — they cannot tell you the algorithm is right, only that it will not move under you. Correctness evidence is the W3C suite for the RDF 1.1 subset and the property relations for the overlay. The README and harness both say so. If you want independent correctness evidence for the overlay specifically, a second implementation running this corpus is the way to get it, and vector candidates from your side are very welcome.

## Verified falsifiable

| Mutation | Caught by |
|---|---|
| sweep disabled | 6 tests |
| namespace narrowed to the two sentinel names | 2 tests |
| one line appended to a golden | corpus harness **and** freeze gate, independently |

Scoreboard: `RDF 1.2 canonicalization profile | purrdf-rdfc12 v1 (first-party) | 5 | 0 | 0 | 0 | GREEN`, labelled first-party beside the vendored-W3C rows because the distinction matters to what it proves.

## What to pin

| Field | Value |
|---|---|
| profile id | `purrdf-rdfc12` (`CANON_PROFILE_ID`) |
| profile version | `1` (`CANON_PROFILE_VERSION`) |
| hash algorithm | `CanonHash::Sha256` (or `Sha384` — not interchangeable) |
| corpus digest | `038f7431e845e63c8bb2122cdfa2c9968f40c17ae7cb6b9e458bbb5cb11375b7` (`CANON_CORPUS_DIGEST`) |
| release | `v0.12.0` |

Happy to have ADR 0007's validation list reviewed against profile §§1–8 — if a clause you need to bind isn't stated there, that is a gap in the profile and we'll close it before the tag.

PR #200, opened 2026-08-07
Prefer GCI conclusions during DL branching

arne309:
## Summary

- internalize complex GCIs with the superclass alternative before the negated subclass
- keep the test-only GCI builder aligned with the RDF parser
- add an RDF-level regression covering inverse roles, a universal restriction, and exact cardinality in both OWL-Direct decision cores

## Root cause

A complex inclusion `sub ⊑ sup` is internalized as the logically equivalent disjunction `¬sub ⊔ sup`. Both decision cores preserve the disjunct order and explore branches depth-first. For the regression ontology, the negated complex subclass expands into existential and cardinality alternatives, so trying it first creates an exponential search tree and reaches the 136,864-step cap.

Putting `sup` first does not change the calculus or model set; it changes only the deterministic branch heuristic so the conclusion of the inclusion is tried before manufacturing evidence that its premise is false.

Before this change, both cores exhausted the step cap. After it, the hypertableau decides consistency in 19 steps and the reference tableau in 12.

## Validation

- `cargo fmt --check`
- `cargo test -p purrdf-entail`

The full entailment suite passes, including the oracle, OWL-RL corpus, reasoner, and doctests.

### DIFF
--- a/crates/entail/src/owl_dl/mod.rs
+++ b/crates/entail/src/owl_dl/mod.rs
@@ -324,7 +324,7 @@ impl Kb {
         if matches!(sub, Concept::Named(_)) {
             self.unfold.entry(sub_id).or_default().push(sup_id);
         } else {
-            let meta = Concept::Or(vec![Concept::Not(Box::new(sub)), sup]);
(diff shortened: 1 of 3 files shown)


arne309:
Closed in favor of #202

PR #201, opened 2026-08-07
feat(sparql)!: add a property-function registry

### LINKED ISSUE #185: Property-function (magic predicate) registry
# Property-function (magic predicate) registry

## Summary

This is the keystone seam of the four identified in the umbrella thesis: a
property-function (a.k.a. "magic predicate") registry, absent from PurRDF today.
Every mainstream engine surveyed (Jena, GraphDB, QLever, Virtuoso) uses this
mechanism — caller-registered, multi-valued relations keyed by IRI and invoked
as if they were triple patterns — to deliver full-text search, GeoSPARQL query
rewrite, path binding, and multi-functions. None of the surveyed engines
dereference the function IRI at query time; registries are keyed by
caller-supplied IRIs, consistent with PurRDF's mint-nothing rule.

Delivering this seam requires: a new algebra node for property-function
invocation, evaluator dispatch, variable substitution through the node,
integration with the scope walker, a governor charge point
(`governor/charge_points.rs`) so property-function evaluation cost is
accounted like any other operator, and a soundness classification
(`governor/soundness.rs`) since property functions can be non-monotonic or
side-effecting relative to the rest of the algebra.

This item unlocks items 11 (full-text search), 12 (path binding), 13 (embedding
kNN), and the property-function-based query-rewrite half of item 15 (GeoSPARQL).

## Source pointers

- `crates/sparql-eval/src/user_fn.rs` `UserFunctionRegistry` — the existing
  scalar-function registry pattern this seam extends to relations.
- `crates/sparql-algebra/src/algebra.rs:891-899` — `PurrdfCall`, the
  caller-supplied-IRI precedent for this registry's keys.
- `governor/charge_points.rs` — governor charge-point integration.
- `governor/soundness.rs` — soundness classification.

## Acceptance sketch

- A property-function algebra node exists with evaluator dispatch, correct
  variable substitution, and scope-walker integration.
- A minimal reference property function (e.g. a deterministic in-memory
  relation) round-trips through parse → algebra → eval → results.
- Every property-function invocation has a governor charge point and a
  soundness classification; unregistered IRIs error, not silently no-op.
- `cargo test -p purrdf-sparql-eval` passes, including a property-function
  end-to-end test and a governor-charge test for property-function evaluation.

Effort: M.

paudley:
Closes #185

## Summary

Adds the SPARQL property-function ("magic predicate") seam: caller-registered, multi-valued relations keyed by IRI, invoked as triple patterns (`?doc ex:contains ("needle" ?score)`). Designed as *query answering over relations with limited access patterns*: a relation declares its access capabilities (adornments over the in-tree `BindingPattern` lattice) and the engine chooses a feasible execution order — capability declarations are satisfied by a planner, never turned into mid-evaluation failures for queries that have a feasible order.

## Changes

- **`BindingPattern` hoisted** from `purrdf-datalog` into `purrdf-core` (old path re-exported; one adornment lattice across the magic-sets rewriter and this seam).
- **Algebra**: `GraphPattern::PropertyFunction(PropertyFunctionCall)` — byte-exact predicate IRI, structural n-ary subject/object argument vectors (no cons-cell desugaring; `()` = zero-length vector; nested collections hard-error). `ParserOptions::property_fn_namespaces` (default empty = seam off; parses stay bit-identical, asserted). Calls emit as a textual-order left-deep LATERAL chain over the residual BGP; serializer round-trips; refused in CONSTRUCT/update templates.
- **Relation contract**: object-safe `PropertyFunction` cursor trait over `TermValue` — declared arity, access modes (feasibility = subsumption: `fb` serves `bb` by generate-then-filter), per-mode row bound, `open(args, ceiling)`/`next()` pull cursor (the engine stops after k rows without the relation's cooperation), deterministic emission-order contract, deaf-relation doctrine, panic containment. `PropertyFunctionRegistry` panics on duplicate IRI (silent shadowing is a wrong-answer channel); `describe()` is the self-describing-data channel. `MemoryRelation` reference implementation, buildable from an `rdf:List` of `rdf:List`s.
- **Planning**: prepare-time greedy feasibility ordering (certainly-bound analysis that never over-claims; ties by declared row bound → IRI → textual position). Unregistered IRIs, arity mismatches, and infeasible chains are refused at admission, before the first governor charge. Plan-cache keys carry the namespace set + a registry fingerprint. Registered IRIs derive the parser namespaces at a single point, so registry and parse configuration cannot drift.
- **Evaluation**: per-input-row dispatch with the row in hand (any term kind binds — literals included), structural row unification (bound positions filter; repeated variables and blank slots enforce equality; blanks projected away), rows ingested through a governed core shared with SERVICE (cell-ceiling observe before intern, stop-poll between rows, certified truncation). A call inside a SERVICE body is refused at the forwarding boundary. SHACL constraint bodies invoke registered relations. Parallel-safe only for `Volatility::Stable` relations, with minted-row re-interning under forked workers pinned by a minting determinism-gate corpus case.
- **Governors (profile v5)**: `property-function-invocation` + `property-function-row` charge points; admission control prices a call from its declared `rows_per_invocation`; frozen corpus gains a relation lane (12 cases) whose deaf/cooperating cancellation pair is byte-identical — the per-row admission point is itself a stop checkpoint, so the seam is bounded per row.
- **Entailment lane**: witness restriction over a call is the identity (its rows come from the registry; its argument variables are observable, so the data leaves binding them are already restricted). Quoted-triple variables made visible to witness restriction, closing a pre-existing under-restriction.
- **Surfaces**: explain receipt names the registered relations (new registry-carrying explain entry); conformance suite `purrdf-property-functions` (14 manifest-driven cases incl. mode-restriction, reordering, and negative refusal cases); report-only bench case; narrowness docs updated.

## Breaking changes

- `GraphPattern` gains the `PropertyFunction` variant; `ParserOptions` gains `property_fn_namespaces` (prefix-matched, caller-declared) and `property_fn_iris` (exact-match, registry-derived) — exhaustive matches and struct literals must be updated.
- `ShaclQueryOptions` is renamed `QueryOptions` (prebinding + user functions + property functions + mint prefix) and is a required parameter of every governed query entry and of `update_governed`; `query_governed_in_operation_with_options` is folded into `query_governed_in_operation`; `query_shacl_view` is renamed `query_with_options_view`; `PreparedQuery` is no longer struct-literal constructible.
- `GOVERNOR_PROFILE_VERSION` 4 → 5; `GOVERNOR_PROFILE_DIGEST` is now `1e8e8f2b340bf85addaa55e3e4b323ca75325ad9465b9846cf96ac13b557317e`; `GOVERNOR_CORPUS_DIGEST` re-pinned (41 cases / 34 bands with per-charge-point `.charges` records). A fuel budget sized against v4 must be re-measured under `QueryGovernors::METERED`.
- `QueryExplanation::relations()` returns `&[PfDescriptor]`; `GovernedOutcome::Complete`/`BudgetExhausted` gain a `relations: RelationIdentity` field; `PropertyFunctionRegistry::describe` and `PreparedQuery::rewritten` return `Result`; rendered `EvalError::Function` text is prefixed `host function error`.

## Verification

Full workspace gates on the final branch state: `make check` green (4716 tests / 0 failures; workspace clippy pedantic/nursery `-D warnings` clean), `make conformance` GREEN (3442 pass / 14 ledgered / 0 fail across 15 suites; SPARQL scoreboard 814 incl. the 14 property-function suite cases; Python binding suite 588 incl. the 35 relation-registration cases), `make wasm` green, `make doc` green (denied warnings). Governor corpus 41/41 cases, 34 bands (11 harness tests) with per-charge-point `.charges` records; frozen-corpus check 7/7 byte-frozen with pre-existing files byte-identical; benches compile (report-only, no speedup asserted). The stage-2 completion adversary's final verdict is PASS, every acceptance criterion demonstrated in production from an out-of-tree consumer.

## Summary by CodeRabbit

* **New Features**
  * Added configurable SPARQL property functions backed by host-provided relations.
  * Supports multi-argument, chained, filtered, correlated, ordered, and serializable calls.
  * Added planning, explanations, caching, and registration/execution APIs.
  * Added property-function support to SHACL-SPARQL evaluation and parallel processing.
  * Added governor accounting for invocations and emitted rows.

* **Bug Fixes**
  * Improved reasoning, pre-binding, graph analysis, and quoted-triple handling.
  * Ensured consistent sequential and parallel behavior.
  * Rejects unsupported calls in remote SERVICE queries and CONSTRUCT templates.

* **Documentation**
  * Updated conformance results and governor documentation.


paudley:
# Implementation Plan for Issue #185 — Property-function (magic predicate) registry

## Issue Summary

Add a SPARQL property-function ("magic predicate") seam: caller-registered,
multi-valued relations keyed by IRI, invoked as if they were triple patterns
(e.g. `?doc ex:contains ("needle" ?score)`). The issue requires: a new
algebra node; evaluator dispatch; variable substitution through the node;
scope-walker integration; a governor charge point; a soundness
classification; no IRI dereference at query time (registry keyed by
caller-supplied IRIs, mint-nothing); a minimal reference property function
(deterministic in-memory relation) round-tripping parse → algebra → eval →
results; unregistered IRIs error rather than silently no-op; `cargo test -p
purrdf-sparql-eval` passes including an end-to-end test and a governor-charge
test.

Downstream consumers this seam is shaped for (issues that declare a
dependency on this one): full-text search (index-specific caller IRIs, ≥2
output variables per row, out-of-core sibling crate, wasm-clean); path
binding (many ordered rows per input row, multi-output tuple); embedding kNN
(relation-controlled deterministic order, governor cost proportional to
work, top-k row ceilings).

**Magnitude note:** the enhancement audit's transformational finding is
adopted (see Design), which grows this from the issue's M estimate toward
M-L: prepare-time feasibility ordering, the `BindingPattern` hoist, a shared
governed row-ingest core, and frozen governor-corpus cases are in scope.

Adversarial review posture; no deferrals/follow-ups — work is NOW; no GitHub
issue/PR numbers or process references in repo files (gated by
`scripts/check-issue-refs.py`); bugs are owned on sight.

GREENFIELD-FIRST (no compromises); RUST-FIRST, PYTHON-SURFACE; SUBSUME,
EXTEND, ENHANCE; LOW/NO OPTIONALITY, HARD FAILS; RDF 1.2 ONLY; MAXIMAL
UTILITY / PERFORMANCE / PORTABILITY (wasm32-unknown-unknown-clean). Repo
contract: no Cargo features; workspace-root dependency pinning only;
deterministic evaluation; no minted vocabulary IRIs (`example.org`
fixtures); perf claims need a criterion bench, never an asserted speedup.

## Design (resolved; grounded in code facts and the enhancement audit)

1. **The concept is a relation with limited access patterns** (foreign
   relations / access-pattern-constrained query answering), not a bag of
   magic predicates. A relation DECLARES its access capabilities and the
   ENGINE chooses a feasible execution order — declarations are satisfied by
   a chooser, never turned into mid-evaluation failures for queries that
   have a feasible order.
2. **Adornments via the in-tree lattice.** `BindingPattern`
   (`crates/datalog/src/binding_pattern.rs` — subsumes/meet/join with a
   proved lattice-law test suite) is HOISTED into `purrdf-core` and
   re-exported from `purrdf-datalog` (one definition across the magic-sets
   rewriter and this seam — SUBSUME/EXTEND). Feasibility is SUBSUMPTION, not
   membership: an invocation is feasible iff some declared mode requires no
   more bound positions than the invocation supplies (a relation declaring
   `fb` serves `bb` by generate-then-filter).
3. **Recognition is parse-time; resolution and ordering are prepare-time.**
   `ParserOptions::property_fn_namespaces: Vec<String>` (default EMPTY =
   seam off — this is the mandated caller-supplied-vocabulary rule, not
   optionality), mirroring `extension_fn_namespaces`
   (`crates/sparql-algebra/src/parser.rs:70-75`). The parser emits nodes in
   TEXTUAL order inside the group's `Lateral` chain. At prepare time (the
   engine/`PlanCache` boundary, where the registry is available) a
   deterministic, statistics-free rewrite orders each maximal chain
   greedily by feasibility (BGP atoms always feasible; property-function
   atoms only in a subsumed mode; ties broken by declared
   `rows_per_invocation`, then lexicographically). Admission-time HARD
   ERRORS, raised before the first governor charge: unregistered IRI
   (naming it); no feasible total order (naming the atom, its free
   positions, and its declared modes); arity mismatch. `PlanCache` keys
   include the property-function namespace set and registry identity.
4. **N-ary structural arguments.** For a recognized property-function
   predicate the parser captures subject/object STRUCTURALLY —
   `subject_args: Vec<TermPattern>` / `object_args: Vec<TermPattern>` (a
   non-list term is a 1-vector; a collection is its element vector; NO
   `parse_collection` cons-cell desugaring, `parser.rs:1560`). Argument
   grammar rules (each documented on the variant and tested): blank nodes
   are non-distinguished variables (the `BLANK_VAR_PREFIX` synthetic-slot
   treatment from `bgp.rs`, projected away); repeated variables within or
   across sides are equality-constrained (post-invocation consistency
   filter); the empty collection `()` denotes a 0-length argument vector
   (an explicit bare `rdf:nil` term is a 1-vector — syntactically
   distinguishable); a NESTED collection inside an argument list is a hard
   parse error naming the position (no defined semantics — fail loud);
   quoted triples are ordinary `TermPattern::Triple` arguments whose
   variables participate in the adornment.
5. **New `GraphPattern::PropertyFunction(PropertyFunctionCall)` variant, NO
   `#[non_exhaustive]`.** `PropertyFunctionCall { iri: String, subject_args,
   object_args }`; `iri` retains the author's byte-exact spelling (the
   `PurrdfCall` precedent, `algebra.rs:891-899`). The wildcard-free
   exhaustive matching across the workspace is protective; the variant
   addition ships as `feat(sparql)!`.
6. **The relation is a CURSOR, not a push callback.** Object-safe,
   `TermValue`-space, `Send + Sync` trait:
   - `volatility() -> Volatility` (reuse `user_fn::Volatility`).
   - `arity() -> PfArity` (declared subject/object arities — checked at
     prepare time; a mis-shaped call never reaches the host, the
     `user_fn::Arity` doctrine).
   - `modes() -> &[BindingPattern]` (declared access patterns).
   - `rows_per_invocation(mode: BindingPattern) -> u64` — an upper-bound
     cardinality class with TWO existing consumers: the admission-time
     `PlanEstimate`/`ChargeLedger::for_plan` peak-cells check (a relation
     emitting millions of rows per row must be refusable at admission, not
     mid-flight) and the prepare-time order tie-break.
   - `open(&self, args: &PfArgs<'_>, ceiling: Option<u64>) ->
     Result<Box<dyn PfCursor>, EvalError>`; `PfCursor::next(&mut self) ->
     Result<Option<PfRow>, EvalError>`. Pull model: the engine stops after
     k rows without the relation's cooperation, charges per row
     structurally, and can hand a licensed cap-pushdown ceiling to the
     relation (a kNN index told "10 rows matter" does 10 rows of work).
     Emission order is the relation's contract and MUST be deterministic.
   - Deaf-relation doctrine documented on the trait verbatim from the
     `remote.rs` analysis: the stop signal is polled before `open` and
     between `next` calls and work is charged before it is consumed, so a
     non-cooperating relation degrades to per-call granularity, forfeiting
     the positional-prefix claim, never soundness.
   - Host panics contained with `catch_unwind(AssertUnwindSafe(..))` and a
     fixed payload-free message (the `NativeFnBody` treatment,
     `user_fn.rs:559`).
   - NO dataset-read façade. None of the three named consumers needs one
     (FTS/kNN read their own out-of-core index; path binding is its own
     design conversation), and an ungoverned read surface would be an
     unmetered-scan hole. Dropped deliberately (see Declined).
7. **Registry.** `PropertyFunctionRegistry`
   (`DetHashMap<String, Arc<dyn PropertyFunction>>`), caller-IRI-keyed,
   `register` HARD-FAILS on duplicate IRI — a deliberate STRENGTHENING of
   `UserFunctionRegistry`'s same-kind last-write-wins (its panic precedent
   at `user_fn.rs:296` is cross-kind only; silent shadowing of a relation
   is a wrong-answer channel, so we refuse it). `resolve`, `is_empty`,
   `len`, and `describe() -> Vec<PfDescriptor>` (IRI, arities, modes as
   lattice codes, volatility, cardinality class; sorted by IRI) — the
   self-describing-data channel callers emit FnO/SHACL from, with no `fno`
   coupling. Registered IRIs (sorted) are named in the query
   explanation/receipt: registry contents change answers, so a governed
   receipt that omits them is not a receipt.
8. **Evaluation binds arguments from the input row DIRECTLY.** The
   per-row `Lateral` substitution helper is IRI-only
   (`expr.rs:1229` — literals deliberately left to join fallback), which a
   property-function node lacks; relying on it would deliver FTS
   needles/kNN vectors as UNBOUND. The dispatch instead computes each
   argument position's value from the current input row via the node's arg
   vectors + schema. Bound OUTPUT positions act as equality filters
   (magic-predicate test mode).
9. **Shared governed row-ingest core.** Extract from `remote.rs` the
   sequence (charge → cell-ceiling observe BEFORE interning → scratch
   intern with promote-to-`Existing` → stop-poll → truncation certificate,
   completes-vs-abandons distinguished) into one core used by BOTH
   `eval_service` and the property-function operator. Labels stay distinct
   (`remote-*` vs `property-function-*`): network work and in-process index
   work must be tellable apart; the mechanism is shared, the vocabulary is
   not. The existing `service-*` frozen vectors pin the shared core from
   the SERVICE side.
10. **Governors.** Two charge points — `property-function-invocation` (after
    the refusal checks, the `UserFunctionInvocation` placement doctrine) and
    `property-function-row` — `CHARGE_SCHEDULE` 9 → 11, profile v4 → v5
    (version, digest, `docs/SPARQL-GOVERNOR-PROFILE.md` v5 section).
    `NodeCharges.fuel: [u64; CHARGE_SCHEDULE.len()]` widens — audit every
    recorded-ledger shape. NEW FROZEN CORPUS CASES via the sanctioned
    generator (never hand-edited): zero / exact / one-below ceilings for
    both points, governed-parallel-vs-sequential trip parity,
    below-first-row-cost deterministic truncation, and a deaf-relation
    stop case — an unpinned charge point is not a contract.
11. **Soundness: the node is a LEAF.** It has no child `GraphPattern`, so it
    belongs with `Bgp`/`Path`/`Values` in `visit_pattern_parts` (no
    `ChildRole` edge — `Opaque` would be a category error). No static
    `child_row_ceiling` (a relation may emit arbitrarily many rows; the
    per-row charge bounds it — never a faked ceiling). `plan_cap_pushdown`
    already records a licensed ceiling AT the node when
    `SpineClass`/`OrderCertainty` permit; the dispatch consumes it via the
    same `row_ceiling`/`cell_row_ceiling` path `eval_bgp` uses and passes
    it into `open`.
12. **Parallel.** `is_parallel_safe_pattern`: safe only when the resolved
    relation declares `Volatility::Stable`; unresolvable or `Volatile` ⇒
    UNSAFE (sequential) — the "when in doubt, UNSAFE" doctrine. A `Stable`
    relation minting terms inside a forked worker goes through the
    existing `portable_row`/`reintern_minted_row` fold-back in source
    order; the determinism-gate corpus case MUST mint terms not present in
    the dataset (a case emitting only existing IRIs would pass while
    proving nothing).
13. **Re-parse hazards (round-trip is `ParserOptions`-dependent).** A
    serialized node re-parsed without the namespace silently becomes
    cons-cell triples. Therefore: forwarding a property-function node
    inside `SERVICE { }` (`pattern_to_select_query` in the remote path) is
    a HARD ERROR at the forwarding boundary; the SHACL body parse
    (`crates/shapes/src/sparql.rs`) carries the same namespace
    configuration alongside the registry RAII scope (a shape body using a
    property function must invoke it, tested — never degrade to an
    always-passing constraint); `PlanCache` keys include the namespace set
    (per Design 3).
14. **Context plumbing.** `EvalCtx.property_functions:
    Option<&'d PropertyFunctionRegistry>` + `with_property_functions`,
    mirroring `user_functions` exactly (`eval.rs:383/1250`) — uniformity
    with the sibling seam; `None ≡ Some(empty)` behavior pinned by a test.
    Propagated through `fork_for_worker` and `child_for_user_fn`; engine
    builder + query entry points; FOURTH field on `ShaclQueryOptions`
    wired ONLY through `apply_shacl_query_options` (`engine.rs:1378`);
    `CURRENT_PROPERTY_FUNCTIONS` RAII scope beside `CURRENT_FUNCTIONS`
    (`crates/shapes/src/sparql.rs:464`).
15. **EXISTS correlation.** `pattern_expr_vars` (`expr.rs:806`) reports the
    node's argument variables (they are invocation inputs, NOT
    join-constrained positions — a correlated EXISTS containing a property
    function must not take the uncorrelated memoized path).

## Completeness Contract

Every requirement and acceptance criterion from the issue maps to a numbered

| Registry: caller-registered multi-valued relations keyed by IRI | 3 |
| New algebra node for property-function invocation | 2 |
| Evaluator dispatch for the node | 4 |
| Variable substitution through the node (both substitution passes) | 4 |
| Scope-walker integration + EXISTS correlation reporting | 2, 4 |
| Governor charge points; cost accounted like any other operator | 5 |
| Soundness classification (leaf; ceiling-opaque; cap-pushdown consumed) | 5 |
| No IRI dereference at query time; caller keys; mint-nothing | 2, 3 |
| Byte-exact IRI round-trip on serialization | 2 |
| Minimal reference property function (deterministic in-memory relation) | 3 |
| Round-trips parse → algebra → eval → results | 6 |
| Unregistered IRIs error (at admission, before the first charge) | 4, 6 |
| `cargo test -p purrdf-sparql-eval` passes | 6 |
| End-to-end property-function test | 6 |
| Governor-charge test for property-function evaluation | 5, 6 |
| Feasibility-ordered execution (declared modes satisfied by the chooser) | 4 |
| `BindingPattern` hoist; subsumption feasibility | 1 |
| Cursor contract; ceiling pushdown into the relation | 3, 4, 5 |
| Admission cardinality estimate (`rows_per_invocation` → `PlanEstimate`) | 5 |
| Shared governed row-ingest core with SERVICE | 4 |
| Frozen governor-corpus cases for both new charge points | 5 |
| Parallel classification, minted-row re-interning, determinism gate | 4, 6 |
| Re-parse hazards closed (SERVICE forward error; SHACL body config; PlanCache key) | 4 |
| Argument-grammar rules (blanks/repeats/empty/nested/quoted) documented + tested | 2, 6 |
| Registry `describe()` + registered IRIs in the explanation/receipt | 3, 6 |
| Conformance suite (manifest-driven via `MemoryRelation::from_graph`) | 6 |
| Criterion bench case (report-only) | 6 |
| Workspace-wide compile + wasm checks for touched release crates | 6 |

- **`PfDataView` dataset-read façade** — no named consumer among the three
  downstream issues, and an ungoverned read trait would be an unmetered-scan
  hole defeating every ceiling; a future consumer that genuinely needs graph
  reads designs a GOVERNED one with full information. Declined as
  producer-without-consumer, not deferred work.
- **CLI/Python/wasm/C-API registration surfaces** — a property function is a
  Rust trait object; it cannot be expressed from CLI flags, a Python
  callable conflicts with the GIL-released query path and `Send + Sync`,
  and all named consumers are Rust sibling crates using the Rust API.
- **FnO catalog integration** — `describe()` provides the self-describing
  data; callers emit FnO from it (the `slice::list_functions_catalog`
  precedent) without coupling `sparql-eval` to `rdf-core::fno`.
- **`#[non_exhaustive]` on `GraphPattern`** — defeats the protective
  wildcard-free exhaustive matching; versioned as breaking instead.
- **Default namespaces / built-in relations / strict-lenient switches /
  caller-settable per-relation row ceilings** — each violates mint-nothing
  or LOW/NO OPTIONALITY; the fuel budget plus licensed cap-pushdown is the
  sanctioned bounding channel.
- **Semiring/provenance-annotated relations** — scores arrive as ordinary
  bound output variables; no consumer needs annotation propagation; the
  access-pattern formalism is the one that pays off here.

1. ```bash
   BRANCH="paudley/185-property-functions"
   WORKTREE="$REPO_ROOT/.worktrees/185-property-functions"
   cd "$WORKTREE" && git branch --show-current   # MUST show the feature branch
   ```
2. Setup: `cargo check -p purrdf-sparql-algebra -p purrdf-sparql-eval -p
   purrdf-datalog` (allowed setup only; kill anything past 10 minutes).
   NEVER run `make check`/`make conformance`/workspace-wide suites in stage
   1. Pre-existing origin/main failures: one line, not chased.
3. Post this plan: `gh issue comment 185 --body-file <this-plan-file>`.

## Task 1: Hoist `BindingPattern` into `purrdf-core`

- Move `BindingPattern` (+ its lattice-law test suite) from
  `crates/datalog/src/binding_pattern.rs` into `purrdf-core`; re-export
  from `purrdf-datalog` at the old path (zero churn for datalog callers).
  Pure move — no behavior change; the lattice tests prove it.
- Executable checks: `cargo test -p purrdf-core binding_pattern`,
  `cargo test -p purrdf-datalog` green; `cargo check -p purrdf-core
  --target wasm32-unknown-unknown` clean.
- Commit (`refactor(core):` — no `!`, old path re-exported) → push → issue
  update → wiring-verifier.

## Task 2: Algebra node, parser, serializer, scope walker (`purrdf-sparql-algebra`)

- `ParserOptions::property_fn_namespaces` (default empty; first-match-wins
  prefix semantics identical to `extension_fn_namespaces`).
- `GraphPattern::PropertyFunction(PropertyFunctionCall)` with byte-exact
  `iri`; parser recognition in the triples-block path emitting TEXTUAL
  order in the group's `Lateral` chain; structural argument capture with
  ALL Design-4 grammar rules (blanks, repeats, `()`, nested-collection
  parse error, quoted triples); a variable predicate is never a property
  function.
- Serializer: byte-exact `<iri>` + collection syntax; parse → serialize →
  parse round-trip tests, including the seam-off case (no configured
  namespace ⇒ identical-to-today parse, asserted byte-identical
  serialization).
- Scope walker (`visible_variables`/`collect_vars`): all argument variables
  visible in the group; rule documented on the variant.
- Executable checks: `cargo test -p purrdf-sparql-algebra` green with the
  new round-trip, grammar-rule, and scope tests.
- Commit (`feat(sparql)!:` breaking footer: new variant) → push → issue
  update → wiring-verifier.

## Task 3: Trait, registry, reference relation, plumbing (`purrdf-sparql-eval`)

- `PropertyFunction` trait per Design 6 (cursor, volatility, `PfArity`,
  `modes() -> &[BindingPattern]`, `rows_per_invocation`, deaf-relation
  doctrine in the docs); `PfArgs`/`PfRow` in `TermValue` space.
- `PropertyFunctionRegistry` per Design 7 (duplicate-IRI hard-fail as a
  documented strengthening; `describe()`/`PfDescriptor`).
- `MemoryRelation`: public deterministic in-memory relation (caller-supplied
  tuples; all modes by scan; stable insertion-order emission; honest
  `rows_per_invocation`), plus `MemoryRelation::from_graph(dataset, head)`
  reading tuples from an `rdf:List`-of-lists (manifest-driven conformance
  cases and a copy-paste starting point for consumers).
- Plumbing per Design 14 (EvalCtx mirror of `user_functions` with
  `None ≡ Some(empty)` pinned; engine builder + entry points;
  `ShaclQueryOptions` fourth field via `apply_shacl_query_options` only;
  `CURRENT_PROPERTY_FUNCTIONS` RAII scope; fork propagation).
- Executable checks: `cargo test -p purrdf-sparql-eval` unit suite green
  (registration, duplicate hard-fail, resolve, describe, MemoryRelation
  over every mode and from_graph); `cargo check -p purrdf-sparql-eval
  --target wasm32-unknown-unknown` clean.
- Commit (`feat(sparql):`) → push → issue update → wiring-verifier.

## Task 4: Prepare-time feasibility ordering + evaluator dispatch

- Prepare-time rewrite per Design 3 (greedy feasible order via
  `BindingPattern` SUBSUMPTION; deterministic and statistics-free; admission
  hard errors for unregistered IRI / no-feasible-order / arity mismatch,
  each with a precise message, all raised before the first charge;
  `PlanCache` key extension).
- Dispatch per Designs 8-9: per-input-row argument binding computed
  DIRECTLY from the row (never via the IRI-only `substitute_pattern`);
  invocation-mode computation; `open(args, ceiling)` with the licensed
  cap-pushdown ceiling; shared governed row-ingest core extracted from
  `remote.rs` and used by both SERVICE and this operator (existing
  `service-*` frozen vectors must stay byte-identical — that is the
  refactor's regression net); bound-output equality filters; repeated-var
  consistency; blank-node synthetic slots projected away; panic
  containment; `stop_check` between rows.
- Substitution: `substitute.rs::substitute_in_graph_pattern` AND
  `expr.rs:1066 substitute_pattern` arms over the arg vectors;
  `pattern_expr_vars` (`expr.rs:806`) reports arg variables (EXISTS
  correlation); `syntactic_schema` (`eval.rs:1704`).
- Parallel: classification per Design 12; minted-row
  `portable_row`/`reintern_minted_row` path; SERVICE-forwarding hard error
  per Design 13; SHACL body namespace configuration + invocation test.
- Executable checks: end-to-end eval tests over `MemoryRelation` covering
  every adornment, literal-argument binding (the FTS-needle shape:
  `?doc ex:contains ("needle" ?score)` with a sibling pattern on `?doc` —
  must invoke in a bound mode chosen by the orderer, never enumerate),
  bound-output filtering, multi-row ordered output, admission errors,
  seam-off inertness, correlated-EXISTS, SHACL-body invocation, and a
  `parallel_determinism_gate` corpus case whose relation MINTS terms.
- Commit (`feat(sparql):`) → push → issue update → wiring-verifier.

## Task 5: Governors — charge points, profile v5, soundness, admission estimate, frozen corpus

- Schedule 9 → 11 per Design 10; `ChargePoint::ALL`, `schedule_index`,
  ledger fuel arrays, `NodeCharges` shape audit; profile v5 (version,
  digest, doc section).
- Soundness per Design 11 (leaf in `visit_pattern_parts`; 19th
  `PATTERN_LABELS` entry + coverage test; `SpineClass`/`OrderCertainty` per
  the deterministic-order contract; cap-pushdown consumed at the node).
- Admission: `rows_per_invocation` feeding `PlanEstimate` /
  `ChargeLedger::for_plan` so a declared-huge relation is refused at
  admission when it must be.
- `charge_points.rs` end-to-end tests: exact invocation+row fuel;
  parallel/sequential trip parity; below-first-row-cost deterministic
  truncation via the governed channel (not error).
- Frozen corpus: new `vectors/sparql-governors/` cases via the SANCTIONED
  generator only (zero/exact/one-below for both points; parallel parity;
  truncation; deaf-relation stop) + pinned-digest reconciliation through
  the repo's regeneration path; existing case files must remain
  byte-identical.
- Executable checks: `cargo test -p purrdf-sparql-eval` green including
  `governor::`/`charge_points::`; the profile-digest-change test green with
  the new constants; frozen-corpus check green.
- Commit (`feat(sparql)!:` breaking footer: governor profile v5) → push →
  issue update → wiring-verifier.

## Task 6: End-to-end round-trip, ripple, conformance, bench, receipts, doc sync

- The acceptance demonstration: SPARQL TEXT → parse (namespace configured)
  → algebra (asserts the node) → eval over `MemoryRelation` → results, and
  the same query serialized and re-parsed byte-exactly.
- Exhaustive-match ripple: every `GraphPattern` match the compiler flags
  (`sparql-conformance/run.rs`, `purrdf/reasoning.rs`,
  `shapes/prebinding.rs`, `rdf/projections/dataset_description.rs`,
  `bgp.rs`/`binop.rs`/`expr.rs`/`construct.rs` arms) — every arm a real
  handling decision, no `unreachable!` stubs. Verified by
  `cargo check --workspace --all-targets` (this is a compile check, not the
  forbidden validation gate) and `make wasm`-equivalent narrow wasm checks
  (`cargo check --target wasm32-unknown-unknown` for each touched release
  crate).
- Registered-IRIs-in-receipt per Design 7 (explanation/profile identity).
- Conformance: `suite/purrdf-property-functions/` (manifest-driven
  `mf:QueryEvaluationTest`s; relations loaded via
  `MemoryRelation::from_graph` beside the existing `EXT_NS` harness config
  in `sparql-conformance/src/run.rs:289-301`) + the three-edit
  suite-addition contract (conformance-matrix script, baseline budget,
  `--write-doc` scoreboard). Stage 1 runs ONLY the new suite's narrow test
  target; the full conformance gate is run in stage 2/3.
- Bench: one `query_eval.rs` case exercising a property-function query
  (report-only; no speedup asserted).
- Doc-claims sync: the three narrowness-claim sites
  (`sparql-eval/src/lib.rs:32-41`, `error.rs:39-50`, `eval.rs:8`) name the
  seam; `scripts/check-doc-claims.py` green. Clean the dangling
  `Per CONSTITUTION Principle 17` process reference at `expr.rs:1956`
  (baseline own-on-sight; no CONSTITUTION.md exists).
- Executable checks: `cargo test -p purrdf-sparql-eval -p
  purrdf-sparql-algebra -p purrdf-core -p purrdf-datalog` green; new
  conformance suite's cases green via its narrow target;
  `cargo bench -p purrdf-sparql-eval --no-run` compiles;
  `python3 scripts/check-doc-claims.py` and
  `python3 scripts/check-issue-refs.py` green;
  `cargo check --workspace --all-targets` green.
- Commit (`test(sparql):` or `feat(sparql):` as content dictates) → push →
  issue update → wiring-verifier.

## Task 7: Create Pull Request (MANDATORY FINAL STEP)

1. Run the Mechanical Deferral Scan from `_no-deferrals.md` against
   `git diff origin/main...HEAD`; resolve every hit before opening the PR.
2. Verify all commits pushed; create the PR:
   `gh pr create --title "feat(sparql)!: add a property-function registry" --body "Closes #185 …"`
   with summary, per-change list, and breaking-change footers (new
   `GraphPattern` variant; governor profile v5).
3. `gh pr comment <PR> --body-file <this-plan-file>`.
4. Post to PR AND issue: answers to "What are you least confident about and
   why?" and "What do I not know but should?".
5. Report: "Stage 1 complete. PR #<num> created for issue #185. Ready for
   /stage2 185".

## Execution Rules (READ BEFORE STARTING)

1. Task 0 first; all work inside `.worktrees/185-property-functions`; never
   on main.
2. Commit after EACH task via github-commit-agent; push via
   github-push-agent; never batch, never run git directly in main context.
3. NEVER run the full validation gate (`make check`/`make conformance`/
   workspace-wide TEST suites) in stage 1 — narrowest command that proves
   the work (`cargo check --workspace --all-targets` in Task 6 is a
   compile-only ripple check and is allowed).
4. Pre-existing failures on origin/main: one line, not chased.
5. NEVER poll with `sleep`; background long commands, check at most twice.
6. If 30+ minutes pass without editing a source file, stop and implement
   the next task.
7. No deferral language anywhere; no issue/PR numbers in repo files;
   `example.org` fixtures only.

paudley:
## Stage-1 confidence questions

**What am I least confident about, and why?**

1. **The certainly-bound analysis in the feasibility orderer is deliberately conservative** (only in-chain bindings and provably-bound outer context count). A query whose only feasible order depends on bindings the analysis declines to claim (e.g. through an OPTIONAL left arm) will be refused at admission even though the evaluator could have run it. Every such refusal is loud and names the stuck atom, so this errs on the honest side — but the boundary is a judgment call worth review.
2. **The admission estimate composition rule** (`rows_per_invocation × driving-arm prediction`, contributed as the node's own peak — no cross-node products, matching the existing survey's max-over-peaks shape). Under-estimation is safe (live ceilings still trip); the rule's usefulness for genuinely huge relations depends on callers declaring honestly.
3. **Registered IRIs acting as parser namespaces** (`prepare_for` derives `property_fn_namespaces` from the registry). Consequence: an IRI that merely EXTENDS a registered IRI's spelling is lowered to a call and hard-errors as unregistered. Fail-loud and documented, but a caller registering `…/rel/a` who also uses `…/rel/ab` as an ordinary data predicate will be surprised.

**What do I not know but should?**

1. Whether the full workspace gates (`make check`, `make conformance` scoreboard write, `make wasm`, `make doc`) are green on the final branch state — stage 1 ran crate-scoped and narrow-suite verification only, per the staged workflow.
2. Whether real downstream consumers (full-text, path binding, kNN) will want the cursor to expose a seek/rewind capability beyond `next()` — the current pull contract is minimal by design; their issues will say.
3. Whether any out-of-tree consumer matches exhaustively on `GraphPattern` or `ShaclQueryOptions` — both changed shape; the breaking-change footers say so, but the blast radius outside this workspace is unverified.

paudley:
# Stage 2 Gap Remediation — Issue #185 / PR #201 (`paudley/185-property-functions`)

## Context

Stage-1 delivered the property-function seam (HEAD `84cb772c`, 6 commits, 90
files). Stage-2 analysis (gap-analyzer + completion-adversary, both against
the live worktree; adversary drove the published `purrdf::sparql` facade from
an out-of-tree consumer) confirms the core executes in production (registry,
algebra node, dispatch, substitution, scope, mint-nothing, reference
relation, unregistered-error, crate tests — all DEMONSTRATED), and finds the
branch **not ready**: one RED CI gate the branch itself caused, a dark
governor half, an IRI prefix-hijack, a confirmed planner defect (CodeRabbit
Major), and a cluster of contract/documentation shortfalls. Adversary
verdict: FINDINGS-OPEN — all findings below are binding.

**Human decisions recorded (AskUserQuestion this session):**
1. Governed-lane registry access = **options-struct refactor** (one
   QueryOptions-style struct threaded once through the governed entries;
   bundled into this already-breaking release).
2. Cursor ceiling = **extend the license** (Lateral's right operand gets a
   licensed row ceiling under the existing order-certainty guard; frozen
   SERVICE vectors must stay byte-identical; spy-relation test asserts
   `Some(k)` arrives in the correlated shape).

Adjudicated NOT a gap (adversary's binding verdicts, kept on record): the
Python/CLI/wasm/C-API registration decline stands (trait-object argument
holds; headline demonstrated through the Rust API — the gap-analyzer's
PYTHON-SURFACE objection is overruled by the adversary's cleared verdict);
the 4 "placeholder" scan hits are the pre-existing corpus idiom (panics if
one survives); `.deficiencies` and follow-up hunts are clean.

## Gap list (merged; adversary verdict wins on overlap)

### CRITICAL

**GAP-1 — `make doc` / CI `doc` job RED** *(CRITICAL-2 / F1)*
`crates/sparql-eval/src/eval.rs:9` redundant explicit link target
(`-D rustdoc::redundant-explicit-links`) kills documentation of the crate.
Sweep the other new intra-doc links with the same pattern
(`property_fn.rs:9`, `engine.rs:1359` flagged — fix every redundant one,
keep the genuinely-different label/target ones). Fix, run the full
`make doc` recipe green.

**GAP-2 — Governed lane cannot carry a registry (dark governor half + silent wrong answers)** *(CRITICAL-3 / F2; human decision 1)*
`query_governed`/`query_governed_with_source_view`/`query_prepared_governed_view`/
`query_governed_fallible_view` hard-code `None` at three `admit` sites and
have no registry parameter; a registered relation's predicate on those paths
parses as an ordinary BGP triple and silently returns empty rows (the exact
"silently no-op" the issue forbids). Remediate per the decision: introduce a
`QueryOptions`-style struct (registry + user functions + prebinding + mint
prefix + whatever `apply_shacl_query_options` carries) threaded ONCE through
the governed entries via `prepare_for` and `admit(relations)`; keep thin
compatibility wrappers where sensible; re-point the corpus harness off the
SHACL-shaped detour (`governor_corpus.rs:958-963`); the admission-estimate
channel (`rows_per_invocation` → `PlanEstimate`) must be live on every
governed path. Breaking-change footer updated. End-to-end test: a governed
property-function query through the NEW surface spends the v5 charge points
and returns correct rows; a registered-but-unconfigured path can no longer
silently no-op.

**GAP-3 — Registered IRIs prefix-match as namespaces (data-predicate hijack)** *(CRITICAL-4)*
`prepare_for` pushes full registered IRIs into `property_fn_namespaces`;
`is_property_fn` uses `starts_with`. Registering `…/rel/a` hijacks data
predicate `…/rel/ab` → spurious unregistered-error. Fix: two sets on
`ParserOptions` — caller-declared `property_fn_namespaces` (prefix, as
documented) and registry-derived `property_fn_iris` (EXACT match);
recognition = prefix-namespace OR exact-IRI. Regression test: register
`…/rel/a`, query `<…/rel/ab>` → stays a BGP triple.

**GAP-4 — Orderer rebuild plans against the final bound set** *(HIGH-1 / CodeRabbit M1 — confirmed)*
`property_fn_plan.rs:216-228`: selection mutates `bound`; rebuild calls
`plan_pattern(..., &bound)` with the fully-accumulated set, so a nested
chain in atom i is planned as if atoms i+1..n's bindings existed —
converting a prepare-time refusal into a mid-evaluation hard error. Fix:
`bound_before` snapshot per atom, replayed in the rebuild. Test: chain where
only a LATER sibling binds the call's required var — refused at prepare
time with the precise message, not admitted.

### HIGH

**GAP-5 — Ceiling license never issued in the Lateral shape** *(F4; human decision 2)*
`child_row_ceiling` has `Lateral` in the `=> None` arm, so `open()` receives
`None` in the parser's normal correlated shape; the kNN/top-k story never
fires. Extend the license: `Lateral`'s right operand receives the row
ceiling where the EXISTING `SpineClass`/`OrderCertainty` guard permits;
spy-relation test asserts `Some(k)` in `{ data . call }  LIMIT k`; frozen
SERVICE vectors byte-identical; soundness coverage tests updated. Also fix
the LSP concern: `MemoryRelation` honors `ceiling` (stop after k rows) and
documents it as the pattern consumers copy (L-6); confirm/correct the
`cell_row_ceiling` design note (L-7).

**GAP-6 — Frozen corpus does not pin the new charge points as promised** *(HIGH-3 / F3)*
Delivered: one aggregate `fuel` band triple whose `zero` member trips at
generic node-entry without touching the seam; no per-point isolation, no
parallel-parity vector (fixture below `PARALLEL_MIN_ROWS`), no
below-first-row truncation vector. Regenerate via the sanctioned generator:
per-charge-point-isolating band cases (invocation-heavy/row-light and
1-invocation/many-row shapes), a parallel-parity case above the parallel
threshold, a below-first-row truncation case; re-pin `GOVERNOR_CORPUS_DIGEST`
+ sha; existing vectors byte-identical. Correct `docs/CONFORMANCE.md:89`
(the "three bands for the two charge points" mis-claim) and `:107` (29→34,
plus whatever the new cases make it).

**GAP-7 — Governed receipt carries no relation identity; explain discards descriptors; fingerprint omits volatility** *(HIGH-4 / F8)*
Attach the sorted descriptor set (or the `registry_fingerprint` digest) to
the governed outcome/evidence; render descriptors (not bare IRIs) in
`QueryExplanation`; include volatility in `registry_fingerprint` (two
registries differing only in volatility must not share a plan-cache entry —
test). Test: two different relation impls under one IRI produce
distinguishable receipts.

**GAP-8 — `from_graph` producer-without-consumer; conformance suite closed-loop** *(HIGH-2 / F5)*
Drive the harness relation tables from manifest fixture graphs via
`MemoryRelation::from_graph` (as the plan committed); keep at least the
data-driven cases graph-loaded.

**GAP-9 — Dead ripple arms with false invariant comments** *(HIGH-5 / CodeRabbit M2)*
`dataset_description.rs` and `slice/ownership.rs` parse with default
`ParserOptions`, so their new arms are unreachable and their comments state
false invariants (dependency extraction DOES emit the relation IRI). Thread
caller-supplied `ParserOptions` into both entry points + coverage (the
CodeRabbit-preferred fix), making the arms real.

**GAP-10 — UPDATE surface: false comment + registry not threaded** *(F6 / MEDIUM-2)*
A call node DOES reach update (demonstrated: INSERT…WHERE errors
unregistered), and `update.rs:629`'s comment claims it cannot. Thread the
registry through the update surface (via the GAP-2 options struct — one
seam), route update WHERE through `prepare_for`/`plan_query`, pass relations
into `admit_where`; e2e test: INSERT…WHERE over a registered relation
inserts the relation's rows. Fix the comment.

**GAP-11 — Conformance suite does not exercise the design's core; weak/vacuous cases; no first-party inventory tripwires** *(MEDIUM-4 / F7 / F8-note)*
Add a mode-restricted (`bf`-only) harness relation + a case whose textual
order is infeasible and must be reordered + negative/ledgered cases
(unregistered, arity, infeasible chain) as the harness supports; make
`bb-member-false` discriminating (falsity requires the relation to have run);
differentiate `filter-over-call` from `fb-bound-output`; add inventory
tripwires for the five first-party suites (deleting a suite dir must fail).

### MEDIUM

**GAP-12 — Panic containment excludes the declaration methods** *(MEDIUM-1 / CodeRabbit M3)*
Contain `arity`/`modes`/`volatility`/`rows_per_invocation` reads behind a
contained helper (the .goals-consistent option); hoist `modes()` out of the
per-input-row loop while at it; module doc then true as written.

**GAP-13 — Doc-block capture defects + sweep** *(MEDIUM-5 / m2 m3)*
`safety_registries` captured `may_fork_row_loop`'s doc; `prepare_for`
captured `eval_ctx`'s. Fix both; sweep every `///` block this branch
inserted above an existing item for the same capture pattern.

**GAP-14 — `EvalError::Function` misnames the callee** *(MEDIUM-6 / m1)*
Render as `host function error:`; update affected assertions.

**GAP-15 — Stale/false claims cluster** *(F9 / m5 m6 m7 / L-1..L-4, L-8, L-9)*
`error.rs:42-44` false claim (a call inside SERVICE does surface
Unsupported); "first-match-wins" doc + misnamed test (order-independent
`any`); `PROFILE.md:199` duplicated preposition; governors README
per-invocation vs per-row contradiction (align to the stronger proven
claim); `pulls` counts the terminating next (restate as "calls to next");
positional-prefix=false rationale note in §8.1; assert the
`native-sparql-property-function` diagnostic code in the refusal tests.
CLI `--explain`: document that the CLI has no registration surface so its
relations block is legitimately empty (or render conditionally — pick the
honest minimal form).

**GAP-16 — Dispatch hot-loop allocations** *(n3 / L-5)*
Hoist the per-input-row `args`/`refs` Vec allocations (fixed width) out of
the `'input` loop like `seed`/`values`. Report-only; the bench exists; no
speedup asserted.

### Wrap-up

**GAP-17 — Full gates on final state**: `cargo fmt --all` then `make check`,
`make conformance`, `make wasm`, `make doc` — all green on final HEAD.
Then Phase 5 feedback loop (CodeRabbit re-review of the new commits),
Phase 6 adversary re-run (binding exit gate; agent ab010a094518df472
continues for delta re-audit), final summaries to issue #185 + PR #201, PR
body updated (options-struct breaking change added; laundering sentence
about deferred gates replaced with the actual green-gate statement).

## Execution (stage-2 Phase 4 discipline)

Order: GAP-1 (unblocks CI) → GAP-3 → GAP-4 → GAP-2 (biggest; options struct)
→ GAP-10 (rides the struct) → GAP-5 → GAP-6 → GAP-7 → GAP-9 → GAP-8 →
GAP-11 → GAP-12 → GAP-13 → GAP-14 → GAP-15 → GAP-16 → GAP-17.
Per gap: implementation sub-agent (NO-DEFERRALS clause) → commit → push →
post summary to issue AND PR → wiring verification where production code
changed. No batching. All work in `.worktrees/185-property-functions`.

## Verification

- `make doc` green (GAP-1) and all four gates green at the end (GAP-17).
- Governed property-function query via the new options surface spends
  `property-function-invocation`/`-row` and returns correct rows; silent
  no-op path eliminated (GAP-2 e2e).
- `…/rel/ab` regression stays a BGP triple (GAP-3).
- Nested-chain refusal at prepare time (GAP-4).
- Spy relation observes `Some(k)` in the correlated shape; SERVICE vectors
  byte-identical (GAP-5).
- Corpus pins both charge points in isolation + parity + truncation
  (GAP-6); corpus-frozen check green.
- Adversary re-run: PASS, every criterion DEMONSTRATED-IN-PRODUCTION.

paudley:
**Stage 2 — GAP-1 closed** (`601fd076`, CRITICAL) — `make doc`/CI `doc` job restored to green (redundant intra-doc link target dropped; sibling modules use fully-qualified bracket labels). Verified with the exact Makefile recipe: exit 0, zero warnings.

paudley:
**Stage 2 — GAP-3 closed** (`54e37fe6`, CRITICAL) — registered relation IRIs were injected into the parser's PREFIX-matched namespace set, so registering `…/rel/a` hijacked the ordinary data predicate `…/rel/ab` (parsed as a call, hard-errored as unregistered, wrong-cause diagnostic). `ParserOptions` now carries a second EXACT-match set (`property_fn_iris`) that the engine fills from the registry; `property_fn_namespaces` keeps prefix semantics for caller-declared namespaces; recognition = prefix-namespace OR exact-IRI; the plan-cache key covers both sets. A stage-1 test that had pinned the hijack as expected behavior now pins the genuine namespace-declared-but-unregistered refusal, and regressions at parser, injection, and e2e level prove `…/rel/ab` stays ordinary data. BREAKING (footered): `ParserOptions` gains a field. Verified: `make check` + `make doc` exit 0, conformance 44/44, workspace clippy clean.

paudley:
**Stage 2 — GAP-4 closed** (`547284b9`, CRITICAL / CodeRabbit Major) — the orderer's rebuild pass planned every atom against the FINAL bound set (selection had mutated it), so a call nested inside an earlier atom was planned as if later siblings' bindings existed — wrongly admitted at prepare time and hard-erroring per row at evaluation. Fix: per-atom `bound_before` snapshots zipped into the rebuild, so each atom is planned against exactly the set it was chosen against. The regression pins the real failing shape (bf-only call nested in a UNION scheduled before the BGP that binds its subject — verified failing pre-fix by reverting): now refused at prepare time with the no-feasible-order diagnostic, zero cursors opened. sparql-eval 482 lib tests green; all existing orderer tests unchanged.

paudley:
**Stage 2 — GAP-2 closed** (`5321203b`, CRITICAL; human decision: options-struct refactor) — the governed lane could not carry a property-function registry (three hard-coded `None` admit sites; the corpus harness detoured through the SHACL entry), and worse, a registered relation's predicate on those paths parsed as an ordinary BGP triple and silently answered the empty bag.

**Resolution:** one `QueryOptions` struct (prebinding + user functions + property functions + mint prefix — the full `apply_shacl_query_options` set) is now a REQUIRED parameter of every governed entry, threaded once through a single body (`prepare_for` → `admit(relations)` → one application seam). `ShaclQueryOptions` is that struct renamed; duplicate entries folded; four hand-rolled ungoverned bodies became delegations. A prepared-plan residue was found and closed: `PreparedQuery` now carries the registry fingerprint it was parsed under and a plan/registry disagreement is refused. Corpus harness re-pointed onto the natural governed entry — pinned consumption bytes did not move (plus a test pinning that the per-call and operation entries spend identically). E2E: governed property-function query returns correct rows, spends `property-function-invocation`/`-row` (ledger-asserted), and the declared-huge-relation admission refusal works through the same entry. BREAKING (footered): governed entry signatures, rename, PreparedQuery construction.

**Verified:** `make check` exit 0, workspace 4677 tests / 0 failed, governor_corpus 7/7, doc gate clean, zero vectors/generated churn.

paudley:
**Stage 2 — GAP-10 closed** (`1b758c99`, HIGH) — a property-function call node reaches an UPDATE's WHERE (a triple-pattern context), but the update seam couldn't hand it a registry and carried a false comment claiming a call "cannot appear here" while hard-coding `None` into the admission survey.

**Resolution:** `QueryOptions` threads through the UPDATE surface exactly as the query lane (required on `update_governed`; new `update_with_options` ungoverned sibling; the `SparqlEngine::update` trait signature unchanged, delegating under `EMPTY`); the WHERE pattern is feasibility-planned against the registry (`plan_where_pattern` extracted from `plan_query`); the WHERE `EvalCtx` builds through the same one application seam; the false comment fixed alongside the code. E2E: `INSERT…WHERE` over a registered relation inserts exactly the relation's rows; a governed update spends the charge points, trips deterministically one unit below exact-spend, and applies nothing on the trip; the unregistered case pins the `native-sparql-property-function` diagnostic code. BREAKING (footered): `update_governed` takes `QueryOptions`.

**Verified:** `make check` exit 0, `make doc` green, workspace check/clippy clean, sparql-eval suites green.

paudley:
**Stage 2 — GAP-5 closed** (`69635da7`, HIGH; human decision: extend the license) — `open(args, ceiling)` never received `Some(k)` in the parser's normal correlated shape `Lateral(Bgp, call)`, so the kNN/top-k design story never fired.

**Resolution — a sounder route than prescribed:** rather than promoting the Lateral right edge (which would have flipped two pinned certificate bits), the fused Lateral+call operator now consumes the ceiling recorded at the **Lateral itself** — the node whose output rows the dispatch produces, and whose own certificate already licenses it. No new license minted; no edge classification moved; `child_row_ceiling`'s Lateral arm documents WHY nothing crosses that edge (the compatibility test drops rows). Spy demonstration: correlated `LIMIT 2` → `open()` observes `Some(2)` (was `None`); ORDER-BY guard still refuses; standalone unchanged.

**Latent unsoundness found and fixed while honoring the ceiling in `MemoryRelation`:** the ceiling counts emitted rows but the engine post-filters (repeated variables, partly-bound quoted triples) — a ceiling-honoring relation could hand back a short bag reported as complete. The ceiling is now withheld unless every emitted row agreeing with bound positions is guaranteed admitted (`args_are_admission_transparent`), and the trait states the accounting precisely. `MemoryRelation` honors the ceiling and is the documented pattern consumers copy.

**Verified:** sparql-eval 13/13 binaries green (incl. the differential governor-correctness oracle), governor_corpus 7/7, frozen vectors byte-identical (7/7), clippy/fmt/doc clean.

paudley:
**Stage 2 — GAP-6 closed** (`c4ec1979`, HIGH) — the frozen corpus banded the two v5 charge points only through aggregate fuel (a schedule moving them in opposite directions passed unchanged), the `zero` case tripped without the seam ever firing, and the parity + below-first-row behaviors existed only as unit tests.

**Resolution — isolation solved at the root:** relation cases pin a **fourth record**, `expected/<case>.charges` — metered fuel decomposed per charge point off the public per-node ledger, re-derived every run and required to sum to the `.metered` fuel. Seven new cases (corpus 34→41, bands 28→34): an invocation-isolating pair (12 invocations × 0 rows), a row-isolating pair (1 invocation × 12 rows), a 1200-row pair above the fork threshold whose boundary IS the sequential/parallel parity claim (METERED takes the sequential driver, fuel-only bands take the parallel one — with a test proving the fixture exceeds the fork threshold so the lane can't go inert), and a below-first-row window case whose authored fuel value the harness re-derives as unique. No `zero` member on the isolating lanes, deliberately and documented (a zero ceiling trips before any relation resolves). All through the sanctioned generator — zero modified `expected/` files; `GOVERNOR_CORPUS_DIGEST` re-pinned; `docs/CONFORMANCE.md`'s band enumeration now arithmetically true (and `:107` corrected).

**Verified:** governor_corpus 11/11 (41/41 cases, 34 bands), conformance crate green, corpus-frozen 7/7, doc-claims OK, clippy clean.

paudley:
**Stage 2 — GAP-7 closed** (`ef61d652`, HIGH) — three receipt-identity holes: `registry_fingerprint` omitted volatility (two registries differing only in determinism class shared a plan-cache entry despite different parallel-safety classifications); `QueryExplanation` reduced relations to bare IRIs (two impls under one IRI → identical receipts, different answers); the governed outcome carried no relation identity at all.

**Resolution:** volatility folded into the fingerprint (pinned by a plan-cache test); `QueryExplanation::relations` renders full `PfDescriptor`s (IRI, arities, volatility, per-mode row bounds); new `RelationIdentity` (fingerprint + sorted IRIs) on `GovernedOutcome::Complete` and `BudgetExhausted`, computed once per execution from the already-validated prepared-plan fingerprint — refusal, complete, and truncated paths all carry the same identity. Corpus impact: none by design (the pinned records never read the new fields — verified, vectors byte-identical). BREAKING (footered).

**Verified:** full workspace tests green, governor_corpus 11/11, all hygiene/doc/wasm gates green.

paudley:
**Stage 2 — GAP-9 closed** (`5a858be7`, HIGH / CodeRabbit Major) — the dataset-description CONSTRUCT validator and slice ownership's dependency walk parsed with default `ParserOptions`, so their new `PropertyFunction` arms were dead code and their comments stated false invariants (the relation IRI WAS being emitted as a dependency edge).

**Resolution:** caller-supplied `ParserOptions` threaded into both entry points (`ConstructViewConfig::new_with_parser_options`; `OwnershipAnalyzer::with_parser_options`), making the arms real: a recognized call is refused as non-reproducible by the projection check, and a relation IRI is excluded from the dependency walk while sibling data predicates still evidence. Both behaviors covered by tests in both configurations (default unchanged, configured firing); comments rewritten to the now-true invariants.

**Verified:** rdf + slice suites green (442+75 lib + integration incl. the new coverage), workspace check clean, fmt/clippy/doc/issue-refs green.

paudley:
**Stage 2 — GAP-8 + GAP-11 closed** (`21615f15`, both HIGH; one commit — the shared-file hunks could not be split without hand-mangling untested intermediate states, noted in the commit body).

**GAP-8:** `MemoryRelation::from_graph` now has its production consumer — the harness relation tables live in `relations.ttl` as rdf:Lists-of-rdf:Lists (the encoding `from_graph` defines) and `harness_relations()` reads them from the graph; IRI→table binding stays harness code, only the data moved; a missing/torn fixture panics loudly (an empty registry would make every case a vacuous pass). All ten existing cases pass byte-unchanged.

**GAP-11:** the suite now exercises the design's core: `BoundSubjectLookup` (bf-only, generate-then-filter) + four cases — textual-order-infeasible-must-reorder (verified: breaking the binding makes it fail with the `ff` refusal), subsumption-served `bb`, unregistered-IRI refusal, infeasible chain. A minimal expected-failure channel rides the `mf:result` extension the harness already routes on (`.err` = substrings the diagnostic must contain; a case that completes FAILS, so it cannot pass vacuously). `bb-member-false` and `filter-over-call` restructured to be discriminating. **Bonus find:** the existing W3C inventory tripwires had NEVER RUN (`harness = false` left their `#[test]` fns dead) — all tripwires now live in a real libtest target covering the W3C groups AND the five first-party suites (verified to trip by deleting a suite dir); both differential corpora pin expected-refusal counts as identities. Scoreboard 810→814.

**Verified:** full conformance crate green (10 targets), sparql-eval green, fmt/clippy/hygiene/doc-claims green.

paudley:
**Stage 2 — GAP-12…GAP-16 closed** (`81491e6d`, `628b7a09`, `3b19032d`, `400f5616`)

- **GAP-13** (`81491e6d`): the two doc blocks captured by adjacent items restored (`may_fork_row_loop`, `eval_ctx`); branch-wide sweep found no further captures.
- **GAP-12 + GAP-16** (`628b7a09`, BREAKING footered): every reachable declaration read (`arity`/`modes`/`volatility`/`rows_per_invocation`/`describe`/fingerprint) is now panic-contained via `declaration_contained`; `modes()` hoisted out of the per-row loop; per-row argument buffers hoisted or stack-allocated (SmallVec at solution width — no heap allocation per row in the common case; report-only, no speedup asserted). The fork-safety volatility read stays uncontained BY DESIGN and the doc says so: a panicking declaration is caught at prepare before classification can run.
- **GAP-14** (`3b19032d`, BREAKING footered): `EvalError::Function` renders "host function error" — the variant carries both SHACL-AF user functions and relation failures.
- **GAP-15** (`400f5616`): eight stale/false claims corrected — SERVICE-surfacing prose, order-independent namespace recognition (test renamed to what it asserts), per-row bound in the governors README, `pulls` counter semantics, positional-prefix rationale in the profile doc, diagnostic-code assertions in the refusal tests, CLI explain relations-block note, duplicated clause removed.

Each commit was verified buildable in isolation (stash-and-test at every boundary). Full workspace tests, clippy, fmt, `make doc`, `make wasm`, and all hygiene scripts green on the final state.

paudley:
**Stage 2 — GAP-17: full gates GREEN on final HEAD `400f5616`**

| Gate | Verdict |
|---|---|
| `make check` (fmt + clippy -D warnings + build + full workspace tests + hygiene + wasm) | **GREEN** — 4713 passed / 0 failed |
| `make conformance` | **GREEN** — 3407 pass / 14 ledgered / 0 fail across 15 suites |
| `make wasm` | **GREEN** |
| `make doc` (workspace, denied warnings) | **GREEN** |

This closes the stage-1 confidence item and replaces the PR's former "gates are stage-2/3 work" deferral with the actual green-gate statement.

paudley:
**Stage 2 — feedback round closed** (`fcbff43b`) — CodeRabbit's second review (3 comments) plus one same-class hole owned on sight:

- **Major:** the federated (`query_with_source(_view)`), prepared (`query_prepared(_view)`), and — found during the fix — fallible (`query_fallible_view`, `query_prepared_fallible_view`) entries were still registry-free: a registered relation's predicate could silently parse as data, or a plan prepared under one registry could evaluate under another. All six now take `QueryOptions`, parse via the registry-aware path, refuse plan/registry disagreements (`native-sparql-property-function`), and apply options at the one seam. A mechanical signature sweep confirms **no registry-free evaluation entry remains** — every optionless signature is a convenience wrapper delegating with an explicit `QueryOptions` expression. BREAKING (footered).
- **Minor:** the hijack regression asserts bound cells (graph values vs the relation's would-be values), not just row count.
- **Minor:** the `apply_query_options` doc block re-attached to its own item (third instance of the capture class on this branch — the mechanical doc-capture sweep now reports zero hits on the final file).

**Verified:** full `make check` exit 0, `make doc` green, workspace clippy clean, all sparql-eval suites green (incl. 15/15 e2e, 13/13 fallible).

paudley:
# Stage 2 complete — issue #185 / PR #201 — READY FOR STAGE 3

## Completion-adversary final verdict (pass 2, HEAD `fcbff43b` + docs `a0104182`): **PASS**

Every acceptance criterion demonstrated in production from an out-of-tree consumer of `purrdf::sparql::*`:
- **C1–C5** registry/node/dispatch/substitution/scope: 3 relation rows over a dataset holding one unrelated triple; explain ledger shows the `PropertyFunction` node; substitution narrows 3→1 rows; `SELECT *` projects argument variables.
- **C6** governor cost like any other operator — now on the HEADLINE governed surface: `query_governed(…, QueryOptions{property_functions}, METERED)` → correct rows, `RelationIdentity` on the outcome, fuel 12 matching the explain decomposition (1 invocation + 3 rows) exactly.
- **C7** soundness: corpus 41/41 cases, 34 bands, per-charge-point `.charges` sidecars.
- **C8** mint-nothing strengthened: exact-IRI vs prefix-namespace split; the sibling-predicate hijack demo now stays ordinary data.
- **C9** reference relation round-trips; **C10** unregistered errors three ways (direct, governed-without-registry, prepared-plan/registry disagreement); **C11** sparql-eval 602/0.
- Plan commitments: feasibility ordering + mode restriction + reorder (suite cases), ceiling pushdown correlated `Some(2)` (was `None`), admission estimates on the governed path, shared ingest core, SERVICE-forward refusal, SHACL-body invocation, conformance 14/14, receipts, bench.

All nine pass-1 findings **DISCHARGED** by re-demonstration. Laundering scans over the 14 remediation commits: clean; `.deficiencies`: no entry (nothing descoped).

## What stage 2 closed (14 commits)

`601fd076` doc gate · `54e37fe6` exact-IRI hijack fix · `547284b9` orderer bound-set snapshots · `5321203b` QueryOptions through the governed lane (human decision) · `1b758c99` UPDATE surface · `69635da7` correlated ceiling license (human decision) + latent short-answer unsoundness fixed · `c4ec1979` per-charge-point frozen corpus · `ef61d652` relation identity in receipts · `5a858be7` reachable ripple arms · `21615f15` graph-driven harness + mode-restricted suite + inventory tripwires (found the W3C tripwires had never run) · `81491e6d`/`628b7a09`/`3b19032d`/`400f5616` doc captures, declaration-panic containment, host-function error prefix, stale-claims cluster · `fcbff43b` QueryOptions on every remaining entry (no registry-free evaluation entry remains) · `a0104182` corpus README note.

All four CodeRabbit reviews addressed (13 findings + 3 follow-ups + 1 same-class hole owned on sight).

## Verification (final HEAD)

`make check` green (4713/0, clippy pedantic/nursery clean) · `make conformance` GREEN (3407/14/0; scoreboard 814) · `make wasm` green · `make doc` green · frozen corpora 7/7 byte-identical · adversary-run independent gate confirmation.

**Ready for `/stage3 185`.**

paudley:
**Stage 3 — human-decided scope addition closed** (`263579ff`) — at the stage-3 gate the human decided to build the Python data-relation surface before merging (the earlier decline had been agent-adjudicated). `Store`/`MutableDataset` `query`/`query_governed`/`update`/`update_governed` accept `relations` (IRI → arity + rows), `relations_from_graph` (IRI → rdf:List-of-lists table head in the store's own default graph), and `property_fn_namespaces` — pure data through `MemoryRelation`, no GIL conflict (rows convert before detach); malformed/torn/duplicate declarations raise with the kernel diagnostic; shared converters de-duplicated. Stubs, README, and 35 new pytest cases; the Python suite is 588 passed / 1 skipped / 4 xfailed via `make pytest`. Also corrected three stale numbers in the PR body (adversary stage-3 F1: suite 14 cases, corpus relation lane 12 cases, make check 4716 tests).

paudley:
feat(sparql)!: add a property-function (magic predicate) registry

Adds the SPARQL property-function seam: caller-registered, multi-valued
relations keyed by IRI, invoked as if they were triple patterns
(?doc ex:contains ("needle" ?score)). Designed as query answering over
relations with limited access patterns: a relation declares its access
capabilities (adornments over the BindingPattern lattice) and the engine
chooses a feasible execution order, so honest capability declarations
are satisfied by a planner rather than punished with runtime failures.

## What this branch accomplishes (24 commits)

- BindingPattern adornment lattice hoisted from purrdf-datalog into
  purrdf-core (old path re-exported) as the shared access-pattern
  vocabulary.
- Algebra: GraphPattern::PropertyFunction with byte-exact predicate IRI
  and structural n-ary subject/object argument vectors (no cons-cell
  desugaring; () is a zero-length vector; nested collections
  hard-error); ParserOptions gains property_fn_namespaces (prefix,
  caller-declared) and property_fn_iris (exact, registry-derived) --
  the split that keeps a registered rel/a from hijacking the data
  predicate rel/ab; textual-order left-deep LATERAL chains; serializer
  round-trip; refused in CONSTRUCT/update templates.
- Relation contract: object-safe PropertyFunction cursor trait over
  TermValue -- declared arity, access modes (feasibility is lattice
  subsumption: fb serves bb by generate-then-filter), per-mode row
  bound, open(args, ceiling)/next() pull cursor with deterministic
  emission order, deaf-relation doctrine, panic containment over every
  reachable host call including the declaration reads.
  PropertyFunctionRegistry hard-fails duplicate IRIs; describe() is the
  self-describing-data channel. MemoryRelation is the reference
  implementation, honours the ceiling, and builds from an rdf:List of
  rdf:Lists via from_graph.
- Planning: prepare-time greedy feasibility ordering under
  certainly-bound analysis that never over-claims, with per-atom
  bound-set snapshots so nested chains are planned against exactly the
  set they were chosen against. Unregistered IRIs, arity mismatches and
  infeasible chains are refused at admission, before the first governor
  charge. Plan-cache keys carry both namespace sets and a registry
  fingerprint that includes volatility.
- Evaluation: per-input-row dispatch binding arguments directly from
  the row (literals included), structural row unification, blanks
  projected away, rows ingested through a governed core shared with
  SERVICE, licensed row-ceiling pushdown that reaches the correlated
  Lateral shape (a ceiling-honouring relation is protected by an
  admission-transparency gate so post-filtering can never shorten a
  certified answer), SERVICE-forwarding refusal, SHACL constraint-body
  invocation, and UPDATE WHERE dispatch.
- One options seam: QueryOptions (prebinding, user functions, property
  functions, blank-mint prefix) is a required parameter of every query
  and update entry -- governed, federated, prepared, fallible -- with
  prepared plans carrying the registry identity they were parsed under
  and refusing a disagreement. No registry-free evaluation entry
  remains; a registered relation can never silently no-op.
- Governors (profile v5): property-function-invocation and
  property-function-row charge points; admission control prices a call
  from its declared row bound; the frozen corpus pins the two points in
  isolation via per-case .charges decompositions that must sum to the
  metered fuel, plus parallel-parity, below-first-row truncation, and
  deaf/cooperating cancellation lanes (12 relation cases). Relation
  identity (fingerprint + IRIs) rides the governed receipt and full
  descriptors ride the explanation.
- Conformance: a 14-case manifest-driven suite whose relation tables
  live in the manifest graph (from_graph), including mode-restriction,
  feasibility-reordering, and negative refusal cases graded by
  diagnostic text; inventory tripwires for the five first-party suites
  (the W3C tripwires were found dead -- compiled under harness=false
  and never run -- and now execute).
- Python surface (human-decided at the stage-3 gate): Store and
  MutableDataset query/query_governed/update/update_governed accept
  relations (IRI to arity-and-rows), relations_from_graph (IRI to an
  rdf:List table head in the store's default graph) and
  property_fn_namespaces -- pure data through MemoryRelation, rows
  converted before the GIL is released, kernel diagnostics propagated;
  35 pytest cases. wasm/C-API remain at UserFunctionRegistry parity per
  the same human decision.
- Report-only criterion bench case; narrowness docs updated; a
  pre-existing dangling process reference cleaned.

## Requirements satisfied

Every acceptance criterion of the issue -- registry, algebra node,
dispatch, substitution, scope walker, governor charge point accounted
like any other operator, soundness classification, mint-nothing exact
IRI keys, reference relation round-trip, unregistered-IRI hard error,
crate tests with end-to-end and governor-charge coverage -- plus the
enlarged plan commitments, demonstrated in production by the
completion adversary from three independent consumers: an out-of-tree
Rust consumer of the published purrdf facade, the shipped CLI, and
actual Python. Charge points were proven live across the FFI by
arithmetic (fuel 6/9/12/18 for 1/2/3/5 relation rows; a 3-row relation
costs exactly one invocation charge more than an equivalent BGP).

Rust-first with the Python surface delivered; hard fails everywhere a
configuration or declaration is wrong (no silent no-ops, duplicate
registration refused, plan/registry disagreement refused); RDF 1.2
first-class (quoted triples as arguments, triple-term positions in the
escape matrix); wasm-clean (make wasm green); deterministic (emission
order contracts, DetHashMap, fingerprinted plan cache); no minted
vocabulary (example.org fixtures; registry keys are caller IRIs);
report-only benches.

## Completion-adversary: PASS (stage 3, final at HEAD b2d6b37f)

Fresh stage-3 adversary drove every criterion itself: registry rows
over a dataset holding one unrelated triple; explain ledger showing the
node with a fuel decomposition that sums; substitution narrowing 3
rows to 1; SELECT * projecting argument variables; exact-IRI
anti-hijack (rel/ab stays data with rel/a registered); mode-restricted
reordering and prepare-time infeasibility refusals with precise
diagnostics; correlated ceiling Some(2) observed by a spy relation;
admission refusal of a declared-huge relation naming the estimate;
governed truncation certificates; SERVICE-forward refusal; panic
containment in open and in declaration reads; SHACL-body invocation
flipping conformance; UPDATE WHERE materializing relation rows; the
Python surface end-to-end including governed trips and four
malformed-config raises. Two stage-3 findings (stale PR-body numbers;
the conformance scoreboard drift the new pytest cases caused) were
fixed and re-verified by demonstration. Laundering scans clean;
.deficiencies carries no entry -- nothing was descoped; the one scope
question was escalated to the human, who chose to build it, and it was
built.

## Deferral gate

Mechanical scan hits are all evidenced false positives: the corpus
"placeholder" lines are the pre-existing regeneration idiom whose
harness panics if a placeholder survives (none do), and the single
"deferred" match is prose arguing for an eager check. PR-comment hits
are the remediation plan and summaries themselves.

## Verification at final HEAD

make check green (4716 tests / 0 failures; clippy pedantic/nursery
warning-free); make conformance GREEN (3442 pass / 14 ledgered / 0
fail across 15 suites; SPARQL scoreboard 814; Python binding suite
588); make wasm green; make doc green under denied warnings; frozen
corpora 7/7 byte-identical with the governor corpus at 41 cases / 34
bands; governor profile v5 digest
1e8e8f2b340bf85addaa55e3e4b323ca75325ad9465b9846cf96ac13b557317e.

## Implementation notes for future developers

- Feasibility is subsumption on purrdf_core::binding_pattern: an
  invocation is served by any declared mode requiring no more bound
  positions. Declare the weakest modes that are honest; the engine
  filters bound positions after emission.
- The ceiling handed to open() is an optimization license, withheld
  unless every emitted row agreeing with the bound positions is
  guaranteed admitted; honour it by emitting at most that many
  agreeing rows.
- QueryOptions is the one options seam; a new per-query capability is
  a field there and one application-site edit, never a new entry
  point.
- The registry is parse configuration: registered IRIs are recognized
  exactly and derive the parser setup at prepare_for; prepared plans
  are fingerprint-checked against the evaluation registry.
- The .charges corpus sidecar decomposes the METERED run per charge
  point and must sum to the metered fuel; relations.tsv records the
  banded run -- different runs, documented in the corpus README.

## Conflicts

None: origin/main did not advance while the branch was open; the
merge-into-main result is the branch tip.

PR #223, opened 2026-09-01
fix(iri): make relative IRI-reference resolution one base-aware layer across every syntax

### LINKED ISSUE #222: turtle parser stumbles over empty IRIs
### What happened, and what did you expect?

According to https://www.w3.org/TR/turtle/#grammar-production-IRIREF empty IRIs should be allowed in turtle.

However, I get purrdf: error native-codec-iri-missing-value: GTS IRI term requires a non-empty value when using turtle documents with empty IRIs.

### Minimal reproduction

<> a <http://example.org/test> .

$ purrdf convert --to ntriples --from turtle /tmp/test
purrdf: error native-codec-iri-missing-value: GTS IRI term requires a non-empty value

### Language surface

Rust crate

### Version

0.12.0

### Platform

Linux x86_64

paudley:
Closes #222

## What was reported

`<> a <http://example.org/test> .` failed with `native-codec-iri-missing-value`.

## What was actually wrong

That error is a **downstream guard**, not the bug. The defect is upstream and larger:

```
$ printf "<foo> a <http://example.org/test> .\n" | purrdf convert --to ntriples --from turtle -
<foo> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://example.org/test> .   # exit 0 (!!)
```

A **non-empty** relative IRI was not rejected. It was silently interned into the frozen IR as the IRI term `"foo"` and emitted as **invalid N-Triples**. `<>` differed only in that it interned as the empty string and tripped the GTS fold. `DocParser::resolve_iri` returned `raw.to_owned()` when no base was in scope, and `validate_iri` was unreachable from that path. PurRDF was minting non-absolute IRIs into an RDF 1.2 graph, contrary to RDF Concepts 3.1.

Behind that sat the real structural problem: **six surfaces each answering the base question
their own way**. Two of them were genuine duplicate implementations of RFC 3986; the other four
already called `purrdf-iri` and got the base *around* the call wrong. This distinction was
overstated in an earlier revision of this description (which claimed five hand-rolled resolvers)
and is corrected here — the counts below were re-verified against `origin/main`.

**Two duplicate RFC 3986 implementations** — each a private `remove_dot_segments` +
`resolve_relative_iri` pair, ~110 lines apiece and byte-identical to one another:

| | on `origin/main` | now |
|---|---|---|
| Turtle / TriG / N-Triples / N-Quads (`text_parse.rs`) | `has_iri_scheme`, `remove_dot_segments`, `split_raw_path_suffix`, `split_base_for_path`, `resolve_relative_iri` | **deleted**; the parser threads a `BaseScope` |
| RDF/XML (`rdfxml.rs`) | the same five functions, a verbatim copy | **deleted** |

**Four surfaces that already delegated to `purrdf-iri`** and are not resolver deletions. Their
defects were real but of a different kind — what happened *around* the correct call:

| | on `origin/main` | the defect |
|---|---|---|
| ShExC / shape map | `purrdf_iri::parse(base)` then `base.resolve(reference)` — main even shipped a test named `base_resolution_uses_purrdf_iri` | with **no** base in scope it returned the raw reference verbatim |
| SPARQL | `purrdf_iri::Iri::resolve(&base_iri, s)` | `BASE <relative>` was taken verbatim instead of rebinding |
| JSON-LD / YAML-LD | `purrdf_iri::parse(base)` then `base.resolve(reference)` | the caller's base was bound and then dropped, so `--base` was a silent no-op |
| TriX / HexTuples | no resolver at all — a `validate_iri` whose whole body was a `contains(':')` test | a `path-noscheme` reference like `a:b` passed as "absolute" |

So: **two implementations deleted, six surfaces unified onto one layer.** Seven different
diagnostic spellings for one condition, now one.

Note also that `validate_iri` still EXISTS in `rdfxml.rs`, `trix.rs` and `hextuples.rs`. It was
not removed; it was reduced to a three-line delegation to `BaseScope::resolve_absolute_only`.

## What this changes

**One layer.** New `crates/iri/src/base.rs` — `BaseIri` (absolute by construction), `BaseScope` (the in-scope base *stack*, since `xml:base`, JSON-LD `@base` frames and Turtle `@base` rebinding are all genuinely stacked), `ScopedBase`, `BaseOrigin` (so a diagnostic can say which `@base` on which line was in force), `BaseIri::relativize` (the exact inverse, beside its forward direction rather than in a serializer), and `IriError::{NoBase, NotAbsoluteByGrammar}` with `diagnostic_code()` as the single owner of the code strings. `BaseIri::resolve` delegates to the existing, already-correct `Iri::resolve`; no second algorithm was written. It is the only place in the workspace permitted to contain RFC 3986 merge/dot-segment arithmetic — the two duplicate implementations are deleted, and `scripts/check-iri-resolver-singleton.py` (`make iri-resolver-hygiene`) now fails the build if a third appears outside `crates/iri/src`.

**Base resolution is now spec-shaped.** Turtle 6.1 chains to RFC 3986 5.1: an in-document directive, else the caller-supplied base, else the retrieval IRI, else hard fail (5.1.4). `purrdf-rdf`, `purrdf-core`, the C ABI, wasm and Python are handed *bytes* and have no retrieval IRI, so they hard-fail with an actionable `iri-relative-no-base`. **Three** surfaces do have one, and all three are ones that opened the file themselves: `purrdf-slice` **derives** the RFC 8089 `file://` IRI (the workspace's single derivation), and `purrdf-shapes`' shape-union loader and `purrdf-cli` **consume** it rather than repeating it. Nothing filesystem-shaped crosses into `purrdf-iri` or `purrdf-rdf`, so the kernel ring-fence and the wasm build are untouched. An explicit `--base` always wins; stdin keeps the hard error.

**The reporters exact command now works:**

```
$ purrdf convert --to ntriples --from turtle empty-subject.ttl
<file:///.../empty-subject.ttl> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://example.org/test> .
```

**Other fixes found and made along the way**, each a real bug rather than a refactor:

- `@base <relative>` was rejected outright; Turtle 6.1 requires it to resolve against the in-scope base.
- `@prefix` resolved its namespace at **use** time, so `p:x` before and after a later `@base` denoted two different IRIs in one document. Now resolved at declaration time (Turtle 4.4). Same fix in ShEx, where prefixed-name `expand` never resolved at all.
- `validate_iri` was a colon test in RDF/XML, TriX and HexTuples, so `a:bLocal` — a `path-noscheme` relative reference — was accepted as absolute.
- The Turtle statement splitter feeding error recovery did not understand `#` comments, so a prose full stop split a statement and reported an **invented** syntax error over the documents real one.
- The format-totality test `continue`d past the tree syntaxes, which is why JSON-LDs dropped base survived. The match is now exhaustive and fails to compile if a codec is added without a spelling.
- Shapes-text seams took no base at all; threaded through to the C ABI, wasm and Python.

## Behaviour change to be aware of

A relative IRI with no base in scope is now a **hard error** instead of silently entering the graph. Documents that previously "worked" by emitting invalid N-Triples will now fail — correctly. Every in-tree consequence was fixed rather than suppressed; no test was deleted, weakened or `#[ignore]`d.

## Verification run

- `cargo test -p purrdf-rdf` — 35 binaries, 674 passed, 0 failed
- `cargo test -p purrdf-shex` — conformance **unmoved**: 1105/1105 validation, 425 ShExC / 420 ShExJ
- `cargo test -p purrdf-rdf --test jsonld_w3c_conformance` — to-RDF 73/73, compaction 13/13, before == after; **zero goldens moved** (`git diff` shows no change under `vectors/`, `generated/` or `tests/fixtures/`)
- `cargo test -p purrdf-sparql-algebra` — 342 passed
- `cargo clippy --all-targets -- -D warnings` — clean, no `#[allow]` added
- `cargo build -p purrdf-rdf --target wasm32-unknown-unknown` — clean
- `make capi-check` — `OK: committed purrdf.h matches the libpurrdf ABI surface`

## Was outstanding — now closed, plus one open policy decision

This list was written when the branch opened and every item on it has since landed. Recorded
here as closed rather than deleted, so the description can be checked against the tree:

- `purrdf-core` interner invariant — **done** (`crates/rdf-core/src/ir/absolute.rs`; a
  non-absolute IRI is unrepresentable in the IR from any ingress, enforced on the store-once
  miss path).
- Serializer base — **done** (`serialize_dataset_with`; `serialize_dataset_base_only` removed).
- `ParseOutcome.document_base` — **done**, with `BaseOrigin` provenance.
- Python / wasm / C-ABI surface parameters — **done** (the C ABI bump to `0.7.0` is one of them).
- W3C `IRI-resolution-*` fixtures — **done** (`crates/rdf/tests/corpus/w3c/{turtle,trig}/iri/`).
- Hygiene guard forbidding another resolver — **done**
  (`scripts/check-iri-resolver-singleton.py`, `make iri-resolver-hygiene`).
- The `crates/iri/tests/PROVENANCE.md` coverage claim — **done**.

Carried forward for a human decision: `scripts/conformance-baseline.json` raises the
`rdflib LSP drop-in gate` ledger budget **1 → 5**, and that file's own `_doc` says the budget may
only ever be edited DOWNWARD, with a raise justified in review. The justification is now recorded
in the file itself under `raise_justification`. In short: four vendored rdflib
`BatchAddGraph` tests name their throwaway terms `URIRef("a")` / `URIRef("b%d")`. Those tests were
already semantically broken — a graph holding such terms could never serialize as conformant
N-Triples — and passed only because nothing froze the graph. rdflib's `URIRef` is an unvalidated
`str` subclass; PurRDF's cannot be under `.goals`, which forbids inventing a base. **This budget
must not be lowered by weakening PurRDF's validation.** It needs an explicit review sign-off.

## Summary by CodeRabbit

* **New Features**
  * Added optional base IRI support across Python, WebAssembly, C, and CLI workflows.
  * Relative IRIs now resolve consistently across RDF, JSON-LD, ShEx, RIF, and SHACL inputs.
  * File inputs derive retrieval bases, with defined precedence for explicit and in-document bases.
  * Supported serializers emit base directives and compact relative IRIs while preserving supported statement data.
* **Bug Fixes**
  * Improved diagnostics for invalid or unresolved IRIs and undeclared shape-map references.
* **Breaking Changes**
  * Updated the C API to version 0.7.0 and replaced the legacy serialization entry point.
* **Tests**
  * Expanded RFC-compliant and cross-format resolution coverage.


paudley:
## What am I least confident about, and why?

**1. The `file://` derivation is the right call, but it is the one place I reversed myself, and reversals deserve suspicion.** My first plan hard-failed the CLI too. That would have "fixed" the report by improving an error message rather than making the command work, and an adversarial review of the plan caught it. I now think the derivation is correct — Turtle 6.1 chains to RFC 3986 5.1.3, the CLI genuinely has a retrieval IRI, and keeping derivation inside `purrdf-cli` leaves the wasm-clean core untouched. But my three original objections were not baseless: a graph converted from `/home/alice/x.ttl` and the same bytes from `/srv/x.ttl` now differ, and a user who publishes that output has leaked a local path. Every other RDF tool behaves this way, which is why I changed my mind, but "everyone does it" is weaker than the spec argument and I want that on the record.

**2. The serializer base is still a lie, and the PR is honest about it but the codebase is not yet.** `RdfSerializeRequest.base_iri` is still dropped. So there is no `@base`-emitting round trip: `<>` resolves correctly on the way in and cannot be written back out. That is half of "the serializers must agree with the parsers", and it is unfinished.

**3. The blast radius is measured only in-tree.** A workspace grep found ~450 `.ttl`/`.trig` files carrying a scheme-less or empty IRIREF with no in-document base. Most are parsed by harnesses that do supply one, and every in-tree test is green — but that number is an upper bound on a question I answered by running tests, not by reasoning about downstream consumers. `gmeow-ontology` cutover is incomplete per AGENTS.md 7, and I did not check it.

**4. Two of my instructions to sub-agents were factually wrong, and both were caught by the agent rather than by me.** I told one that `remove_base` "never verifies by re-resolution" — it did; only candidate *construction* was hand-rolled. I told another that JSON-LD ingress was already done, reasoning from a commit subject line rather than the diff, and then relayed that error onward. The work survived because the agents opened the files. That is the right outcome from the wrong process.

**Whether `make check`, `make test` and `make conformance` pass on the merged branch.** I ran narrow, per-crate gates throughout — `cargo test -p purrdf-rdf` 674 passed, `purrdf-shex` conformance unmoved at 1105/1105, JSON-LD 73/73 and 13/13 with zero goldens moved, clippy clean, wasm32 clean, and `make capi-check` confirming the committed header. I did not run the full gates. `make check` excludes rustdoc *and* conformance, so a green `make check` would not have told me either.

**Whether the interner invariant (Phase B) will hold without carve-outs.** I planned to validate absoluteness once at the `purrdf-core` store-once interner so a relative IRI becomes unrepresentable in the IR from any ingress. I never found out how many internally-synthesized IRIs — skolem constants, SPARQL internals — would need an exemption, and if the answer is "many", the invariant is weaker than advertised and the design needs revisiting.

**Whether the JSON-LD compact-spelling selection is right, or merely equal to what shipped.** Zero goldens moved, which is reassuring — but it also means the new selection layer was never observed to *disagree* with the old hand-rolled one. The property tests exercise it directly, so it is not untested; it is unconfirmed against the corpus.

**Whether the `--base` derivation interacts correctly with multi-source `convert`.** Each source should get its own retrieval IRI. I specified it and the tests pass, but I did not personally verify the multi-source ordering case.

paudley:
## The plan this was worked to

**Diagnosis.** `native-codec-iri-missing-value` is a downstream guard, not the bug. `<foo>` — a *non-empty* relative IRI — did not error at all: it was silently interned into the frozen IR and emitted as invalid N-Triples. Underneath sat **five hand-rolled reference resolvers** that disagreed, and seven diagnostic spellings for one condition.

**Base decision.** Turtle 6.1 chains to RFC 3986 5.1: in-document directive, else caller-supplied, else retrieval IRI, else hard fail (5.1.4). Library layers, C ABI, wasm and Python get bytes and have no retrieval IRI, so they hard-fail. The CLI has one and derives it, inside `purrdf-cli` only.

**Completeness contract.** Every acceptance criterion is an executable check on a production surface.

| Phase | Tasks | State |
|---|---|---|
| A. The layer | `BaseIri` / `BaseScope` / `ScopedBase` / `BaseOrigin`, `relativize`, `IriError::diagnostic_code()`, full RFC 3986 5.4.1+5.4.2 tables | **done** |
| B. IR invariant | absoluteness validated once at the `purrdf-core` interner; criterion bench | **outstanding** |
| C. Codecs | registry capability columns, symmetric codec seam, Turtle/TriG, RDF/XML, absolute-only syntaxes, JSON-LD/YAML-LD, ShEx, SPARQL; five duplicates deleted | **done** |
| D. Egress | serializer base made real, `sort_canonical` determinism guard, `emits_base` serialize-leg check, `ParseOutcome.document_base` | **outstanding** |
| E. Surfaces | Rust facade, Python, wasm, C ABI parameters | **partly** (shapes seams threaded to all three; the rest outstanding) |
| F. Regression | CLI tests from the reporter input **done**; W3C `IRI-resolution-*` fixtures, workspace blast-radius sweep, hygiene guard, docs — **outstanding** |

**Enhancements considered and declined — exactly one.** A `--base-from-file` / `--allow-relative-iris` escape hatch: the project forbids semantic optionality, and retrieval-IRI derivation is unconditional spec behaviour rather than a flag, so no toggle is needed. Everything else surfaced by review — the base *stack* rather than an `Option`, base provenance for diagnostics, `relativize` in the foundation crate, the symmetric seam, the IR invariant, table-driven policy, SPARQL, ShEx `expand`, `@prefix` timing, network-path references, the query/fragment corners, the bench, the hygiene guard, the W3C corpus — was folded in rather than declined.

**Review gates the plan passed.** Two `plan-reviewer` passes (SOLID/ETHOS, and an enhancement audit that restructured it around the base-stack and IR-invariant items) plus a completion adversary, which returned `PASS` at pass 2/3 after three findings were fixed: the CLI hard-fail that would have met the headline with an error, an unmapped Rust-facade surface, and a T17/T23 contradiction where the serialize-leg check would have broken `--base ... --to ntriples`.

## Honest status

Phases A and C plus the CLI derivation and regression tests are landed here. B, D, most of E, and most of F are not. The PR body lists them. Nothing was deferred to a follow-up issue and `.deficiencies` is untouched — this is simply where the work reached, and it is stated rather than implied.

paudley:
## Squash-merge subject and footers — required to keep the BREAKING markers

**Why this is needed.** `make changelog` runs `git-cliff --output CHANGELOG.md`, which
regenerates the whole file from commit **subjects**. `cliff.toml` renders one line per
commit (`commit.message | split(pat="\n") | first`), sets `footer = ""`, is given no
`--prepend`, and declares no keep-region or fence. There is therefore no mechanism by
which hand-written release-note prose survives a regeneration.

This branch carries **four breaking changes**, and none of its commits carries a
`!` / `BREAKING CHANGE:` marker. History rewriting is not an option here, so the marker
has to ride the **squash-merge commit**. Without it, `cliff.toml`'s own rule — *"the
changelog is the only place a consumer can see that a bump carries [a breaking change] —
it must never be silent"* — is violated for all four.

A pre-flight now blocks the destruction itself: `make changelog` refuses while
`CHANGELOG.md`'s `[Unreleased]` section holds hand-authored prose
(`scripts/check-changelog-handwritten.py`; `CHANGELOG_ALLOW_REGENERATE=1` to override
once the section has been preserved).

### Use this subject verbatim

```
fix(iri,rdf,cli,capi,shex)!: resolve relative IRI references through one base-aware layer
```

The `!` is what git-cliff reads. Dropping it, or dropping a scope, silently drops the
`**BREAKING**` marker for that scope.

### Use these footers verbatim, in the squash body

```
BREAKING CHANGE: a relative IRI reference with no base IRI in scope is a hard error
(iri-relative-no-base) instead of being interned verbatim. Only purrdf-cli has a
retrieval IRI, deriving each filesystem input's RFC 8089 file:// IRI; the library,
wasm, C ABI, Python and CLI stdin are handed bytes and hard-fail per RFC 3986 5.1.4.
Callers on those surfaces must give the document a base directive or pass one to the
API. N-Triples, N-Quads, TriX and HexTuples report iri-not-absolute-by-grammar instead,
which supplying a base cannot fix.

BREAKING CHANGE: the C ABI moves 0.6.0 to 0.7.0. purrdf_shacl_validate_to_sarif and
purrdf_shacl_entail_to_ntriples each gained a shapes_base_iri parameter MID-LIST,
between shapes_ttl and data_nt, and purrdf_serialize_jsonld_configured gained base_iri
after media_type. A host compiled against 0.6.x and run against 0.7.0 without
recompiling passes data_nt into the shapes_base_iri slot and its PurrdfBuffer ** into
data_nt, which the boundary reads as a NUL-terminated C string. Every C host must
recompile against the new purrdf.h.

BREAKING CHANGE: a ShEx shape map naming a shape label the schema does not declare, and
START against a schema declaring no start shape, are now a hard refusal
(ShexError::UnknownShape; CLI exit 1) instead of a "status":"nonconformant" result at
exit 0. Scripts that parsed the JSON result and ignored the exit code will now see a
failure where they previously saw a definite negative about the data.

BREAKING CHANGE: purrdf-cli refuses --base by name, as a usage error (exit 2), when
neither leg of the operation can spend it, instead of accepting it and never reading it.
convert --from ntriples --to ntriples --base <IRI> previously exited 0 having done
nothing. A base any leg can spend is still honoured. Scripts passing --base
unconditionally across format pairs must drop it on the pairs that cannot use it.
```

### What this does and does not preserve

It preserves the **marker** — each break will render as a `**BREAKING**` bullet after a
regeneration. It does **not** preserve the prose: git-cliff renders only the subject
line, so the four detailed entries currently in `CHANGELOG.md` would still need
re-authoring after any `make changelog`. That limitation is repo-wide, not specific to
this branch: `[Unreleased]` currently holds 72 hand-authored entries, 37 of them
`**BREAKING**`, none reproducible from a commit subject.

### Also note

The **Still outstanding** section of the PR description is now stale: the W3C
`IRI-resolution-*` fixtures, the hygiene guard forbidding a sixth resolver outside
`purrdf-iri`, and the false coverage claim in `crates/iri/tests/PROVENANCE.md` are all
done.

PR #224, opened 2026-09-01
feat(sparql-eval,cli,python,purrdf,rdf-core)!: bind path witnesses as property functions, registered over the entailment closure

### LINKED ISSUE #194: Path binding (step number / path id / intermediate nodes) as a property function
# Path binding (step number / path id / intermediate nodes) as a property function

## Summary

Neither SPARQL 1.1 nor SPARQL 1.2 property paths expose the intermediate nodes,
step numbers, or a path identifier for a matched path — only the endpoints.
Virtuoso's `OPTION(TRANSITIVE ...)` extension is the de-facto reference for what
callers want here. Rather than extending the path grammar itself, this is
delivered as a property function (depends on item 3), which keeps the core path
grammar untouched and puts path-binding behind an explicit, caller-invoked
relation.

## Acceptance sketch

- A property function exposes, for a matched property-path binding: the
  sequence of intermediate nodes, a step number per hop, and a stable path
  identifier for the match.
- Behavior is validated against at least one Virtuoso `OPTION(TRANSITIVE ...)`
  reference case (same path pattern, equivalent binding semantics).
- `cargo test -p purrdf-sparql-eval` passes, including a path-binding
  property-function test over a multi-hop transitive path.

Effort: M.

paudley:
Closes #194

## What this is

Path binding delivered on the **already-shipped** property-function seam. The registry (roadmap item 3) is closed and present in `crates/sparql-eval/src/property_fn.rs`; this branch adds its first path consumer and proves the seam serves it. `crates/sparql-eval/src/property_fn.rs` and the whole of `crates/sparql-algebra/` are **byte-unchanged** — asserted as gates, not claimed in prose.

## The object is a WALK, not a node sequence

A path binding is a witness `n0 -e1-> n1 -e2-> ... -ek-> nk`, where each `ei` is a traversed **statement**. A node-only model silently erases derivations: with an alternation of directed predicates, two different statements can join the same node pair. In RDF 1.2 a statement *is* a term, so every traversed hop is emitted as a `TermValue::Triple` in asserted orientation and joins straight back into the dataset by ordinary BGP.

Row shape (`PfArity::new(1, 6)`), one row per hop:

```sparql
?start <caller-iri> ( ?end ?pathId ?len ?step ?node ?edge )
```

`?step` and `?len` are `xsd:integer`. The node sequence is `?start` followed by `?node ORDER BY ?step`.

## Surface

- `PathStep` — ordered alternation of directed predicates; hard-fails on an empty alternation, a non-IRI predicate, or a duplicated `(predicate, direction)` pair.
- `PathLimits` — `min_hops`, `max_hops`, `max_paths_per_seed`, `max_expansions_per_invocation`; checked constructor, **no `Default`**. `min_hops == 0` is rejected: a zero-hop path has no witness. `max_hops` is capped at `MAX_HOPS_CAP` because an unbounded depth is a stack overflow, which is an *abort* and escapes `open_contained`s panic containment.
- `PathGraph::from_dataset` — snapshots a **multigraph** into dense `u32` id space (nodes sorted by `TermValue::Ord`, neighbour lists sorted by `(to, statement)`, never deduplicated across distinct witnesses). A declared predicate the dataset carries no in-scope quad for — interned or not — contributes ZERO EDGES rather than failing, exactly as the core grammar's `p|q` does not fail when `q` matches nothing. (This reverses an earlier revision of this branch, which hard-failed on an un-interned predicate; refusing there keyed a failure on which dataset happened to be in front of a fixed step vocabulary.) Exposes `snapshot_fingerprint`.
- `PathWitnessRelation` (every simple-prefix walk) and `ShortestPathWitnessRelation` (one shortest witness per reachable endpoint pair) — **two values, not one type with a mode switch**.
- `TermValue::canonical_bytes` in `purrdf-core` — an injective, self-delimiting encoding joining `Ord` and `Hash` as the third hand-written encoding of the type, proven injective by proptest.

## Termination

**Simple-prefix walks**: every node distinct except that the final node may repeat an earlier one, closing a cycle and terminating there. Finite on cyclic input twice over (the proper prefix is simple, so depth is bounded by node count, and independently by `max_hops`). Traversal is an **explicit heap stack, never recursion**.

This choice makes the endpoint projection *equal* `p+` reachability rather than silently omitting self-reachability: a node reaches itself iff it lies on a cycle. Asserted against the core grammar in the same engine.

## Determinism

The identifier is **content-derived, never a counter**: the full 32-byte SHA-256 of `PATH_ID_DOMAIN_V1 || canonical_bytes(n0) || (canonical_bytes(e_i) || canonical_bytes(n_i))* || hop_count`, folded incrementally down the traversal stack over per-edge bytes precomputed at snapshot time. The digest is **not truncated** — the identifier is a grouping key in query answers, so a collision is a wrong answer.

It deliberately excludes the graph selector (`GraphMatch<D::Id>` is parameterized by dataset-local ids, so digesting it would make the identifier a function of intern order), the step definition, and the limits (the same walk must get the same identifier under different `max_hops`). Stable across runs, processes, ceilings, read-subsets, and independently built datasets — except through blank nodes, where the label and scope depend on how the data was parsed. All three cases are asserted, not documented.

## Correctness details worth reviewing

- **Ceiling accounting.** The cursor applies its own equality filter on *every* bound position before emitting, so every row it emits agrees with its bound arguments. That is what the trait ceiling contract requires and what closes the short-answer-labelled-complete channel that would otherwise open under `LIMIT` with a bound `?end`.
- **Honest cardinality.** `rows_per_invocation` is a real function of the mode (free vs bound start, bound `?step`, bound `?pathId`), all `saturating_*` so no bound can wrap to a dishonest zero. Swept by test.
- **Pushdown**, not just filtering: bound `?start` seeds one node; bound `?len` bounds the depth exactly; bound `?end` prunes with an exact reverse-reachability-by-depth table cached behind a `OnceLock`.
- **Guards hard-fail, never truncate.** Both `max_paths_per_seed` and `max_expansions_per_invocation` return `Err`. `max_expansions_per_invocation` exists because the engine charges the governor per invocation and per *emitted* row only, so explored-but-unemitted prefixes are otherwise unbounded, uncharged work inside one `next()`. Both are resource guards over work actually performed, so the row-ceiling licence legitimately changes whether they fire — stated in the type docs, in the error messages, and pinned by a test asserting both the `Err` and the `LIMIT 1` `Ok` branch.

## Mints nothing

No vocabulary IRI, no default registration. The relation is reachable only under a caller-supplied IRI the caller both registers and lists in `ParserOptions::property_fn_iris`. An IRI configured but unregistered hard-fails with the IRI in the message, never a zero-row `Ok`.

## Acceptance suite

`crates/sparql-eval/tests/path_witness_property_function.rs` — twelve tests, every one driving SPARQL **query text** through `NativeSparqlEngine::query_with_options_view`, all with exact expected vectors:

| Check | What it pins |
|---|---|
| A1 | Exact hop-by-hop rows for a 3-edge chain, plus the `GROUP_CONCAT` reconstruction |
| A2 | Virtuoso `OPTION(TRANSITIVE ...)` reference — the expected tuples are **observed output of Virtuoso OSE 07.20.3243**, re-executed and matched row for row (see *Virtuoso validation* below), not derived from the documentation |
| A5 | Unregistered IRI hard-fails, is not a zero-row `Ok` |
| A7 | Cycles terminate, exact finite walk set |
| A8 | RDF 1.2 triple term as an intermediate node |
| A10 | One registry serving the path consumer *and* a scored top-k consumer, seam unchanged |
| A12 | Guard `Err` without `LIMIT`, `Ok` with `LIMIT 1` |
| A15 | Two statements joining one node pair are two witnesses, two `?edge`s, two `?pathId`s |
| A16 | Endpoint projection equals `p+`; the `{2,2}` relationship pinned |
| A17 | Identifier stability, including the blank-node boundary |
| A18 | `?edge` joins to its RDF 1.2 annotation via `rdf:reifies` |
| A19 | Expansion budget fails a fruitless search rather than answering empty |

## Gates run

- `cargo fmt --all`
- `cargo clippy -p purrdf-sparql-eval --all-targets` — warning-free
- `cargo clippy -p purrdf-core --all-targets` — warning-free
- `cargo test -p purrdf-sparql-eval` — 0 failed (incl. 3 doctests, one of them the new host-wiring example)
- `cargo test -p purrdf-core` — 0 failed
- `cargo build -p purrdf-sparql-eval --target wasm32-unknown-unknown`
- `cargo build -p purrdf-core --target wasm32-unknown-unknown`
- `cargo doc -p purrdf-sparql-eval --no-deps` — no warnings
- `[ -z "$(git diff origin/main...HEAD --numstat -- crates/sparql-algebra/)" ]` — core path grammar untouched
- `[ -z "$(git diff origin/main...HEAD --numstat -- crates/sparql-eval/src/property_fn.rs)" ]` — seam untouched
- Deferral-keyword and issue-number scans over the diff — zero hits

`make check` / `make test` / `make conformance` are stage-3 gates and were not run here.

## Summary by CodeRabbit

* **New Features**
  * Added configurable path-relation property functions for CLI and Python queries and updates.
  * Supports directed walks, shortest paths, per-hop bindings, traversal limits, and RDF 1.2 statement terms.
  * Added deterministic canonical byte encoding for RDF term values, including nested triples and scoped values.
  * Added RDF 1.2 reifier and annotation traversal.
  * Added quad-aware query results and serialization-loss reporting.
  * Added SHA-3 built-in functions and named-graph construction support.
* **Documentation**
  * Added guidance and examples for path relations, traversal configuration, and quad results.
* **Tests**
  * Added comprehensive CLI, Python, traversal, serialization, and encoding coverage.

---

## Virtuoso validation — the issue's acceptance criterion, RUN

The issue requires validation against at least one Virtuoso `OPTION(TRANSITIVE ...)` reference case. **Virtuoso was executed.** Not quoted, not hand-derived.

Reference engine: `openlink/virtuoso-opensource-7`, reporting `OpenLink Virtuoso Server Version 07.20.3243-pthreads for Linux` — the exact version the test comments name. Driven over `isql`. Five runs, each reproducing what the branch had pinned:

| Run | Result |
|---|---|
| §9.32 SQL `select transitive` over the documented `knows` table | The documented 8-row table, same values, same order |
| §16.2.11 SPARQL `OPTION(TRANSITIVE …)` at `t_min(1)` (A2's fixture) | 4 rows, identical to the pinned vector — `path_id` values 1/0/2/2 included |
| The same at `t_min(0)` | 8 rows, including the identity row `ex:p1 → ex:p1` on path 0 |
| `t_shortest_only` over the diamond-with-shortcut (A24's fixture) | 3 rows, identical to the pinned vector, `path_id` 1/2/0 |
| 3-cycle without `t_distinct` (the module docs' non-termination claim) | No output; killed at 130s — the divergence claim reproduced |

Every empirical claim in `path_relation.rs` and the acceptance suite was therefore confirmed against the reference engine rather than trusted. The reproduction recipe is now recorded in the A2 doc block, so the transcription is checkable instead of a magic constant with a citation.

**What Virtuoso does NOT validate**, stated plainly: the harness still does not run Virtuoso — no test in this repo speaks to a database. What is pinned is a transcription of observed output, now independently re-observed. And on the 3-cycle this relation is deliberately *more* complete than the reference (it produces the cycle-closing walk Virtuoso omits, which is what keeps the endpoint projection equal to `p+`); that divergence is intentional and pinned by A16.

## FIXED: `--path-relation` under `--entailment` read the pre-closure view

A path relation was snapshotted from the pre-closure view while evaluation ran against the
materialized closure, so under a regime that derives quads beneath a step's predicate the walk
answered over a **strictly smaller edge set** than every other pattern in the same query — a short
answer, at a success exit, with no diagnostic. Measured, over
`ex:sub rdfs:subPropertyOf ex:p . ex:a ex:p ex:b . ex:b ex:sub ex:c .` at `--entailment rdfs`:
`ex:a ex:p+ ?end` answered `{b, c}` and the walk answered `{b}`. That is a wrong answer at exit 0,
and it broke A16's own invariant (the endpoint projection equals `p+`) in the one place nothing
downstream could notice, because the property-function seam hands a relation no dataset at
evaluation time.

**It is fixed here, by changing the ORDER rather than filtering the symptom.**
`purrdf::query_with_entailment` and `purrdf::query_with_entailment_governed` take a new
`relations: &ClosureRelations<'_>` argument (after `options`, before `governors`). The closure is
materialized first, the caller's rebuilder derives the relations from THAT dataset, and the query
is re-prepared and evaluated against it — re-prepared because a plan carries the identity of the
registry it was prepared against and refuses to run against any other (`RegistryId`).

* **BREAKING (Rust):** callers whose relations are dataset-independent — an in-memory table, an
  empty registry — pass `ClosureRelations::NONE` and keep their present behaviour byte for byte.
  The C ABI and WebAssembly surfaces register no relation and pass `NONE`.
* **BREAKING (behaviour, CLI + Python):** `--path-relation` beside `--entailment`, and
  `path_relations` / `relations_from_graph` beside `entailment=`, now read the closure. Both
  surfaces are unchanged in SHAPE and supply the rebuilder themselves.

**One pairing is refused rather than answered**, and it is the only one: a rebuilder combined with
an OWL Direct-Semantics run whose restricted chase MINTED existential witnesses. Those are blank
nodes that are not terms of the scoping graph, and `restrict_witness_bindings` treats a
property-function call as the identity on grounds that re-deriving a relation from the closure
would falsify — a walk over the chase's edges could hand one back as an observable binding. The
refusal carries the stable code `reasoning-closure-relation-witness` (CLI exit 2, Python
`ValueError`) and names the regimes that accept the pairing.

**A refusal is a claim, so both neighbours are executed beside it** in all three hosts' tests: the
same ontology under the same regime with no relation still returns the combined approach's certain
answer, and an `owl-direct` run over a TBox outside the Horn fragment — which mints nothing — still
walks its relation. The refusal is keyed on witnesses actually minted, never on the regime's name.

Tests: `crates/purrdf/tests/closure_relations.rs` (5 cases at the public API, including that
`NONE` still produces the old smaller answer, so the fix cannot pass by having widened the
relation), `a_path_relation_under_entailment_walks_the_closure` and
`a_path_relation_is_refused_only_where_the_chase_minted_a_witness` in `crates/cli/tests/query_cli.rs`,
and three cases in `bindings/python/tests/test_property_functions.py`.

The test that pinned the bug (`a_path_relation_under_entailment_answers_over_the_pre_closure_edges`)
is gone; the one that replaced it asserts the corrected behaviour and compares the two answers to
each other rather than to a constant.

## Merged `origin/main`

`main`'s RFC 3986 base-resolution layer (#223 / #222) is merged in, with its five breaking changes.
The only conflict was `docs/CONFORMANCE.md`, resolved by recomputing rather than picking a side:
the Python binding suite is **768** (701 at the merge base + 35 from main + 32 here) and the rdflib
gate is main's 81 pass / 5 ledgered. `python3 scripts/conformance-matrix.py --write-doc`
regenerated the block and the hand-written scoreboard row was updated to match.

## CodeQL

The failing CodeQL check was a real alert, not job setup: `rust/cleartext-logging` (high) on a
`panic!` in `crates/sparql-eval/tests/path_witness_property_function.rs` that formatted
`exhausted.tripped` — flagged because the taint source is `self.certificate.tripped` in
`governor/lift.rs` and the rule keys on the NAME `certificate`. It is a false positive over a
resource-accounting enum in a test, and it is fixed by not giving the gate the shape: the `else`
arm rules out exactly two of three variants, so naming them is the whole of what a `{:?}` carried.


paudley:
## What are you least confident about in this work, and why?

**1. The Virtuoso reference case (A2) is a documentation-derived oracle, not a live one.** No Virtuoso instance was run. The test mirrors Virtuoso Documentation §16.2.11 (Transitivity in SPARQL) and §9.32 (Transitivity in SQL) — §16.2.11 explicitly defers the meaning of `step_no` and `path_id` to §9.32, which works them out over a concrete `knows` table with an 8-row result. Those passages are quoted verbatim in the test comment and the expected tuples are derived from them term by term, so the derivation is auditable. But it is a derivation from prose, not an observed run, and if the documentation is imprecise about `T_DISTINCT` interaction the derivation inherits that imprecision. This is the single weakest evidentiary link in the branch.

**2. The `{n,m}` divergence boundary is asserted at one point, not characterised in general.** A16(b) pins that at `k = 2` over a 2-cycle the relation and the core grammar coincide (`a → b → a` is simple-prefix). The module documents that interior-revisiting walks at higher `k` diverge. I have not proved the general characterisation of exactly which `p{n,m}` bindings the relation does and does not witness — I pinned the coincidence point so a future move is visible, which is weaker than a theorem.

**3. `rows_per_invocation` for `ShortestPathWitnessRelation` in the free-start mode.** The bound is `node_count * min(node_count, max_paths_per_seed) * max_hops`, argued sound because BFS records each node once so there is at most one witness per `(seed, end)` pair. The mode sweep test confirms `emitted <= declared` on the fixtures used, but those fixtures are small; the argument is analytic and I would want an adversarial dense fixture before calling it proven.

**4. Performance is reasoned, not measured.** Per project policy I did not benchmark. The dense-`u32` traversal, the `OnceLock` reverse-adjacency cache, and the incremental digest fold are all allocation/complexity arguments from reading, and no criterion bench was added because nothing here claims a speedup over a prior implementation.

**5. `max_expansions_per_invocation` has no principled default because there is no default.** Callers must pick a number. A too-low value turns a legitimate query into an error. That is the correct fail-fast behaviour, but it puts a genuine tuning burden on the host and I have no guidance to offer beyond the error message.

**1. Whether the branch is green under the full repo gates.** Stage 1 forbids `make check` / `make test` / `make conformance`. I ran targeted `cargo fmt --all`, `clippy` on both touched crates, `cargo test -p purrdf-sparql-eval` and `-p purrdf-core`, `cargo doc`, and per-crate wasm builds — all clean. What I have NOT run is the workspace-wide test sweep, the hygiene lane, `make wasm` itself (only the equivalent cargo build), or the conformance suites. `TermValue::canonical_bytes` touches `purrdf-core`, which everything depends on; it is additive and no existing behaviour changed, but that reasoning is not a substitute for the sweep. **Stage 3 must run the full gates.**

**2. Whether `PathGraph::snapshot_fingerprint` is the right host-facing surface.** It exposes `stats_fingerprint`, `term_count`, and node/edge counts so a host can assert the relation is being queried against the dataset it was built from. The seam gives a relation no dataset at evaluation time, so this genuinely cannot be enforced engine-side — it is documented loudly as a precondition and the fingerprint is the surface a host checks. Whether real hosts will actually check it, or whether something stronger is warranted, is unknown to me.

**3. Whether the multigraph snapshot is affordable at real scale.** It materialises every edge for the declared predicates as dense ids plus one `TermValue` per distinct node and per distinct statement, and precomputes per-edge digest bytes. That is proportional to the predicates edge count, which is honest and bounded — but on a large graph with a hot predicate it is a real allocation, and on wasm32 (32-bit address space) I do not know where the practical ceiling sits.

**4. The self-loop merge rule.** Neighbour lists keep every distinct witness, but entries identical in *both* `to` and `statement` are merged. Such a pair can only arise from a self-loop under a step declaring one predicate in both directions — the same statement traversed to the same node, recorded twice. Merging is required for the identifier to stay injective over walks, which `rows_per_invocation`s `pathId` branch depends on. I believe that is the only way such a duplicate arises; I have not exhaustively proved it.

**5. Whether the two other planned consumers will actually fit.** A10 registers a path relation and a scored top-k `MemoryRelation` in one registry and joins them in one query, and the seam file is byte-unchanged — that is real evidence the seam admits a differently-shaped consumer. It is not evidence that a full-text index or an ANN search will fit without friction. The scored-retrieval stand-in is a table, not an index.

PR #225, opened 2026-09-01
feat(cdt,sparql-algebra,sparql-eval,rdf,rdf-core,xsd,conformance)!: implement SEP-0009 SPARQL composite datatypes (#196)

### LINKED ISSUE #196: SPARQL CDTs (SEP-0009: cdt:List/cdt:Map, FOLD/UNFOLD)
# SPARQL CDTs (SEP-0009: cdt:List/cdt:Map, FOLD/UNFOLD)

## Summary

Composite Datatypes (`cdt:List`, `cdt:Map`) plus the `FOLD`/`UNFOLD` operators,
per sparql-dev SEP-0009. This originated as an AWS specification and ships with
a test suite; Jena already implements it. One normative detail called out
explicitly: the blank-node-label relabeling rule that applies when loading data
inside a CDT literal — this needs to be respected by whatever parsing/loading
path produces `cdt:List`/`cdt:Map` values, consistent with the project's
existing blank-node-label discipline elsewhere in the codebase.

## Spec references

- sparql-dev SEP-0009: https://github.com/w3c/sparql-dev/blob/main/SEP/SEP-0009/sep-0009.md

## Acceptance sketch

- `cdt:List` and `cdt:Map` datatypes parse, evaluate in FILTER/BIND expressions,
  and serialize correctly.
- `FOLD` and `UNFOLD` operators are implemented and covered by the SEP-0009 test
  suite (or an equivalent first-party corpus derived from it).
- The blank-node-label relabeling rule for data loaded inside a CDT literal is
  implemented and covered by a dedicated test (a CDT literal containing a blank
  node round-trips with legal, non-conflating labels).
- `cargo test -p purrdf-sparql-eval` and `cargo test -p purrdf-sparql-algebra`
  pass, including CDT parse/eval/FOLD/UNFOLD cases.

Effort: M.

paudley:
Implements SEP-0009 SPARQL Composite Datatypes end to end: a new closed-leaf
crate carrying the value space, the full function library, the `FOLD` aggregate,
the `UNFOLD` graph pattern, composite-aware ordering, blank-node identity inside
composite lexical forms, and the vendored upstream conformance corpus wired in
and passing.

## The crate

`purrdf-cdt` is a closed leaf over `purrdf-iri` + `purrdf-xsd` only — no
dependency on the kernel, and `wasm32-unknown-unknown`-clean like every other
release crate, which takes the published set from 18 to 19. It carries the
SEP-0009 value space, the lexical parser and renderer, the value relations, and
the function library: the spec's fifteen functions (`cdt:List`, `concat`,
`contains`, `get`, `head`, `tail`, `reverse`, `size`, `subseq`, `cdt:Map`,
`containsKey`, `keys`, `merge`, `put`, `remove`). The set is CLOSED — there is
no registry to configure and no way for a caller to shadow a spec function — so
the same query means the same thing on every host, and a configured extension
namespace cannot capture a CDT name.

Nesting is fully ITERATIVE, under three bounds (depth 64, 2^20 elements,
64 MiB). `crates/cdt/tests/adversarial.rs` drives a million-deep input on a
256 KiB stack and requires a typed error at exit status zero. This is not
defensive styling: a recursive walk here would overflow the stack, and a stack
overflow ABORTS the process rather than raising, so no caller could contain it.
The programmatic constructors enforce the same three bounds the parser does, so
the limits are an invariant of the value rather than of one code path.

PurRDF mints nothing here. The SEP-0009 namespace is the spec's own fixed
string, recognized and never invented. IRIs inside a composite lexical form must
be absolute: a composite's contents are its own and are not resolved against the
enclosing document's base, which is pinned from both sides — a relative
reference inside a composite is refused whether or not the document declares a
base.

## SPARQL integration

The fifteen functions are resolved at parse time — with arity enforced there,
before evaluation — and dispatched by the evaluator; a test walks the whole
registry so a future member cannot be added to `purrdf-cdt` and left silently
unreachable from SPARQL.

`FOLD` rides the aggregate ACCUMULATOR seam as a keyword aggregate carrying its
own `ORDER BY`, which makes it the engine's first order-dependent aggregate. It
does NOT ride the `AGG(<iri>, …)` surface: that production consumes a literal
`AGG`, `(`, an IRI and a `,`, so a keyword-spelled aggregate can never reach it.

`UNFOLD` is its own `GraphPatternNotTriples` alternative over an expression —
structurally where `LATERAL` sits — reusing the row-expansion driver. It does
NOT ride the property-function seam: that registry keys on a BGP predicate IRI,
no `UNFOLD` predicate IRI exists to key on, and inventing one would be minting a
vocabulary, which this repository forbids.

## Ordering, and one real defect it uncovered

`ORDER BY`, `MIN`, `MAX` and `FOLD`'s own `ORDER BY` now order composite
literals by the value they denote, through ONE projection and ONE total order
rather than the six hand-written implementations of section 15.1 that were there
before.

That collapse turned up a defect and fixed it. The obvious reading of section
15.1 — "by value, else by a deterministic syntactic key" — is INTRANSITIVE:
`"9"^^xsd:double` < `"P1D"^^xsd:duration` < `"8"^^xsd:float` <
`"9"^^xsd:double`, and Rust's sorts may PANIC on a comparator like that. The
cycle is broken by ranking the comparability class before the value, plus a new
exact `purrdf_xsd::value_total_cmp` for the numeric tower, whose promotion-based
sibling is itself intransitive across the exact/IEEE boundary. `MEDIAN` and
`PERCENTILE` sort their numeric series through the same comparator and were the
two sites the collapse had missed; they are folded in here too. The regression
test checks transitivity of the RELATION rather than of a sorted sequence — a
sort can emit a plausible order from an intransitive comparator, so the sequence
proves nothing — and it pins both reachable cycles so nobody "simplifies" the
fix away. The exported total order over composites is purely SYNTACTIC for this
reason.

## Blank nodes

A blank-node label written inside a composite lexical form binds through the
SAME ingress rule as a bare `_:` token, on every codec path (Turtle/TriG/
N-Triples/N-Quads, RDF/XML, TriX, HexTuples, JSON-LD, and both GTS import
paths), through one choke point — `RdfDatasetBuilder::intern_literal_bound`.
That agreement is what makes `_:b` written as a subject and `_:b` written inside
a `cdt:List` in the same document one node, and what keeps the same label in two
different documents two different nodes. Composite-embedded blanks also
participate in canonicalization and in skolemization.

SEP-0009 defines no canonical form, so canonical rendering applies ONLY to
literals PurRDF itself mints. C0.1 byte fidelity is untouched, with one carve-out
for blank labels, which ingress must rewrite in order to bind them.

## Conformance

The vendored upstream suite (`vectors/sparql-cdt/`, `awslabs/SPARQL-CDTs` at
commit `e0a7465`, 658 cases across six groups) is run and reports its own
scoreboard row. It had been vendored, frozen and inventoried but never
EXECUTED — so until now every claim about this engine's CDT behavior rested on
tests written beside the implementation by the same hand, which grade the engine
against itself.

Wiring it up immediately earned its keep: 646 of 658 passed and the remaining 12
were real defects, in three classes, all fixed at root cause rather than
ledgered. The final row is 658 / 658, 0 ledgered.

1. The harness merge was a text concatenation, not an RDF merge. It serialized
   each `qt:data` file to N-Quads, joined the TEXTS, and re-parsed the join as
   one document — so every source's `_:b` collapsed onto one blank node.
   Combining documents is an RDF 1.1 4.1 merge; the sources are standardized
   apart first. Each source is now parsed on its own and merged over the OWNED
   model under its own `BlankScope`. This was mis-grading every multi-file case
   in every suite, not just the CDT ones, and merging over the owned model also
   carries the RDF 1.2 statement layer, which the N-Quads round trip could only
   flatten.
2. Nothing gave query text its own blank scope. A query-authored
   `"[_:b, 42]"^^cdt:List` kept its bytes verbatim, so the scratch interner —
   which promotes any value found in the dataset — turned the query's `_:b` INTO
   the data's. Note the trap: fix 1 alone turns these four cases green, because
   the harness then leaves scope 0 unused. That is a harness accident, not a
   fix, and a host loading a single Turtle file would still collide. So the
   engine is fixed too, at the one query-text-to-value boundary, and the engine
   fix was confirmed load-bearing by reverting it alone.
3. `qt:constructDataFile` was unmodeled, so three cases silently loaded zero
   quads and failed as empty-result mismatches — invisible to the loader's
   silent-skip guard, because the entries are fully described. The action is now
   modeled end to end, with the graph genuinely written and re-parsed, since the
   syntax round trip is the point of those cases.

Two loader defects found along the way are fixed for every suite, not only this
one: `mf:include` is now followed, so an aggregator manifest can no longer load
zero cases and report green, and each manifest's base is derived from its own
path so two manifests cannot mint the same case IRI.

## The divergence, stated rather than argued

PurRDF's reader admits two element forms the published SEP-0009 lexical space
does not: an RDF 1.2 triple term and a directional language-tagged literal (as
list elements and map values; the map-key production is untouched). The reason
is that refusing a term type the data model defines, inside a container the data
model also defines, is not an admissible outcome for this toolkit — a `cdt:List`
that cannot hold half the term model is not a carrier. The extended forms are
emitted only for values the published grammar cannot express at all.

A conformant SEP-0009 reader handed one of those literals will call it
ill-formed. No argument in this repository changes that, and it is now recorded
in `docs/CONFORMANCE.md`, in `CHANGELOG.md`, in the conformance-matrix row, in
the book's SPARQL chapter and in the crate's own documentation, instead of
living only in a code comment.

The mitigation is executed, not asserted.
`crates/sparql-conformance/tests/cdt_lexical_divergence.rs` tokenizes all 2,885
corpus files this workspace ships with the production lexers, resolves each
file's own prefix bindings, recovers all 959 `cdt:List`/`cdt:Map` literals, and
grades every one. All land inside the published space, so no vector anywhere
here can distinguish PurRDF's reader from a strict one. Every count is pinned as
an exact equality, never `>=`, so the scan cannot pass by finding nothing, and a
positive control proves the grader can answer `PurrdfSuperset`.

That scan had a blind spot, and this branch closes it: a file its lexer refused
was dropped by a bare `continue`. Thirteen files were being dropped that way,
and every one is a NEGATIVE-SYNTAX vector — precisely the file class the
mitigation was specified against. The refusals are now collected, counted,
pinned, and text-checked for the SEP-0009 namespace.

The scan's scope is also stated honestly. The mitigation was originally
specified as "no upstream negative-syntax entry contains the new production".
Stated that way it is VACUOUS: all 658 entries upstream declares are
`mf:QueryEvaluationTest`, so upstream ships no negative-syntax CDT vector at all
and nothing it publishes could observe the divergence either way. The broader
claim the scan does establish is the one worth having. What no test here does is
compare PurRDF against another SEP-0009 implementation.

A second divergence, in the strict direction, is recorded alongside it: an
ill-formed composite literal refuses the whole document, where an ill-formed
`"abc"^^xsd:integer` does not. That is deliberate — the embedded-blank scanner
is lexical, so admitting an unparseable form opaquely would leave a half-bound
blank-node scope — but it is a real asymmetry and is now written down rather
than left for a user to discover. In query text the same literal parses, as an
ill-typed literal should.

## Documentation

The book's SPARQL chapter walked SEP-0002, SEP-0006, SEP-0007 and SEP-0008 in
detail and said nothing about SEP-0009. It now carries the namespace, both
lexical forms and their key restrictions, the function table with its arities
and 1-based indices, both new productions with accepted spellings, the four
`FOLD` behaviours that differ from every other aggregate, the `UNFOLD` edge that
decides whether an enclosing `OPTIONAL` does what a reader expects (a
well-formed EMPTY composite yields zero rows; a non-composite operand passes the
row through unchanged), the syntactic-order rationale, and the divergence.

Closes #196

BREAKING CHANGE: `purrdf_sparql_algebra::GraphPattern` gains an `Unfold`
variant. Consumers matching the enum exhaustively must add an arm; there is no
Cargo feature to opt out, because CDT is unconditional like every other part of
the engine.

BREAKING CHANGE: `purrdf_sparql_algebra::Function` gains a `Cdt(CdtCall)`
variant. Consumers matching the enum exhaustively must add an arm.

BREAKING CHANGE: `purrdf_sparql_algebra::AggregateExpression::into_parts` now
returns the `AggregateParts` alias, a five-tuple carrying the aggregate's own
`ORDER BY` clause, rather than the previous four-tuple. Callers destructuring
the result must add the new element.


paudley:
## What I am least confident about in this work, and why

**1. The two lexical-space supersets are the highest-risk decision on the branch.** SEP-0009 admits neither triple terms nor directional literals as CDT elements. I extended the lexical space for both, on the reasoning that refusing an RDF 1.2 term type contradicts the project goals, and that emitting these forms only when such a term is present means we never differ from the spec on any value the spec can express. But this is the one place where PurRDF deliberately parts company with a spec it otherwise implements exactly, and my confidence rests on an argument rather than on a test against another implementation. A conformant reader given one of our extended literals will call it ill-formed. The mitigation I specified — a mechanical scan proving no upstream negative-syntax entry contains the new production — has not been run yet, because the harness is not wired.

**2. Nothing has run against the 658-entry corpus.** The corpus is vendored and frozen; not one test has been executed against an implementation. Every claim in the plan about FOLD, UNFOLD, function, ordering and blank-node semantics is currently derived from the spec text and the corpus queries I read, not from a passing run. That is the normal stage-1 position, but it means the design confidence and the evidence are not yet the same thing.

**3. The blank-node work reaches deeper than the issue suggests.** The issue frames the relabeling rule as a parsing concern. Reading the code, it is not: `CanonicalRelabeler` implements only `map_blank`/`map_iri`, so literals pass through untouched. Without a literal hook, `canonical_relabel` would emit `c14n0` outside a literal and the original `_:b1` inside — a dangling reference — and RDFC-1.0 would hash CDT-embedded blanks as opaque input labels, so `datasets_isomorphic` would return false for isomorphic datasets. I am confident the defect is real. I am less confident about the blast radius of the fix, because it modifies canonicalization, which many things depend on.

**4. The FOLD combine path.** I specified that `combine` must downcast and append element vectors, never finish-then-step, because `finish()` returns a `cdt:List` literal and re-stepping it would nest a finished list as one element. That follows from the accumulator docs, but it is a silent-wrong-answer failure mode rather than a crash, and it only manifests when a group is large enough to be split into parallel chunks. It needs the chunk-boundary test I specified and does not yet have.

## What you do not know but should

**The issue estimates Effort: M. That estimate is wrong by a wide margin.** This is a new workspace crate, a `parse_by_iri` migration across five crates, a refactor collapsing four duplicated comparison sites, two new SPARQL grammar constructs, three tasks inside the frozen IR and its canonicalization, four host surfaces, and a 658-test corpus. Two of fourteen chunks are done. Nothing about that is a reason to do less, but planning around M will mislead you.

**A roadmap document sent me down a false path.** The umbrella roadmap states the evaluator hard-errors on custom aggregates and cites a line number. Both are stale: `modifier.rs:867` dispatches to a working registry with ten shipped aggregates, and the cited error is an unreachable defensive arm that says so in its own comment. If other issues were scoped off that same sentence, they are scoped off a false premise.

**A real defect exists on the branch and cannot be fixed from here.** Commit `6a5be4e5` has a garbled message from a heredoc that failed to expand through a nested shell wrapper. Amend and reset are both blocked by policy, and there is no non-rewriting way to correct a commit message. It will not reach main given squash-merge, but the squash message must be written deliberately.

**This work will change `purrdf-core` and the algebra**, so the full suite is mandatory at stage 3, not optional. The narrow per-crate gates used at stage 1 cannot see the regressions that matter here.

**One upstream corpus defect is now pinned rather than papered over:** the `orderby` manifest lists 28 entry slots but one IRI is duplicated, so the true count is 27 and the total is 658. Anyone re-deriving the number from the manifest list rather than the loaded set will get 659 and think the corpus drifted.

PR #226, opened 2026-09-01
feat(text): deterministic full-text search over RDF 1.2 through the property-function seam

### LINKED ISSUE #193: `purrdf-text`: deterministic full-text search via the property-function seam
# `purrdf-text`: deterministic full-text search via the property-function seam

## Summary

No W3C standard exists for SPARQL full-text search: five incompatible vendor
dialects are in the field (Jena text, GraphDB FTS, Virtuoso `bif:contains`,
Stardog, AllegroGraph), and the W3C SPARQL working group dropped FTS from
scope. This item is a deterministic, in-memory inverted-index implementation
exposed through the property-function seam (item 3), with caller-bound index
IRIs — no PurRDF-minted vocabulary. Ships as an out-of-core sibling crate
(`purrdf-text`), keeping heavy indexing machinery out of `purrdf-core`. Depends
on item 3 (property-function registry).

## Acceptance sketch

- `purrdf-text` builds a deterministic in-memory inverted index over a caller-
  specified set of literal predicates.
- A property function (caller-bound IRI, e.g. an index-specific IRI the caller
  registers) performs a text search and returns matching subjects/scores
  through the property-function seam.
- Two independent runs over the same input produce byte-identical results
  (determinism test).
- `cargo test -p purrdf-text` passes; the crate builds `wasm32-unknown-unknown`
  clean (`make wasm` or equivalent).

Effort: M.

paudley:
Closes #193

## What this is

A new out-of-core sibling crate, `purrdf-text`, delivering deterministic in-memory ranked retrieval to SPARQL through the **already-shipped** property-function seam. `crates/sparql-eval/src/property_fn.rs`, `property_fn_eval.rs`, `property_fn_plan.rs` and all of `crates/rdf-core/` and `crates/sparql-algebra/` are **byte-unchanged** — asserted as a gate, not claimed in prose.

The roadmap text calling the property-function registry "absent" is stale; the registry shipped. This is a consumer of that seam.

## Two design hinges

**The shipped row ceiling *is* top-k.** Ranked retrieval's central optimization is already the seam's `ceiling: Option<u64>` licence, whose soundness condition ("the emitted rows are the first rows the unbounded scan would have produced") holds precisely because emission order *is* rank order. No parallel mechanism was invented. This is **measured, not asserted** — `tests/ceiling_pushdown.rs` records what the engine actually offers:

| Query shape | Recorded ceiling |
|---|---|
| `… } LIMIT 3` | `[Some(3)]` |
| `ORDER BY DESC(?score) LIMIT 3` | `[None]` |
| repeated variable + `LIMIT 1` | `[None]` |
| BGP-driven `Lateral`, `LIMIT 3` | `[Some(3), Some(2), Some(1)]` |

The last row shows the ceiling shrinking across a `Lateral`'s per-row invocation storm, with the fourth invocation never opening. The `None` cases are not defects: `governor/soundness.rs` states `ORDER BY` + `LIMIT` is "a top-k problem with no certified lower bound". The `ORDER BY` is redundant anyway, so **bare `LIMIT k` is the documented idiom** and `ORDER BY ?rank` is the reproducing one.

**Determinism is a type property, not a test observation.** The crate contains no floating-point arithmetic at all. BM25 is evaluated in exact base-10 fixed point (`i128`), including a natural logarithm in integer arithmetic at a *fixed* 20-iteration count. A libm `ln` differs by ULPs between implementations, and one ULP can flip two near-tied ranks — a native-vs-wasm answer divergence, which AGENTS.md §1 ("one behavior, carried verbatim into Rust, Python, WebAssembly, and C") forbids. Enforced by a crate-level `float_arithmetic` lint gate plus a grep gate, both in CI's clippy run.

## Surface

Two relations, not one type with a mode switch:

```sparql
?doc <caller-iri>  ( "needle" ?score ?rank ?lang ?matched )   # ranked retrieval
?doc <caller-iri2> ( "term" ?lang ?position )                 # one row per occurrence
```

`?score` is `xsd:decimal` (exact, no rounding of the underlying value); `?rank` is 1-based **within its `(graph, language)` partition**; `?matched` counts distinct needle terms present.

**Phrase and proximity are delivered, not declined** — through composition in SPARQL rather than an embedded query dialect:

```sparql
?doc <ex:occurs> ( "quick" ?l ?p1 ) .
?doc <ex:occurs> ( "brown" ?l ?p2 ) .
FILTER(?p2 = ?p1 + 1)
```

`FILTER(ABS(?p2 - ?p1) <= 3)` gives proximity; `FILTER(?matched = 2)` gives conjunctive retrieval. An embedded syntax would have minted a sixth incompatible vendor dialect — the exact thing the issue's Summary indicts.

## RDF 1.2 first class

`RdfDataset::quads()` returns only the asserted triple table; `annotation_quads()` is a **separate side table** whose trait default is `std::iter::empty()`. An index reading only `quads_for_pattern` indexes **zero** annotation literals. The index reads both layers, so text carried only by `:s :p :o {| :note "..." |}` is searchable with the reifier as the document subject. Guarded by a test that first asserts a second index over an asserted-only predicate answers empty, so it cannot pass for the wrong reason.

## Mints nothing

No vocabulary IRI, no default predicate set, no default registration. Three tests pin the boundary: an IRI declared but unregistered hard-fails naming the IRI; a declared *namespace* containing an unregistered IRI hard-fails; an IRI in neither stays an ordinary BGP triple pattern. There is deliberately **no `?graph` row position** — the graph is in the document key so text is never merged across graphs, but RDF has no term denoting the default graph and a fixed-width `PfRow` would force a minted sentinel.

## Things worth reviewing closely

- **Ceiling accounting.** `args_are_admission_transparent` returns `true` for `Arg::Constant(_)`, so a constant at *any* position still causes a ceiling to be offered. A constant at a post-rank position (`?doc`, `?score`, `?matched`) must therefore withhold the ceiling from the selector. `a_ceiling_with_a_bound_doc_still_emits_the_matching_row` was **verified to fail against the naive implementation** (it returned `[]` where a rank-2 row belonged).
- **Cardinality honesty.** `rows_per_invocation` is a real function of the mode, swept over all 64 search and 16 occurrence patterns, asserting both `declared >= emitted` *and* tightness (`== 1` attained) wherever the table claims 1. Verified to bite by clamping a bound and observing the sweep fail.
- **`explain()`** returns per-term tf/df/idf/contribution; a test asserts contributions sum **exactly** to the score — an equality, not a tolerance, which is the point of exact arithmetic.
- **`verify_binding()`** closes a silent-wrong-answer channel: `open()` receives no dataset, and while the engine validates registry identity at prepare time, nothing validates dataset identity, so an index paired with the wrong dataset emits documents that join back to zero rows with no diagnostic.

## Honest limits, stated rather than papered over

- **Unicode table version skew.** Case folding is at Unicode 16.0.0 while normalization, segmentation and `std` are at 17.0.0. Answers stay deterministic (the lockfile pins them); the risk is a dependency bump, which a golden token-vector test turns into a loud failure. All four table versions are folded into the index fingerprint.
- **Blank-node labels.** Document ids are a function of content *except* through blank-node labels, which are a parsing artifact. Two isomorphic datasets with different labels produce different fingerprints. Pinned by an adversarial test rather than glossed.
- **`ORDER BY DESC(?score)` may disagree with `ORDER BY ?rank`**, because the reported score is rounded while the internal order is exact. Documented and tested.

`docs/design/purrdf-text-scoring.md` records all of these.

## Gates

`cargo fmt --check`; `clippy -p purrdf-text -p purrdf --all-targets -D warnings`; `cargo test -p purrdf-text` (**152 tests**); `cargo test -p purrdf`; wasm32-unknown-unknown release build of both; rustdoc `-D warnings`; `check-no-features`; `check-versions` (now reporting 19 publishable crates); `check-doc-claims` / `check-brand-casing` / `check-issue-refs` / `check-licenses`; `check-generated.sh` (no drift); `cargo build --locked`. Deferral scan: 0 hits over 10,971 added lines. `.deficiencies`: untouched.

`make check` / `make test` / `make conformance` were **not** run locally (30+ min); CI covers them.

## Summary by CodeRabbit

* **New Features**
  * Added deterministic full-text search over RDF literals through the main PurRDF API and SPARQL property functions.
  * Supports Unicode normalization, case folding, word segmentation, and CJK tokenization.
  * Added reproducible fixed-point BM25 ranking with graph and language filtering.
  * Added term occurrence lookup, phrase and proximity matching, result limits, and subject-based document lookup.
  * Added deterministic behavior across native and WebAssembly environments.

* **Documentation**
  * Added usage, scoring, integration, benchmark, and release documentation for text search.

* **Tests**
  * Expanded coverage for indexing, ranking, SPARQL integration, determinism, and WebAssembly compatibility.


paudley:
## Plan as executed

**Transformations the work is built around**
- **A — the shipped row ceiling is top-k**, conditionally. Measured: bare `LIMIT 3` gives `Some(3)`; `ORDER BY … LIMIT 3` gives `None`; repeated variable gives `None`; `Lateral` gives `[Some(3), Some(2), Some(1)]`.
- **B — determinism as a type property.** No floating point in the crate; BM25 in exact base-10 fixed point with an integer `ln` at a fixed iteration count.

**Completeness contract** (every check executable on a production surface)

| # | Requirement | Executable check |
|---|---|---|
| R1 | deterministic in-memory index over caller-specified literal predicates | `--test index two_independently_built_datasets_share_a_fingerprint` |
| R2 | property function under a caller-bound IRI returns subjects and scores through the seam | `--test search_property_function ranked_rows_for_a_needle` (SPARQL text through `NativeSparqlEngine::query_with_options_view`, exact rows) |
| R3 | two independent runs byte-identical | `--test determinism byte_identical_json_across_two_independently_built_datasets`, `byte_identical_json_across_repeated_runs` |
| R4 | `cargo test -p purrdf-text` passes; wasm32 clean | both run; `purrdf-text` in the `make wasm` and `ci.yaml` allowlists |
| R5 | out-of-core sibling crate | `git diff --stat origin/main...HEAD -- crates/rdf-core/` empty |
| R6 | no minted vocabulary; unregistered IRI hard-fails | `an_iri_declared_but_unregistered_is_a_hard_error`, `a_namespace_declared_with_an_unregistered_iri_is_a_hard_error`, `an_iri_in_neither_is_an_ordinary_triple_pattern` |
| R7 | no Cargo features | `check-no-features.py` |
| R8 | reproducible ranking, explicit total order | `golden_scores_are_exact_decimals`, `equal_scores_are_broken_by_document_id_deterministically`, float lint + grep |
| R9 | Unicode / case folding / language tags / base direction | `full_case_fold_matches_sharp_s`, `compatibility_fold_matches_fullwidth_and_ligatures`, `cjk_tokenization_matches_the_documented_model`, `golden_token_vectors_pin_the_unicode_tables`, `language_splits_documents_but_base_direction_does_not` |
| R10 | RDF 1.2 first class, incl. the statement layer | `annotation_text_is_searchable_through_sparql`, `annotation_layer_literals_are_indexed`, `blank_node_subjects_are_indexed` |
| R11 | honest `rows_per_invocation` | sweep of all 64 search + 16 occurrence patterns; declared >= emitted, and tightness where 1 is claimed |
| R12 | scoring maps onto the existing ceiling contract | `tests/ceiling_pushdown.rs` (4 measured cases) + `a_ceiling_with_a_bound_doc_still_emits_the_matching_row` |
| R13 | seam byte-unchanged | empty diff over `sparql-eval`/`sparql-algebra`/`rdf-core` |
| R14/15 | no member pins; release-lane registration | `check-versions.py`, `check-doc-claims.py` |
| R16 | phrase and proximity expressible | `tests/phrase_composition.rs` |
| R17 | index/dataset binding verifiable | `verify_binding_accepts_the_source_dataset_and_rejects_another` |

**Enhancements considered and declined**
- *Embedded boolean/phrase/field syntax in the needle* — the capability is delivered instead (R16) via SPARQL composition; an embedded syntax would mint a sixth vendor dialect.
- *Stemming / stop-word lists* — on retrieval grounds, not scope: BM25 IDF already down-weights high-document-frequency terms, and a per-language stemmer corrupts multilingual data.
- *tantivy or any third-party engine* — mmap, threads, filesystem; not wasm32-clean.
- *`f64` BM25 with libm `ln`* — ULP differences flip near-tied ranks between native and wasm.
- *Persisted index artifact bound to a pack* — `EmbeddingTarget` binds to a pack plus a certified RDFC digest, but the seam operates on a `DatasetView` and a consumer with a plain `Arc<RdfDataset>` has no pack; the issue specifies in-memory.
- *Python / npm / C-ABI / CLI surfaces* — `bindings/python` depends on nine crates and omits `purrdf-datalog`, `purrdf-entail`, `purrdf-columnar`, `purrdf-iri`, so no repo-wide mandate exists, and the issue asks for none.
- *Incremental index mutation* — `TextSearchRelation` declares `Volatility::Stable`, which the fork-join gate reads to admit it across workers; a mutable index would make that declaration false.

---

## What am I least confident about, and why?

**1. The Unicode table version skew is the weakest part of the determinism story.** Case folding sits at Unicode 16.0.0 while normalization, segmentation and `std` are at 17.0.0. Within a locked build this is deterministic, and the fingerprint plus the golden token-vector test convert a bump into a loud failure — but it means the analyzer is not internally consistent with any single Unicode version today. I could not close this without either pinning an older normalization crate or writing case folding by hand. I chose to surface it rather than hide it, but I would not describe the analyzer as "Unicode 17 correct".

**2. `?rank` being per-partition is a semantic judgement, not a forced move.** It follows from per-partition statistics, which follow from not wanting to compute an English needle IDF against a mostly-non-English corpus. But a user with one language and one graph will never observe the distinction, and a user with several may be surprised that `LIMIT 10` walks partitions in key order rather than returning a globally best ten. The alternative — a global rank — is defensible if you accept cross-corpus score comparison. I do not, but this is the decision most likely to be re-litigated.

**3. Performance is unmeasured in any meaningful sense.** There is a criterion bench, and it is report-only per repo discipline. I have not profiled the bound-`?doc` invocation storm, where the relation computes a full partition ranking per invocation because rank is a property of the whole partition. On a large corpus inside a `Lateral` that is O(documents) per left row. I flagged this as a real cost and did not fix it, because fixing it well means a memo keyed on the analyzed needle and I did not want to add a cache whose invalidation story I had not thought through.

**4. I did not run `make check`, `make test`, or `make conformance`.** Every narrow gate passes, including the workspace `cargo build --locked` and the umbrella own tests, but the full suites are CI.

- **Whether the `caseless` dependency is acceptable long-term.** It is MIT-only (the workspace is MIT OR Apache-2.0), it is at Unicode 16, and its last release is old. Nothing in the repo gates dependency licences, so it passed — but a maintainer may have a policy I could not read off the tree.
- **Whether `crates/text` is the intended directory name.** The workspace convention is inconsistent (`rdf-core` for `purrdf-core`, `shapes` for `purrdf-shapes`), and I picked `text` for `purrdf-text` by analogy. Renaming later is cheap; renaming after publication is not.
- **How this should interact with the vector/embedding side.** The PURREMB machinery in `purrdf-core` already models targets, digests and token spans for dense retrieval. I deliberately did not build on it (the pack binding is unavailable on the seam production path), but hybrid lexical+dense fusion is the obvious next thing someone will want, and I do not know whether the intended shape is `purrdf-text` growing toward PURREMB or a third crate fusing both.
- **Whether the absent-predicate hard-fail matches maintainer intent.** A configured predicate missing from the dataset is an error here, on the reasoning that a mistyped IRI would otherwise silently remove part of the corpus. `DatasetView::term_id_by_value` documents the opposite posture for structural walks. Both precedents exist in the tree; I picked the stricter one.

PR #227, opened 2026-09-01
feat(shapes): SHACL 1.2 SPARQL Extensions / Node Expressions / Rules alignment

### LINKED ISSUE #191: SHACL 1.2 SPARQL Extensions / Node Expressions / Rules alignment
# SHACL 1.2 SPARQL Extensions / Node Expressions / Rules alignment

## Summary

Align the existing SHACL-AF implementation
(`crates/shapes/src/expression.rs`, `crates/shapes/src/rules.rs`,
`crates/shapes/src/user_fn.rs`) with the SHACL 1.2 SPARQL Extensions, Node
Expressions, and Rules Working Drafts. Concrete alignment points: layered rule
execution order, `sh:condition` semantics, SHACL Rule Language (SRL) awareness,
and tracking the prebinding "at risk" threads noted in the umbrella issue's
watch list — the SHACL 1.2 WDs are still churning on exactly this point, so this
item should track WD updates rather than target a single frozen snapshot.

## Source pointers

- `crates/shapes/src/expression.rs` — node-expression evaluator.
- `crates/shapes/src/rules.rs` — SHACL-AF rule execution.
- `crates/shapes/src/user_fn.rs` — user-defined function integration.

## Spec references

- W3C SHACL 1.2 SPARQL Extensions, Working Draft: https://www.w3.org/TR/shacl12-sparql/
- W3C SHACL 1.2 Node Expressions, Working Draft: https://www.w3.org/TR/shacl12-node-expr/
- W3C SHACL 1.2 Rules, Working Draft: https://www.w3.org/TR/shacl12-rules/

## Acceptance sketch

- Rule execution order matches the layered-execution model in the SHACL 1.2
  Rules WD, covered by a test with rules that would produce different results
  under an unordered vs. layered execution.
- `sh:condition` is implemented per the current WD semantics with a positive and
  negative test.
- Prebinding behavior is documented against the WD's current (possibly "at
  risk") wording, with a comment noting the tracked WD section so a future WD
  revision is easy to reconcile against.
- `cargo test -p purrdf-shapes` passes, including new rule-ordering,
  `sh:condition`, and prebinding cases.

Effort: M-L.

paudley:
Closes #191

## What this is

Aligns the SHACL-AF implementation with the SHACL 1.2 family of Working Drafts across all three nouns in the issue title: **Rules**, **Node Expressions**, and **SPARQL Extensions**.

## Three corrections to the issue's own framing, verified against the live drafts

1. **`https://www.w3.org/TR/shacl12-rules/` is titled "SPARQL 1.2 RL"** (short name SRL, namespace `http://www.w3.org/ns/sparql-rl#`, WD 26 Aug 2026). It does **not** define `sh:rule`, `sh:condition`, or `sh:order` — those remain SHACL Advanced Features terms. The issue's "SHACL Rule Language (SRL)" is a misnomer for the **SPARQL** Rule Language. SRL supplies the *layered execution model*; SHACL-AF supplies the vocabulary. Both are cited by title in the code.
2. **The issue's source pointer `crates/shapes/src/user_fn.rs` does not exist.** The real files are `crates/shapes/src/shapes/parser/functions.rs` and `crates/sparql-eval/src/user_fn.rs`.
3. **The upstream W3C SHACL suite still contains no SHACL-AF rule manifests.** There is no vendored AF conformance suite to lean on. Validation here is against first-party corpus cases plus the pySHACL/DASH-derived vendored AF material — not a W3C AF suite, because none exists.

## The headline defect

`sh:order` was **semantically inert**. `apply_rules` sorted every rule into one flat list and ran a single global fixpoint against a snapshot rebuilt only *between* rounds, so no ordering could change the closure. The repo's own test was named `sh_order_is_applied_and_result_is_order_independent`.

Replaced with the SRL §6.5 stratified scheduler: strata keyed by `sh:order`, `once`/`general` partition per SRL §4.4, each stratum materialized before the next runs. `swapping_orders_changes_the_inferred_set` is the discriminating proof.

## Commits

| | |
|---|---|
| `78cbcc67` | Layered stratified rule execution (SRL §6.5) |
| `59922d80` | Value-based strata, growing divergence universe, bounded expression depth |
| `7353fdf5` | The SHACL 1.2 node-expression language, scoped and dual-spelled |
| `b5d6f0d0` | SPARQL-based node expressions, value-level aggregates, load-time conditions |
| `85bfc32e` | Expression-bodied SPARQL functions and the dataset-aware seam |

## Bugs found and owned along the way

- **Named-node cycles aborted the process.** The node-expression parser's cycle guard existed but was gated on `matches!(node, Term::BlankNode(_))`, so `ex:E sh:union ( ex:E )` recursed unbounded. Rust stack exhaustion aborts and is not catchable.
- **Structural expression nesting was unbounded.** `RecursionGuard` had no depth mutator at all; the only ceiling lived in the constraint engine. Added `MAX_NODE_EXPR_DEPTH = 256`, deliberately distinct from the 64 re-entry ceiling (the parser emits three structural levels per authored paging node, so reusing 64 would reject valid shapes).
- **RDF 1.2 hole:** `sh:min`/`sh:max`/`sh:sum`/`sh:orderby` hard-errored on triple terms and blank nodes, because the seam serialized operands to N-Triples text and spliced them into a `VALUES` block. Two tests *pinned that refusal as expected*. Removed the text bridge in favour of the evaluator's own value-level comparator and aggregate fold; the tests now assert the capability.
- **Unbounded, input-driven memory growth:** because the spliced query text embedded operand data, every aggregate/order call inserted a never-evicted `PlanCache` entry. Removing the bridge removes the cause; a test asserts 100 distinct operand sets add zero entries.
- **`sh:condition` could be silently skipped.** It resolved by string lookup at *firing* time, so a rule with `sh:condition ex:NotAShape` on an untargeted shape loaded green and entailed silently. Now resolved to a parsed `Shape` at shapes-load; unresolvable is a load error. Inline/anonymous and property-shape conditions now work.
- Two further latent aborts fixed: `collect_expression_classes` and `node_expr_mints_blank` both descended into function bodies that may call themselves.
- Two bugs introduced by the layering itself, caught in review and fixed before merge: `f64::total_cmp` stratum grouping split `0.0` from `-0.0`; and the divergence term-universe was computed once, so a `once` rule minting blanks in stratum 0 caused a *false* divergence error in stratum 1.

## Termination and determinism

- **Termination:** the scheduler is iterative. Expression evaluation is depth-bounded on every arm; parsing is cycle-guarded on all node kinds. Every limit fails closed with an error naming it — never an abort. Deep-nesting tests build the tree programmatically, not as Turtle, so they exercise the shapes parser rather than the Turtle parser.
- **Determinism:** `layered_entailment_is_byte_identical_under_insertion_permutation` builds the same shapes graph twice with rule triples in different insertion orders, entails both, serializes to N-Triples and compares **bytes**.
- **Anti-regression:** `sh_order_is_applied_and_result_is_order_independent` still passes **unweakened** — independent monotonic strata still commute.

## Vocabulary boundary

Every IRI implemented is defined by a W3C specification (`sh:`, `shnex: http://www.w3.org/ns/shacl-node-expr#`, `sparql: http://www.w3.org/ns/sparql#`). Implementing a spec-defined IRI is not minting; inventing one would be. **No PurRDF-authored IRI is added.** Several of my own descriptions were wrong and were corrected against the specs during implementation — `sh:TripleTermExpression` is a *constant* rule and not a triple-term constructor; the 1.2 declaring class is `sh:ListParameterExpressionFunction`, not `sh:SPARQLFunction`; the operator is `sparql:add`, not `sparql:plus`; there is no `shnex:union`.

## Pre-binding: a recorded WD-vs-conformance conflict

The 1.2 WD says a query "MUST not contain a `VALUES` clause **that mentions any potentially pre-bound variable**". The vendored W3C test `pre-binding/unsupported-sparql-002.ttl` uses `VALUES ?any { true }` — `?any` is *not* pre-bound — and requires rejection. The WD wording and the frozen conformance contract directly conflict.

Per AGENTS.md ("conformance corpora are the contract") the enforced behaviour stays at the conformance-backed rule, and the divergence is recorded in code at the check, citing the WD, its Appendix A, the pinning test, and the WD's own at-risk marker quoted as a spec citation. No behaviour change, no conformance regression.

## Gates

**Run:** `cargo fmt --all`; `cargo clippy -p purrdf-shapes --all-targets -- -D warnings`; `cargo clippy -p purrdf-sparql-eval --all-targets -- -D warnings`; `cargo test -p purrdf-shapes`; `cargo test -p purrdf-sparql-eval`; `cargo test -p purrdf-sparql-algebra`; `cargo build` for both crates on `wasm32-unknown-unknown`; `cargo check --workspace --all-targets`; `check-corpus-frozen.py`; `check-issue-refs.py`; deferral scan over the diff.

**Not run (stage 1 scope):** `make check`, `make test`, `make conformance`, `make wasm`. These are 30+ minute gates and must run before merge. In particular `make conformance` is excluded from `make check`, and this branch adds corpus cases, so conformance needs its own run.

## Summary by CodeRabbit

* **New Features**
  * Added SHACL 1.2 Node Expressions, including variables, paths, lists, matching, shape conformance, and custom functions.
  * Added `sh:nodeByExpression` validation with configurable messages and severity.
  * Added SPARQL-backed expression functions, focus-graph support, RDF 1.2 terms, and SHA-3 functions.
  * Added base IRI support for resolving relative references.
  * Improved SHACL Rules with ordered layers, run-once rules, fixpoint processing, and deterministic blank-node handling.

* **Bug Fixes**
  * Improved recursion protection and diagnostics for invalid expressions, graph templates, and unresolved references.

* **Tests**
  * Expanded SHACL-AF and conformance coverage.


paudley:
The full stage 1 plan, plus honest answers to "what are you least confident about" and "what do I not know but should", are posted on #191 and apply to this PR verbatim: https://github.com/Blackcat-Informatics/purrdf/issues/191#issuecomment-latest

Two items from that list are merge-blocking and I want them visible here:

1. **`make conformance` has not been run.** It is excluded from `make check`, this branch adds corpus cases and rewrites `scripts/conformance-frozen/vectors-shacl.sha256`, so it can redden independently. `make check` and `make wasm` were also not run — only crate-scoped gates were.
2. **`sh:order` layering is an observable behaviour change** for any rule set with a non-monotonic `sh:condition`. That is the fix working as intended, but it lands with no deprecation path.

Issue #186, opened 2026-08-06
Generalized `ServiceResolver`


paudley:
# Generalized `ServiceResolver`

## Summary

The fourth extension seam: a single resolver trait covering the whole SERVICE
request, generalizing the existing `HttpTransport` inversion in
`crates/sparql-eval/src/remote_http.rs`. Per-service context — headers,
credentials, capabilities — belongs on the resolver, not encoded into the
service IRI itself. This seam is what makes in-process, capability-gated
federation possible (a resolver that never touches the network, for callers who
want SERVICE-shaped composition without SERVICE-shaped risk), and it is the
prerequisite for W3C Data Façades / Facade-X support (see the umbrella issue's
watch list). The SILENT-clause contract (behavior when a service is unreachable
or a capability is denied) needs to be documented as part of this work.

## Source pointers

- `crates/sparql-eval/src/remote_http.rs` `HttpTransport` — the existing
  inversion this generalizes.

## Acceptance sketch

- A `ServiceResolver` trait exists covering the full SERVICE request (endpoint,
  headers/credentials, capabilities), replacing or wrapping `HttpTransport`
  without changing its existing byte-level HTTP behavior.
- An in-process, capability-gated resolver implementation exists that never
  performs network I/O, and is exercised by a test that asserts no network
  access occurs.
- SILENT-clause behavior is documented and covered by a test for both the
  network resolver and the in-process resolver.
- `cargo test -p purrdf-sparql-eval` passes, including SERVICE tests against
  both resolver implementations.

Effort: M.

Part of the SPARQL breadth roadmap: #182

PR #233, opened 2026-09-02
fix(shapes,sparql-eval): restore the workspace to compiling under its own nightly pin

paudley:
## What broke

`main` does not build under its own toolchain pin. Two independent regressions, both from the same collision pattern.

**1. A compile error** — `cargo check`/`clippy` both die:

```
crates/shapes/src/rules.rs:1150: error[E0369]: binary operation `==` cannot be
applied to type `Vec<shapes::Shape>`: std::vec::Vec<shapes::Shape>, [_; 0]
```

**2. A denied lint** — `cargo clippy --workspace --all-targets -- -D warnings`:

```
crates/sparql-eval/src/service.rs:1073: error: used `assert!` to check that a value is empty
```

Between them, `workspace`, `test`, `capi`, `wasm` and `conformance` are red on `main`.

## Why they got through

#231 (the nightly pin) mechanically rewrote `assert!(x.is_empty())` into `assert_eq!(x, [] as [_; 0])` at all 78 then-existing sites, to satisfy `clippy::assert_is_empty`. Two PRs then landed within minutes of it, each green on its own base, neither able to see the sweep:

- **#227** changed `Rule::conditions` from `Vec<String>` to `Vec<Shape>` — a `sh:condition` now resolves to a parsed shape at shapes-load rather than by IRI-string lookup at rule-firing time. `Vec<String> == []` type-checks; `Vec<Shape> == []` does not, because `Shape` is not `PartialEq`. The sweep's rewrite of that one site stopped compiling.
- **#186** added a *new* `assert!(…is_empty())`, in the spelling the sweep had just finished removing everywhere else.

Both merges were textually clean — different lines, no conflict — so the breakage exists only in the combined tree, which no PR's CI ever built.

## The fix

Two lines, each taking the spelling that fits its own type:

- `crates/shapes/src/rules.rs` — `assert_eq!(…conditions.len(), 0)`. The same claim, equally free of `clippy::assert_is_empty`, and already what this workspace uses wherever the element type is not `PartialEq` (e.g. `crates/rdf-core/src/ir/pack/bits.rs:1142`). A comment at the site records why the array form cannot be used there, so the next sweep does not re-break it.
- `crates/sparql-eval/src/service.rs` — `assert_eq!(…request_headers(), [] as [_; 0])`. `Vec<(String, String)>` *is* `PartialEq`, so this one takes the dominant spelling, matching the sibling assertion twenty lines above.

No test is weakened: both assertions make exactly the claim they made before.

## Verification

On `nightly-2026-09-01` (one day behind the pin, same lint surface):

- `cargo clippy --workspace --all-targets --locked -- -D warnings` — **clean, exit 0**. Run across the whole workspace, so these two are confirmed to be the only collisions the sweep left behind.
- `cargo test -p purrdf-shapes --locked` — 599 lib tests pass, 0 failures, including `rules::tests::parses_order_deactivated_and_conditions`, which carries the first assertion.

### DIFF
--- a/crates/shapes/src/rules.rs
+++ b/crates/shapes/src/rules.rs
@@ -1147,7 +1147,11 @@ mod tests {
         assert!(matches!(shape.rules[0].body, RuleBody::Triple { .. }));
         assert!(!shape.rules[0].deactivated);
         assert!(shape.rules[0].order.is_none());
-        assert_eq!(shape.rules[0].conditions, [] as [_; 0]);
+        // `conditions` is `Vec<Shape>` — conditions resolve to parsed shapes at
+        // shapes-load, not to IRI strings — and `Shape` is not `PartialEq`, so the
+        // `[] as [_; 0]` spelling the rest of the workspace uses does not compile
+        // here. Asserting the length is the same claim and is equally lint-clean.
+        assert_eq!(shape.rules[0].conditions.len(), 0);
     }
 
     #[test]
--- a/crates/sparql-eval/src/service.rs
+++ b/crates/sparql-eval/src/service.rs
@@ -1070,7 +1070,7 @@ mod tests {
         let ungranted =
             ServiceProfile::new(ServiceCapabilities::granting([ServiceCapability::Query]))
                 .with_credential(ServiceCredential::Bearer("t".to_owned()));
-        assert!(ungranted.request_headers().is_empty());
+        assert_eq!(ungranted.request_headers(), [] as [_; 0]);
     }
 
     #[test]


paudley:
Superseded by #232, which landed the same two fixes with the same diagnosis while this was in CI.

#232's  fix is the better of the two: it goes back to `assert!(conditions.is_empty())` rather than asserting the length, because `clippy::assert_is_empty` is applicability-gated — it only fires where the array-comparison rewrite actually type-checks, so a non-`PartialEq` element type is not a finding at all. That is a smaller diff and needs no explanatory comment. The `service.rs` fix is identical in both.

Verified independently on the merged `main` (5d881d81): `cargo clippy --workspace --all-targets --locked -- -D warnings` is clean, exit 0. Nothing from this branch is still needed.

PR #276, opened 2026-09-08
perf: bound prepared-query reuse and preserve typed RDF identity

### LINKED ISSUE #273: perf: bound and reuse prepared query execution for native dataflow consumers
## Problem and approved implementation

GMEOW correspondence and deontic DAG consumers repeatedly construct query engines, parse static query text, and cannot directly enter an already-admitted prepared operation under a shared governor. PurRDF 1.1.0 already has PreparedQuery, registry-aware preparation and DatasetView execution, but its parse cache and BGP-order cache are unbounded. Reuse those semantics and complete the missing production seams.

- Add explicit entry/byte cache limits with deterministic eviction and hit/miss/eviction/retained-byte counters. Existing constructors and results remain compatible; the GMEOW producer explicitly selects the bounded surface. Retained cache bytes and still-pinned plans have distinct accounting.
- Admit compiler-built Query algebra through the same feasibility/registry checks as parsed queries, without rendering/reparsing SPARQL and without treating PreparedQuery::rewritten as a bypass of admission.
- Expose prepared execution under an existing operation governor, retaining registry checks, prebinding behavior, completeness, operational-failure precedence, cancellation and atomic publication.
- Reuse immutable Arc<PreparedQuery> across workers; mutable evaluation state remains worker-local. No global lock may serialize query execution. Operand data arrives as typed bindings, not interpolated query strings.
- Preserve parse/registry/base-IRI key distinctions. Statistics fingerprints are only join-order hints, never answer identities. Result caching in GMEOW additionally binds exact dataset, context, bindings, registries and contract; incomplete or ambient/effectful results are not promoted to reusable complete answers.
- Add controlled preparation/repeated-evaluation benchmarks and structural counters, measuring parse removal independently of execution/fusion.

## Owned verification

Test bounded retention and oversize entries, live Arc handles after eviction, registry/base-IRI distinctions, malformed algebra refusal, shared-governor charging/cancellation, prepared/text outcome parity, and public compatibility. Reuse existing upstream conformance instead of reproducing it. Existing downstream consumer compilation must continue to work.

## Compatibility and ownership

PurRDF is an editable first-party dependency, already released as 1.x. Keep changes non-breaking where practical: add constructors/methods/types rather than changing existing signatures, public struct literals, required trait methods or exhaustive result variants. Improve internals freely while preserving observable semantics and deterministic output. A public-contract obstacle needs an explicit versioned migration. Keep MSRV, wasm support, no semantic Cargo features, and caller-supplied vocabulary contracts.

New generic regressions and compatibility checks live only in PurRDF. GMEOW consumes the tested release and tests its own boundary behavior; it does not rerun or duplicate PurRDF's broad conformance suite.

## Shared acceptance contract

This is current, user-approved implementation scope under Blackcat-Informatics/gmeow-ontology#1704, not a follow-up. Work in .worktrees/. SOTA and greenfield engineering govern architecture; existing code is evidence rather than a constraint. Preserve existing work. Retain standpoint, provenance, complements, loss evidence, logical stage contracts and required checks. Bounded-corpus evidence does not authorize unrestricted optimizer rewrites.

Measure on the authenticated O3/full-LTO producer; tests/debugging are separately built. Preserve both independent cold CI generations. Tests never produce the GMEOW corpus. Generic RDF/SPARQL/codec/engine conformance is owned and run only by PurRDF; GMEOW asserts its own semantic and product contracts without duplicating upstream fixtures or suites.

Completion requires implementation, focused owned verification, the owning repository's required gate evidence, and a reviewable PR. Principles 1, 2, 3, 4, 5, 6, 7, 9, 12, 15, 17, 18, 19.

## Dependencies

Blocked by: none.

paudley:
Repeated preparation retained unbounded parsed-query and join-order caches, and compiler-built or caller-mutated algebra could bypass admission. Both caches now have deterministic entry/byte limits, separately account for caller-held plans, and support prepared execution under an existing operation governor.

- Keep exact query/base/registry keys, bounded retention, oversize refusal and eviction counters. Immutable plans can be shared across workers with local evaluation state.
- Admit parsed, compiler-built, rewritten and mutated algebra before governed work. Malformed VALUES, unsafe nesting and infeasible property-function plans return diagnostics. Repeated SELECT/GROUP BY variables keep normalized schemas, and aggregate name collisions retain their error even under exhausted budgets.
- Inspect aggregate sort expressions throughout registry/property planning, governor soundness, Basic-profile checks and witness-result restriction. FOLD ordering cannot hide unregistered calls, bypass bound-only relation requirements or leak existential witness values.
- Preserve graph-qualified reifiers and annotations through GTS composition, deduplication and relocation. Typed import and canonical relabeling avoid unused owned representations.
- Correct directional literal datatype identity across carriers and bindings. C/Python and binary readers reject inconsistent expanded components. Python equality, hashing, store operations and query bindings agree on direction, language case and opaque blank labels. Pack dictionaries require canonical, unique, ordered entries without expanding shared triple trees.

The literal correction changes newly generated bytes and content identities for affected datasets. Regenerate derived products from canonical inputs; authenticated historical bytes keep their original identities. Existing constructors and execution signatures remain available.

Validation at `aa28a3a7`: the isolated pinned `make check` passed 7,804 workspace Rust tests plus the generated-artifact test, 10 existing ignores, and all 21 release-crate WASM builds. `make pytest` rebuilt the extension and passed 788 tests (one existing skip, four expected failures). `make conformance` passed 4,488 cases with the same 19 ledgered expectations and no failures. Compiler/Cargo hashes and source identity match before and after the complete run. CI is green and all seven actionable review threads are resolved.

Three alternating benchmark pairs used identical harnesses and authenticated O3/fat-LTO/one-codegen-unit builds. Host one-minute load ranged from 35.32 to 57.65, producing substantial variance: cold-preparation candidate/base ratios were 3.024, 0.952 and 1.243; prepared-execution ratios were 0.363, 1.146 and 0.943. Warm-preparation ratios were 0.573, 0.266 and 0.233. These are shared-host component observations; the variance prevents a reliable execution speedup conclusion.

The separate 100-operation counters verify 100 allocations / 12,800 requested bytes for cached preparation, versus 3,500 / 457,300 without retention. Churn retains four entries / 2,960 charged bytes within the four-entry / 4,096-byte policy, accounts for all 96 evicted caller-held plans separately, and releases every plan charge after final drop. Prepared execution performs no query-cache lookup; shared-governor execution completes with 25,900 fuel, 6,400 rows and 128 peak cells. Allocation requests, admitted payload, governor work and process memory remain distinct measurements.

Combined GMEOW producer measurements are owned by its existing Blackcat-Informatics/gmeow-ontology#1745 and Blackcat-Informatics/gmeow-ontology#1738 workstream under Blackcat-Informatics/gmeow-ontology#1704, which consumes these upstream APIs. Component measurements do not establish downstream product acceptance.

Fixes #273.


paudley:
Implemented and pushed the statement-layer graph repair in `23606b05eccc7991c2a10a972c4bed3bbde70c1a`.

`SnapshotBuilder` now retains graph identity for reifiers and annotations through native ingestion, optional default-graph relocation, canonical ordering, complete-row deduplication and snapshot emission. Default-graph rows retain their original encoding; graph-qualified rows use the existing optional GTS graph field. The obsolete prose claiming reification was never graph-scoped is removed, along with its now-unused brand exception.

Validation: all 18 composer tests passed; the complete `make check` passed, including workspace Clippy, tests, frozen-corpus checks, hygiene and the optimized release WASM build. No frozen vector was changed. Earlier attempts caught an unnecessary raw-string delimiter and the obsolete brand exception; both were repaired before the successful full gate.

The downstream GMEOW modal branch now pins this exact commit. Its product regression checks attributed opposition retains the same verifier-input identity across GMEOW's production bundle boundary. Full downstream gate validation and end-to-end performance measurements remain outstanding; the earlier component measurements are not measurements of this new head.

Issue #281, opened 2026-09-09
feat: share selected native GTS payloads and parsed shape datasets


paudley:
GMEOW's current pipeline implementation needs the native GTS dataset and a small explicit set of archive payloads from the same read. The existing native importer preserves segment-scoped RDF identity but exposes blob references without bodies, forcing callers through a second reader. SHACL consumers also reparse shape text solely to recover the dataset that `Shapes` already retains.

This is an implementation dependency of Blackcat-Informatics/gmeow-ontology#1704, #1732 and #1745.

Acceptance:
- Add `import_gts_events_with_blobs` over the same `SegmentResolver` and native builder. Preserve all existing importer signatures and behavior; no wire, vocabulary, feature, version, dependency or release change.
- Require explicit exact digest or final public representation selectors. Missing, ambiguous, corrupt or unsupported selected inputs fail closed.
- Return exact decoded shared bytes, metadata, and independently accurate payload and metadata frame/segment/head provenance. Preserve native RDF 1.2 and blank scopes from the original importer.
- Bound encoded, decoded/intermediate, retained-total and metadata bytes before retention. Do not retain or decode unrelated known-digest payloads. Document bounds and the explicit refusal when metadata selects an earlier unretained payload without resupplying it.
- Expose the immutable shared dataset retained by `Shapes` without reparsing, copying or changing document-prefix resolution.
- Keep all generic format/codec/SHACL/import conformance upstream. GMEOW tests its product semantics and adoption only. Add focused upstream selection, scope, corruption, ordering, provenance and bound controls; preserve hosted broad conformance and wasm/MSRV gates.

Implementation and verification plan:
1. Preserve the existing `.worktrees/native-import-blobs` branch and its completed source batch; inspect shared reader/event compatibility and the immutable shape accessor.
2. Run focused native-import tests and warning-denied checks over the affected Rust surfaces, formatting and rustdoc. These local checks are scoped to this change per the user's instruction to avoid repeated broad conformance runs; they do not claim the full upstream gate has passed.
3. Commit through normal verification/signing, push, and open a focused PR. Reconcile every required current-head CI and review result before merging through ghprsq. GMEOW then pins the merged revision and validates actual producer adoption, preserving both independent cold CI generations.

[...]

PR #277, opened 2026-09-08
perf: share RDF views and publish typed graph results

paudley:
Fixes #274.

Repeated small dataset contributions, SHACL validation and CONSTRUCT results can now retain native RDF dictionaries and indexes through shared views, with explicit ownership limits and typed publication. The implementation preserves scoped RDF 1.2 identity, statement metadata and complete-result publication using the existing model and evaluator.

- Immutable composite views support independent or shared blank scopes and explicit graph placement. Delta snapshots preserve suppression, reinsertion, all three RDF record tables, nested terms and declaration-only graphs. Finite retention limits and copy/freeze/materialization counters expose ownership costs without affecting content identity.
- Prepared physical probes are reused across native, composite, delta and SHACL adapters. Graph placement and union projection select plans for the actual source constraints. Matrix regressions cover bindings, graph placement, deletion/addition, ordering and duplicate-free native parity; debugger evidence confirms that row probes no longer rebuild plans.
- PreparedShapes shares exact immutable analysis and rebinds dataset-specific IDs and targets. Core and SPARQL share one class index when they use the same view. Borrowed projections and property-pair constraints avoid materialization; concrete compatibility accessors materialize lazily.
- Typed prepared CONSTRUCT stages and validates results before appending to a destination. It avoids an intermediate freeze, preserves destination-aware blank freshness, and publishes nothing after errors, cancellation or governor exhaustion. Fallible operations retain operational-error precedence. UPDATE WHERE reads mutation snapshots.
- Located duplicate rows now retain matching provenance during owned insertion and GTS import. The additive `push_quad_with_handle` returns the actual deduplicated row handle. Unrealized handles cannot acquire a later unrelated row, and validated append safely handles unmatched locations. Regressions reproduced the bounds panic, lost provenance and wrong-row association before these fixes.
- Visualization preserves cross-graph annotation relations for known reifiers; affected newly generated visual JSON/SVG/hashes change, and the final-head book build passes. The hosted workspace test job omits debug symbols after an observed disk failure; its optimization, assertions and overflow behavior remain unchanged.

[...]

Issue #14, opened 2026-07-02
ShEx phase 2: imports, semantic actions (Test extension), query shape maps


paudley:
## Goal

Close the two skipped trait categories in the ShEx validation harness and round out the shape-map surface. Current state: **1,051/1,051 attempted validation tests pass, 0 xfail**, with 54 skipped by trait (`crates/shex/tests/validation_conformance.rs`).

## Deliverables

- [ ] **IMPORT (32 skipped tests)**: transitive, cycle-tolerant schema imports with per-document base-IRI resolution. Resolution must be injectable (no ambient I/O — wasm32-clean); the conformance harness supplies a file resolver.
- [ ] **Semantic actions (22 skipped tests)**: parse+dispatch is already done; add an extension registry and ship the `http://shex.io/extensions/Test/` extension (print/fail semantics) to run the tagged tests. Arbitrary code evaluation stays out of scope.
- [ ] **Query shape maps** (`{FOCUS p _}` / `{_ p FOCUS}` node selectors resolved against the data graph) — the ShapeMap spec's query form, needed for "validate everything of type X" workflows.
- [ ] Optional stretch: result-shape-map JSON output at the suite's result-conformance level.

## Gates

Skip count drops to 0 (trait-skip list emptied); exact-count assertions updated; `cargo test -p purrdf-shex` green; clippy -D warnings; wasm32-clean; no Cargo features.

paudley:
# Implementation Plan for Issue #14 — ShEx phase 2

## Context

`purrdf-shex` ships a conformant ShEx validator, but four trait categories are
still *counted skips* in the conformance harness
(`crates/shex/tests/validation_conformance.rs:37`):

```rust
const SKIP_TRAITS: &[&str] = &["Import", "SemanticAction", "Extends", "ExtendsDiamond"];
```

Issue #14 closes the two that have corpus coverage (**Import** ~32 entries,
**SemanticAction** ~22 entries) and rounds out the shape-map surface with the
ShapeMap-spec **query form** (`{FOCUS p _}` / `{_ p FOCUS}`) plus
**result-shape-map JSON** output. `Extends`/`ExtendsDiamond` have **0** corpus
entries, so emptying the whole list is safe once the first two are handled.

**Why now / intended outcome:** move the trait-skip count to **0**, add a hard
assertion that keeps it there, and give ShEx a usable "validate everything of
type X" workflow. Per the user's scope decision this PR builds the *complete*
[...]

Issue #13, opened 2026-07-02
SHACL-SPARQL: pre-binding substitution semantics + $shapesGraph/$currentShape


paudley:
## Goal

Implement SHACL's pre-binding semantics precisely: pre-bound variables (`$this`, …) behave as **substituted values**, visible even in groups with no triple pattern — plus expose the shapes graph for `$shapesGraph`/`$currentShape`.

Covers the 3 pre-binding tests in the W3C ledger (`crates/shapes/tests/w3c_conformance.rs` `XFAIL`):
- `sparql/pre-binding/pre-binding-002` — `$this` inside FILTER-only UNION branches
- `sparql/pre-binding/pre-binding-005` — `$this` inside a FILTER-only group (`{ FILTER(bound($this)) }`)
- `sparql/pre-binding/shapesGraph-001` — `$shapesGraph`/`$currentShape` pre-bound variables

## Design notes

- The engine's standard SPARQL scoping is correct for SPARQL; SHACL-SPARQL needs a *substitution* evaluation mode (SHACL-SPARQL §5.2). Must not regress normal query evaluation — likely a flag on the eval request used only by the shapes engine.
- `$shapesGraph` requires the validator to expose the shapes graph as a queryable named graph during constraint evaluation (today shapes and data datasets are deliberately separate) — needs a union-dataset decision.
- The restriction *rejection* rules (MINUS/SERVICE/VALUES with pre-binding) already work (`crates/shapes/src/prebinding.rs`).

## Gates

Ledger shrinks by 3 (XPASS-enforced); `cargo test -p purrdf-shapes -p purrdf-sparql-eval` green; clippy -D warnings; wasm32-clean.

paudley:
# Implementation Plan for Issue #13

## Issue Summary

Implement SHACL-SPARQL **pre-binding substitution semantics** so that pre-bound variables (`$this`, `$shapesGraph`, `$currentShape`, `$value`) behave as if their values were syntactically substituted, not merely joined via SPARQL `VALUES`. This makes them visible inside FILTER-only groups and FILTER-only UNION branches. Also expose the shapes graph to SHACL-SPARQL queries via the pre-bound variables `$shapesGraph` and `$currentShape`.

The targeted W3C SHACL-SPARQL conformance tests are:

- `sparql/pre-binding/pre-binding-002` — `$this` inside FILTER-only UNION branches
- `sparql/pre-binding/pre-binding-005` — `$this` inside a FILTER-only group
[...]

PR #283, opened 2026-09-10
fix(rdf): admit archives holding unselected oversized payloads

paudley:
Follow-on defect fixes for the selected-blob importer merged in #282, plus the defects this branch introduced while fixing them.

## What started this

A selected import's byte budget names what *that import will retain*, not what the container is allowed to contain. But the reader enforced it for every blob frame with no knowledge of selection, and the IR path treats every reader diagnostic as a hard fold failure:

```
archive = [ 1 KiB blob "wanted", 100 KB unrelated compressed blob ]

import_gts_events(&bytes)                                    -> Ok
import_gts_events_with_blobs(&bytes, &[], limits(100, 100))  -> Err(rdf-ir-gts-fold-diagnostic)
```

The previous test suite asserted that as correct, with an empty selector list.

## What two audits then found

The first fix was itself defective, and re-auditing the repairs found three more. Every item below was demonstrated on the public API before being fixed, and each regression test was confirmed to **fail against the commit before it**.

**A container could talk its way past the budget seam.** The fix decided whether a diagnostic was a budget refusal by searching its human-readable detail for `"blob exceeds"`. Details interpolate container-supplied bytes — the reader echoes a header's `gts` field verbatim. A file whose header magic *was* `"blob exceeds"` suppressed the magic/version guard, so the bounded importer accepted, dataset and all, a container `import_gts_events` rejects. Byte ceilings now carry a typed `BLOB_BUDGET_DIAGNOSTIC`.

**A byte budget was silently deciding selection.** Two blobs sharing `rep="wanted"` errored correctly under a generous ceiling and returned **one blob — the wrong one — under a tight one**, because a refused payload vanished before `finish()` could count it. Refusals are now first-class data (`blob_refused`), stay selection candidates, and surface in `GtsImportWithBlobs::refused`. Naming one reports the real bound instead of the false "resolves to 0 blobs".

**A compressed snapshot defeated every blob ceiling.** The decode limit was gated on the frame being a blob frame, so an enclosing RDF frame decoded unbounded and its embedded blob map was retained entry by entry with no check. Both are bounded now.

**A refused frame was reported as success.** The frame ceiling shared the blob ceiling's code, which the importer treats as recoverable — so one compressed frame carrying 20,000 quads imported as **zero quads, `Ok`**. Declining a payload leaves the dataset intact; dropping a frame does not. It has its own code and is terminal.

**One blob stored twice became ambiguous.** Refused occurrences were counted without regard to identity. A refusal with no transform chain can be identified for free (its wire bytes are its decoded bytes), so the reader hashes those and a refusal already retained is recognised as the same payload.

**A forward reference was called external.** Whether a blob is external was decided at the occurrence, so a metadata-only frame preceding its own inline payload — accepted by `import_gts_events` — was reported as external. Decided at `finish()` now.

## Spec fidelity

- **`len` refusal deleted.** GTS-SPEC defines it only as an extension key and §22.2 states a reader MUST NOT reject a payload map solely for carrying an unknown key. Nothing in the workspace emits it. A payload with a disagreeing `len` now imports — the positive case the refusal never had.
- **External blobs** get their own code and message instead of a false claim about retention ordering.
- **Digest selectors** are documented as canonical `blake3:<hex>`, with the bare-hex negative pinned.

## Also

`Shapes::dataset()` — the whole of issue #281's criterion A5 — had zero tests and no callers; `Arc::ptr_eq` now pins the no-copy claim. A panic inside a `Result` API is a diagnostic. The retained-byte total uses checked arithmetic. Dead unbounded `cose::decrypt0` removed. README sections for both surfaces, a report-only criterion bench, and the three focused `make` targets renamed off a merged-and-deleted branch name.

## Verification

Per commit: `cargo fmt --all`, `cargo clippy -- -D warnings`, and the narrowest tests that prove the claim. `gts_selected_blobs` 31/31, `bounded_keyed_blobs` 6/6, `shared_shapes_dataset` 3/3, `purrdf-gts`/`purrdf-rdf`/`purrdf-shapes` suites green, rustdoc `-D warnings` clean. Full gate is CI's.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ endif
 CAPI_HEADER := crates/rdf-capi/include/purrdf.h
 
 .PHONY: help doctor metadata fmt check geo-determinism book book-samples book-pot book-po-update book-zh check-i18n check-issue-refs check-brand-casing check-spec-attribution changelog bump release-tags test doc bench bench-prepared-reuse bench-python columnar-oracle csvw-conformance csvw-oracle obographs-oracle projection-oracles pydantic-oracle linkml-oracle typescript-oracle graphql-oracle pytest conformance iri-resolver-hygiene rdf-core-hygiene wasm wasm-test wasm-pkg wasm-pkg-test wasm-pkg-bench playground playground-smoke \
-	capi-build capi-header capi-check capi-install test-native-import-blobs lint-native-import-blobs doc-native-import-blobs
+	capi-build capi-header capi-check capi-install test-gts-selected-blobs lint-gts-selected-blobs doc-gts-selected-blobs
 
 # The changelog generator is pinned so the committed CHANGELOG.md and the notes
 # the release workflow slices out of it stay byte-reproducible across machines.
@@ -156,19 +156,20 @@ release-tags: ## Cut + push rust-v/py-v/npm-v tags for VERSION after coherence c
 test: ## Run the workspace test suite.
 	cargo test --workspace --locked
 
-lint-native-import-blobs: ## Lint the selected native-import production and test surfaces only.
+lint-gts-selected-blobs: ## Lint the selected native-import production and test surfaces only.
 	cargo clippy -p purrdf-gts --lib --test bounded_keyed_blobs --locked -- -D warnings
 	cargo clippy -p purrdf-rdf --lib --test gts_selected_blobs --locked -- -D warnings
 	cargo clippy -p purrdf-shapes --lib --locked -- -D warnings
 
-doc-native-import-blobs: ## Check the native-import and shared shape-dataset public API documentation.
+doc-gts-selected-blobs: ## Check the native-import and shared shape-dataset public API documentation.
 	RUSTDOCFLAGS="-D warnings" cargo doc -p purrdf-gts -p purrdf-rdf -p purrdf-shapes --no-deps --locked
 
-test-native-import-blobs: ## Check bounded selected-blob import and native scope contracts only.
+test-gts-selected-blobs: ## Check bounded selected-blob import and native scope contracts only.
 	cargo test -p purrdf-gts --lib codec::tests:: --locked
 	cargo test -p purrdf-gts --test bounded_keyed_blobs --locked
 	cargo test -p purrdf-rdf --lib gts_import_sink::tests:: --locked
 	cargo test -p purrdf-rdf --test gts_selected_blobs --locked
+	cargo test -p purrdf-shapes --test shared_shapes_dataset --locked
 
 doc: ## Build docs for the 21 publishable crates with rustdoc warnings denied.
 	RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps --exclude purrdf-capi --exclude purrdf-python --exclude purrdf-sparql-conformance --exclude purrdf-cli
--- a/crates/gts/src/codec.rs
+++ b/crates/gts/src/codec.rs
@@ -66,13 +66,21 @@ pub enum CodecError {
     },
     /// The codec is known but the data is corrupt — the frame is damaged.
     Failed(String),
+    /// The data is well-formed but a caller-supplied byte ceiling refused it.
+    ///
+    /// Distinct from [`Self::Failed`] because refusing to *spend* bytes is not
+    /// evidence that the bytes are corrupt: a caller that set the ceiling may
+    /// legitimately skip this payload and keep reading, whereas corruption is a
+    /// statement about the container. Only a decode given an explicit limit can
+    /// produce this.
+    Limit(String),
 }
 
 impl fmt::Display for CodecError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Self::Unavailable { reason, detail } => write!(f, "{reason}: {detail}"),
-            Self::Failed(detail) => f.write_str(detail),
+            Self::Failed(detail) | Self::Limit(detail) => f.write_str(detail),
         }
     }
 }
@@ -379,7 +387,7 @@ pub fn decode_chain(chain: &[Codec], data: &[u8]) -> Result<Vec<u8>, CodecError>
 }
 
 fn decoded_limit(limit: usize) -> CodecError {
-    CodecError::Failed(format!("decoded transform output exceeds {limit} bytes"))
+    CodecError::Limit(format!("decoded transform output exceeds {limit} bytes"))
 }
 
 /// Reverse a codec chain, bounding every decoded intermediate and final buffer.
--- a/crates/gts/src/cose.rs
+++ b/crates/gts/src/cose.rs
@@ -274,29 +274,6 @@ pub fn recipient_kid(blob: &[u8]) -> Option<String> {
     parse_encrypt0(blob).map(|p| p.kid)
 }
 
-/// Open a COSE_Encrypt0 using a content key resolved by `kid` (§9.3).
-pub fn decrypt0(
-    blob: &[u8],
-    resolve: impl Fn(&str) -> Option<[u8; 32]>,
-) -> Result<Vec<u8>, Encrypt0Error> {
-    let parts = parse_encrypt0(blob).ok_or(Encrypt0Error::Malformed)?;
-    let key = resolve(&parts.kid).ok_or(Encrypt0Error::MissingKey)?;
-    if parts.iv.len() != 12 {
-        return Err(Encrypt0Error::Malformed);
-    }
-    let aad = enc_structure(&parts.protected);
-    let cipher = Aes256Gcm::new((&key).into());
-    cipher
-        .decrypt(
-            Nonce::from_slice(&parts.iv),
-            Payload {
-                msg: &parts.ciphertext,
-                aad: &aad,
-            },
-        )
-        .map_err(|_| Encrypt0Error::AuthFailed)
-}
-
 /// Internal bounded decryption failures, without extending the public error enum.
 #[derive(Debug)]
 pub(crate) enum BoundedDecrypt0Error {
--- a/crates/gts/src/reader.rs
+++ b/crates/gts/src/reader.rs
@@ -385,6 +385,15 @@ pub trait StreamingSink {
     fn blob_decode_limit(&self) -> Option<usize> {
         None
     }
+    /// Optional bound for decoding an ordinary (non-blob) frame's payload.
+    ///
+    /// A `snapshot` frame carries embedded blob bytes *inside* an RDF frame, so
+    /// a blob ceiling alone leaves the real bomb unbounded: the memory is spent
+    /// decoding the enclosing frame, long before any embedded entry is seen.
+    /// Sinks that set no bound keep the previous unbounded behavior.
+    fn frame_decode_limit(&self) -> Option<usize> {
+        None
+    }
     /// Opaque frame produced by unknown, encrypted, or damaged payloads.
     fn opaque(&mut self, _segment_index: usize, _opaque: &OpaqueNode) {}
     /// Signature status observed on a frame.
@@ -659,13 +668,16 @@ impl Folder<'_, '_, '_> {
                 ));
             };
             let chain = self.resolve_codecs(ids)?;
-            let limit = blob
-                .then(|| {
-                    self.sink
-                        .as_deref()
-                        .and_then(StreamingSink::blob_decode_limit)
-                })
-                .flatten();
+            // A blob frame is bounded by the blob ceiling; every other frame by
+            // the frame ceiling, which is what keeps an embedded snapshot blob
+            // map from decoding without limit.
+            let limit = self.sink.as_deref().and_then(|sink| {
+                if blob {
+                    sink.blob_decode_limit()
+                } else {
+                    sink.frame_decode_limit()
+                }
+            });
             let decoded = if let Some(content_key) = self.content_key {
                 let limit = limit.unwrap_or(usize::MAX);
                 let decrypt =
@@ -1191,9 +1203,24 @@ impl Folder<'_, '_, '_> {
             self.h_annot(&Value::Array(annot.iter().map(sh_row).collect()), index);
         }
         if let Some(Value::Map(blobs)) = map_get(entries, "blobs") {
+            let ceiling = self
+                .sink
+                .as_deref()
+                .and_then(StreamingSink::blob_decode_limit);
             for (_, b) in blobs {
                 if let Value::Bytes(bytes) = b {
                     let digest = digest_str(bytes);
+                    // An embedded entry obeys the same ceiling as a standalone
+                    // payload; without this a snapshot is a way to hand a
+                    // bounded consumer arbitrarily many unbounded blobs.
+                    if let Some(limit) = ceiling
+                        && bytes.len() > limit
+                    {
+                        let detail = format!("snapshot blob exceeds {limit} bytes");
+                        self.emit_blob_refusal(Some(&digest), None, bytes.len(), &detail);
+                        self.diag(BLOB_BUDGET_DIAGNOSTIC, detail, Some(index));
+                        continue;
+                    }
                     if self.materialize {
                         self.g.set_blob(digest.clone(), bytes.clone());
                     }
--- a/crates/gts/src/reader.rs
+++ b/crates/gts/src/reader.rs
@@ -308,6 +308,31 @@ pub struct BlobPayload<'a> {
     pub codecs: &'a [Codec],
 }
 
+/// One inline blob occurrence a sink's byte ceiling refused before decoding.
+///
+/// Reported so a selecting consumer still learns a candidate was present. It
+/// carries everything decidable without spending the bytes: the container's
+/// declared public metadata (so a representation selector can still match), the
+/// physical source, and the encoded length that was rejected.
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug)]
+pub struct BlobRefusal<'a> {
+    /// Zero-based segment containing this occurrence.
+    pub segment_index: usize,
+    /// Declared content digest, when the container published one.
+    ///
+    /// `None` for a payload with no public digest: refusing happens before the
+    /// decode that would compute it, so no identity is available. Such a blob is
+    /// matchable by representation and provenance, never by digest.
+    pub digest: Option<&'a str>,
+    /// Public metadata available at this occurrence, independent of RDF rows.
+    pub metadata: Option<&'a Value>,
+    /// Original blob byte-string length that was refused.
+    pub encoded_len: usize,
+    /// Which ceiling fired, in the reader's own words.
+    pub detail: &'a str,
+}
+
 /// Sink for [`read_to_sink`] events.
 ///
 /// Term ids in events are segment-local ids. A sink that needs a file-level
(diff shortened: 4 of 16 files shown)


paudley:
# Stage 2 — Gap Remediation for PR #283

## Context

PR #283 began as a one-commit fix for a blocker found by an adversarial audit of
PR #282 (issue #281, merged as `77fe898c`): a selected GTS import refused any
archive that merely *contained* an oversized payload the caller never named,
while the authoritative `import_gts_events` accepted the same bytes.

Stage 2 ran two independent audits — a gap-analyzer and the completion-adversary
— against `22c6c8bb`. Both **independently reproduced a security-relevant defect
that the fix itself introduced**, and the adversary returned `FINDINGS-OPEN` with
three of seven acceptance criteria not demonstrated.

### Already fixed this session (`cb66b727`, pushed)

The fix decided whether a reader diagnostic was a budget refusal by searching its
human-readable `detail` for `"blob exceeds"`. Diagnostic details interpolate
container-supplied bytes — `reader.rs:1591` echoes the header's `gts` field into a
`DamagedFrame` detail. A file whose header magic *was* `"blob exceeds"` therefore
suppressed the magic/version guard, and the bounded importer accepted, with a full
dataset, a container the authoritative importer rejects.

Byte ceilings now report through `CodecError::Limit` → `PayloadError::Budget` →
the `BLOB_BUDGET_DIAGNOSTIC` code, which only the reader's own ceilings emit;
refused frames carry the opaque reason `"over-budget"` rather than `"damaged"`.
The importer matches the code and never reads the detail.

Verified properly: the first regression I wrote **passed under both old and new
code** — a second `BrokenChain` diagnostic masked the admission. Instrumenting the
reader showed why; the test now forges a *header-only* container with a recomputed
self-hash, and **fails against `22c6c8bb`, passes against `cb66b727`**. That
subsumes adversary finding F5 (unpinned cross-crate string coupling).

## Standing Constraints (`.goals`, inviolable)

GREENFIELD-FIRST · RUST-FIRST · SUBSUME/EXTEND/ENHANCE · **LOW/NO OPTIONALITY,
HARD FAILS — no swallowed errors** · RDF 1.2 ONLY · MAXIMAL UTILITY / PERFORMANCE
/ PORTABILITY (wasm-clean).

`.baseline`: adversarial not sycophantic · **NO DEFERRALS — work is NOW** · no
issue/PR numbers in committed files · **when you see a bug, you own it**.

CLAUDE.md: **a refusal is a claim too** (every refusal needs the invalid case *and*
a neighbouring valid one) · **perf changes need a bench**.

The no-deferrals rule plus "you own it" is why the 15 verified defects still
sitting in `main` from #282 are remediated here rather than filed elsewhere.

## The unifying design decision

Five separate findings (F2, F3, F4, A5, A6, and CodeRabbit's open inline comment)
are all the same defect wearing different hats: **a budget-refused payload
currently vanishes.** It never reaches `BlobCollector`, never reaches the
envelope, and leaves only an opaque node indistinguishable from corruption.

Consequences, each demonstrated on the production surface by the adversary:

* **Ambiguity is silently erased.** One archive, two blobs both carrying
  `pub.rep = "wanted"`, one of them digest-less and oversized:
  generous budget → correctly `Err("resolves to 2 blobs")`; tight budget →
  **`Ok`, one blob, silently the wrong answer.** The byte budget changes
  *selection semantics*, the exact opposite of the PR's stated principle.
* **"Absent" and "over budget" are indistinguishable** — both yield
  `rdf-ir-gts-blob-selection … resolves to 0 blobs`, which is a false statement
  about an archive that does contain the blob.
* **The envelope contract is broken.** `gts_import_blobs.rs:102` documents "the
  same scoped native dataset and envelope as `import_gts_events`", and a
  pre-existing test asserts `envelope ==`. It now differs (2 blob records vs 1
  blob record + 1 opaque node), and the new test was narrowed to
  `dataset.owned_quads()` only, hiding it.

**One change closes all of them: make the refusal data.** The reader emits a
refused payload as a first-class event carrying its declared `pub` metadata, its
encoded length, and the ceiling that refused it. Then:

* `BlobCollector` counts a refused payload as a **selection candidate**, so
  `finish()`'s "exactly one" check still sees it → ambiguity fails closed under
  every budget;
* a selector that matches a refused payload reports `rdf-ir-gts-blob-limit`
  naming the payload and the bound, instead of "resolves to 0 blobs";
* the envelope records it, so `envelope ==` holds and the documented contract
  becomes true again;
* `GtsImportWithBlobs` gains `refused: Vec<GtsRefusedBlob>` so a *successful*
  import with an unselected refusal is observable rather than silent — the
  `.goals` "no swallowed errors" requirement.

Digest-less refused payloads have no digest (that is why they were refused before
decoding). They are matchable by `Representation` and by frame provenance, not by
`Digest`; the plan states that limit explicitly and tests it rather than pretending
otherwise.

## Task 1 — Refused payloads become data (closes F2, F3, F4, A5, A6, CodeRabbit)

`crates/gts/src/reader.rs`, `crates/rdf/src/gts_import_blobs.rs`,
`crates/rdf/src/gts_import_sink.rs`.

Executable acceptance:
1. Two blobs sharing `rep="wanted"`, one refused by a tight budget → `Err`
   **under both a generous and a tight budget**, same code.
2. A named over-budget payload → `rdf-ir-gts-blob-limit` naming the bound, *not*
   `resolves to 0 blobs`; a genuinely absent digest → still
   `rdf-ir-gts-blob-selection`. The two must not be confusable.
3. `selected.bundle.envelope == plain.envelope` for the oversized-unselected
   archive — restoring the assertion the current test dropped.
4. A successful import with an unselected refusal exposes it in `refused`.

## Task 2 — 🔴 Snapshot frames decode unbounded (residual #1, CRITICAL)

`reader.rs:596-613`: `let limit = blob.then(…)`, so `payload(frame, false)` takes
the unbounded `decode_chain` arm; `h_snapshot` (`reader.rs:1099-1109`) then emits
every `blobs` entry with **no byte check at all**. A zstd-compressed snapshot
frame carrying a multi-GiB `blobs` map **defeats `GtsBlobLimits` entirely** — in
the API whose entire purpose is bounded retention. The only snapshot test uses an
8-byte payload.

Plumb the ceiling into the snapshot path and check each entry before
`emit_blob`/`set_blob`. Then correct `gts_import_blobs.rs:28`, which currently
claims "Every decoded intermediate is bounded" while `:117` admits the opposite —
**a false bound guarantee must not ship**. Both sides tested: an over-budget
snapshot refuses; an in-budget snapshot still imports.

## Task 3 — Wire and spec fidelity (residual #2, #3, #11)

1. **`len` — DELETE the check** *(user decision this session)*. `validate_length`
   hard-fails on a key `docs/GTS-SPEC.md:2266` defines only as
   `* extension-key => any`, while §22.2 (`:2382-2385`) states a reader **MUST
   NOT** reject solely because it contains an unknown extension key. PurRDF's
   writer never emits it; `"len"` appears exactly once in the tree — as this
   refusal. Remove `validate_length` (`gts_import_blobs.rs:440-453`), its two call
   sites, the `rdf-ir-gts-blob-length` code, and its negative-only fixture
   (`gts_selected_blobs.rs:328`). Add a test that a blob carrying a *disagreeing*
   `len` now imports — the spec-mandated behaviour, and the positive case the
   deleted refusal never had.
2. **External blobs — accurate refusal** *(my call; the user deferred to
   whichever is better)*. Detect `"d"` absent + `pub.digest` present + no prior
   inline occurrence, and report a new `rdf-ir-gts-blob-external` naming the
   digest, instead of today's false *"became selected without a payload; earlier
   unselected bytes are not retained"* — a claim about retention ordering for
   bytes that were never in the file. Document it and test both the external case
   and the metadata-only update it is currently confused with.

   *Rejected: returning external blobs as a bytes-less `GtsImportedBlob` variant.*
   Their digest and metadata are already returned in
   `bundle.envelope.lookaside.blobs` as an `RdfBlobRecord` — that is exactly the
   pre-existing state issue #281 described ("exposes blob references without
   bodies"). What `GtsImportedBlob` uniquely provides is bytes, so a bytes-less
   variant would force every consumer to handle an empty body for information
   already available elsewhere: public API growth with no reachable new
   capability. A truthful error naming the digest serves MAXIMAL UTILITY better.
3. **Digest selector spelling** — selectors must be canonical `blake3:<hex>`;
   bare hex silently matches nothing. Document on `GtsBlobSelector::Digest` and
   test the negative.

## Task 4 — Discharge the refusal contract (residual #12, A4)

The PR's "both sides proven" claim is **false**: its invalid neighbour uses
`Writer::add_blob`, which always emits `pub.digest`, taking the eager branch that
never reaches the modified code. Add the real digest-less invalid neighbour.

Fill every remaining blank so each `rdf-ir-gts-blob-*` code has a negative *and* a
positive case: 1024-digest cap positive; valid-`len` positive (or removal per
Task 3); `rdf-ir-gts-blob-digest` at the metadata≠identity site (reachable via a
31-byte `digest-ref`); a `decoded blob exceeds` case for `reader.rs:629`.
Delete `rdf-ir-gts-blob-provenance` (`:289-298`) — verified **unreachable**:
`sink.frame` fires on all three branches before `fold_frame` with the same
`segment_index`, so the filter can never fail. An untestable refusal does not ship.

## Task 5 — One path, no panics (residual #5, #8, #9, #6, #7, #10)

* Split `import_with_collector` so the selected path returns the collector
  unconditionally — deletes the `Option`, the `.expect()` at
  `gts_import_blobs.rs:142`, and the `self.blobs.is_some()` branch in one move.
* Collapse the two metadata-validation paths (`gts_import_sink.rs:281-289` vs
  `gts_import_blobs.rs:287-288`) onto one snapshot; document which is authoritative.
* Delete the dead unbounded `pub fn decrypt0` (`cose.rs:277-296`). *(The other two
  "dead path" claims were false positives — `decode_chain_with_decrypt` is called
  from `decode_chain:377`, and `reader.rs:610` is the live non-blob path.)*
* `checked_sub` + `debug_assert!` on the `total` arithmetic; assert the
  `segment_head` / `frame_id` / metadata-provenance invariants in `finish()`.
* Name the `64 * 1024` metadata default as a `pub const`.

## Task 6 — `Shapes::dataset()` has zero tests (residual #4)

The entire deliverable for issue #281's criterion A5, with no caller and no test.
Assert no-copy via `Arc::ptr_eq` (idiom already at
`crates/shapes/tests/shared_validation_views.rs:91`) and unchanged prefix
resolution through `parse_shapes`. Fix its self-contradicting doc, which claims
the accessor works "without incrementing its reference count" and then tells the
caller to `Arc::clone` it. *(The "not re-exported" half was a false positive —
`purrdf_shapes::shapes::…` matches existing precedent.)*

## Task 7 — Docs, bench, Makefile (residual #13, #14, F6, A8)

* Rewrite `gts_import_blobs.rs:116-118`, which still documents the pre-fix
  behaviour ("obey both … limits … including when ultimately unselected").
* README sections in `crates/rdf/README.md`, `crates/gts/README.md`,
  `crates/shapes/README.md` — zero exist today; PR #277 set the precedent.
* Criterion bench for the one-pass selected import. **Report-only, no asserted
  speedup.**
* Rename the three `*-native-import-blobs` Makefile targets — named after a
  merged-and-deleted branch, wired into no standing gate. A process-flow
  reference in a committed file, which `.baseline` bans.
* Close CodeRabbit's docstring-coverage warning on the touched files.

## Verification

Per commit: `cargo fmt --all`, `cargo clippy -p <crate> --lib --tests -- -D warnings`,
and the narrowest tests that prove the claim. Before the final push:
`cargo test -p purrdf-gts -p purrdf-rdf -p purrdf-shapes --locked` plus
`cargo check -p purrdf-gts --all-targets --target wasm32-unknown-unknown`.
Full `make check` / `make conformance` remain stage 3 / CI.

**Every new refusal test must be shown to fail against the commit before it** —
the lesson from this session's first regression, which passed against both.

Exit gate: re-run the completion-adversary; stage 2 is not complete until it
returns `PASS` or every remaining item carries verbatim human sign-off.

## Magnitude

7 tasks, ~7 commits, touching `purrdf-gts` (reader, codec, cose), `purrdf-rdf`
(import blobs, import sink, tests, README, benches), `purrdf-shapes` (tests,
README), and the Makefile. Tasks 1 and 2 are correctness-bearing and change
observable behaviour; the rest are hardening, tests and docs.

paudley:
**Task 1 complete** — `2a49b65c`: budget-refused payloads are now selection candidates.

Closes adversary findings F2, F3, F4, gap-analyzer A5/A6, and CodeRabbit's open inline comment on `gts_import_sink.rs:400` — all five were the same defect: a refused payload vanished entirely.

- **Ambiguity no longer depends on the budget.** Two blobs sharing `rep="wanted"` now fail closed under both a generous and a tight ceiling. Previously the tight ceiling deleted one candidate and returned the survivor as if unique.
- **Naming a refused payload** reports `rdf-ir-gts-blob-limit` with the bound, not the false `resolves to 0 blobs`.
- **Refusals are visible on success** via `GtsImportWithBlobs::refused`.
- **Envelope divergence documented and asserted exactly** (`over-budget` opaque node, never conflated with corruption), replacing the quads-only comparison that hid it.

Both behavioural tests were verified to **fail** when refused candidates are not counted.

paudley:
**Stage 2 remediation complete** — `c5c8acdd`, 11 commits.

Three adversary passes. Passes 1 and 2 each found real defects; pass 3 confirmed all six of issue #281's acceptance criteria now **demonstrate on the production surface**, discharged F7–F10, and raised three final findings against the pass-2 repairs. All three are now fixed:

- **F11 over-refusal** — a digest that briefly carried a representation and then gave it up was still refused as external, over a blob the selector does not finally name, in a file `import_gts_events` accepts. Deferred digests are re-checked against final metadata.
- **F12 input-controlled refusal suppression** — dedup trusted a container-*declared* digest. A payload refused before any decode has no checked identity, and the encrypt branch reaches the ceilings with public metadata intact, so a file could name a retained blob on an oversized encrypted frame and erase a genuine ambiguity. Only a digest the reader computed may collapse two candidates.
- **F13** — the README still said a bound refusal is never fatal; that stopped being true for the frame ceiling.

Every behavioural test in this branch was run against the commit before it and confirmed to fail there. That check caught two tests of mine that were vacuous and one whose assertion was simply wrong.

`gts_selected_blobs` 33/33 · `bounded_keyed_blobs` 6/6 · `shared_shapes_dataset` 3/3 · clippy `-D warnings` and rustdoc `-D warnings` clean. `.deficiencies` remains notice-only.

paudley:
fix(rdf): bound selected GTS payload retention without it deciding selection

The selected-blob importer refused any archive that merely CONTAINED an
oversized payload the caller never named, while the authoritative
importer accepted the same bytes. Repairing that exposed a whole class of
defect underneath it: a retention ceiling leaking into selection
semantics, so the size of a memory budget silently changed which payload
a caller received, or refused a container that was valid.

## What was wrong

A byte budget names what an import RETAINS. It was also, in four
different disguises, deciding what a selector COUNTS:

* An unselected oversized payload aborted the whole import. The previous
  test suite asserted that as correct, using an empty selector list.
* A refused payload vanished before the cardinality check, so two blobs
  sharing a representation resolved to two under a generous ceiling and
  to one -- silently, arbitrarily -- under a tight one.
* A compressed `snapshot` frame decoded unbounded, so an RDF frame could
  hand a bounded consumer arbitrarily many payloads and defeat every blob
  ceiling beneath it.
* A refused frame reported through the blob ceiling's diagnostic code,
  which the importer treats as recoverable, so one frame carrying 20,000
  quads imported as zero quads and returned Ok.

Alongside those, three wire-fidelity defects: a `len` refusal that the
GTS specification forbids a reader from making, external blobs reported
with a false claim about retention ordering, and metadata inheritance
that silently stopped at a segment boundary despite documenting
otherwise.

## What changed

Selection reads a projection carrying only what a selector can name.
It is bounded by construction rather than by a budget: a selector is at
most 4096 bytes, so a representation longer than that cannot equal one
and discarding it is lossless for matching. Nothing in it is ever dropped
to satisfy a ceiling, so no ceiling can change a count. The full metadata
map remains bounded, but only for what the caller gets back.

That separation is the substance of the change. Three questions that were
answered from one structure now have one home each:

| Question | Evidence required |
|---|---|
| What is this digest called? | the container's assertion |
| Did the container carry a payload under that label? | any refusal naming it |
| Are two occurrences the same payload? | a hash this reader computed |

A refused payload is now data rather than an absence: it stays a
selection candidate, is reported through `GtsImportWithBlobs::refused`,
and naming one yields the bound that fired instead of a false claim that
the archive does not contain it. A refused frame -- which loses RDF rows
rather than merely declining bytes -- is terminal and carries its own
diagnostic code.

## Files

* `crates/rdf/src/gts_import_blobs.rs` -- the collector: selection
  projection, refusal records, budget accounting, verdicts.
* `crates/rdf/src/gts_import_sink.rs` -- which reader diagnostics are
  recoverable on the IR path, and which are not.
* `crates/gts/src/{reader,codec,cose}.rs` -- typed byte-ceiling
  diagnostics, the frame decode bound, refusal events carrying identity
  and whether it was proved; removes an unbounded `decrypt0` with no
  remaining caller.
* `crates/shapes/src/shapes.rs`, `crates/shapes/tests/shared_shapes_dataset.rs`
  -- the shared-dataset accessor and the tests it shipped without.
* `crates/rdf/README.md`, `crates/shapes/README.md` -- the bounds, what
  they do not cover, and the refusals worth knowing in advance.
* `crates/rdf/benches/gts_selected_blobs.rs` -- one fold against two,
  report-only, asserts nothing.
* `Makefile` -- three focused targets renamed off a merged branch name.

## Requirements satisfied

All six acceptance criteria of the originating issue were demonstrated on
the production surface (`import_gts_events_with_blobs`,
`import_gts_events`, `Shapes::dataset`), not inferred from tests passing:
the importer runs over the same resolver and native builder with the
existing importer's behavior unchanged; selectors are exact digest or
final representation and fail closed; bytes, metadata and independent
frame/segment/head provenance are exact; the encoded, decoded,
retained-total, metadata and frame ceilings each bound before retention;
`Shapes` exposes its retained dataset without reparsing or copying; and
generic conformance stays upstream with focused controls added here.

## Standing constraints

No Cargo features added. No dependency, version or lockfile change. The
wasm32 and MSRV lanes pass. No vocabulary IRI is minted; fixtures use
`example.org`. Serializer and GTS writer determinism is untouched --
nothing here writes containers. No issue or PR reference appears in any
committed file.

## Verification

`cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings`
and all ten hygiene scripts pass on the repository's pinned toolchain.
All CI lanes pass, including `test`, `conformance`, `wasm`, `msrv`,
`capi`, `pytest` and CodeQL.

Every behavioural test added here was run against the commit before it
and required to fail there. That discipline caught two tests of mine that
were vacuous -- one asserted a refusal its budget never triggered -- and
one whose assertion was simply wrong. Without it they would have shipped
looking like coverage.

## Notes for whoever touches this next

**One rule governs this file.** Identity requires proof; naming does not;
a retention ceiling touches neither. A digest is an identity only when
this reader hashed the bytes — an unverified claim may fail a selection
closed but never resolve or redirect one. What a blob is *called* is the
container's assertion, read through a projection no ceiling can drop.
What a ceiling governs is bytes.

That rule exists because the recurring failure here was never a bug, it
was a shape: state answering both "what do we keep" and "what do we
count" leaks the first into the second. Six defects were that one
coupling. If you add a ceiling, check what reads the structure you bound.

The projection's losslessness rests on both bounds being byte lengths
with the same boundary; pinned by a test at exactly 4096 and 4097 bytes.

Declared metadata is kept per digest with its own physical source, so an
occurrence in a later segment inherits the declaration whole and points
at the frame that made it — the reader's own inheritance stops at the
segment boundary.

Externality is only assertable when nothing unidentified was refused: a
payload refused before identification could be the bytes in question.

**One improvement ships without a dedicated control.** The pre-copy
metadata gate now fires on a digest's first declaration. Both orderings
refuse the same container with the same code, so the difference is peak
allocation and has no black-box observation through this API; the test
beside it is named for the refusal it actually pins, not for the fix.

Local `cargo test --workspace` could not be completed on the development
machine (three OOM kills under concurrent load); 259 suites passed with
zero failures before the kill, and the maintainer signed off on CI's
`test` lane -- which runs that exact command on the pinned toolchain in a
clean environment -- as the evidence for that step.

PR #285, opened 2026-09-10
feat(markdown): structural Markdown-to-RDF 1.2 slicer under a caller-supplied vocabulary

### LINKED ISSUE #284: purrdf-markdown: a structural Markdown-to-RDF 1.2 slicer under a caller-supplied vocabulary
## What

A new crate, `purrdf-markdown`, that turns a Markdown document into RDF 1.2 claims along the document's own structure: one document node, one section node per ATX heading and per movement marker, and one unit node per numbered verse or blank-line paragraph. Each unit carries exactly one `xsd:string` literal holding the verbatim byte span of the source, plus its ordinal, byte span, section, and heading lineage. A `## Concordance` table lifts into citation triples from each verse in a range to its canon anchors and source paths. A unit over a byte bound splits at a newline (else at a scalar boundary) with a snapped overlap, never inside a scalar and never across a heading.

The output is a pure, deterministic function of the bytes and a declared profile whose parameters are inside its identity (derived with `derive_chunking_contract_id`), so a change to the split law or its constants re-mints the profile rather than drifting.

## Why here

It is a format converter into RDF 1.2, and PurRDF is the RDF engine the consumers already depend on. The first consumer is a private downstream corpus (its ADR 0023, a source-projected workspace for that corpus's document collection), which built the slicer purrdf-shaped from the start so it could move upstream verbatim. Two stores and an authoring step then slice a document the same way, and a `.purremb` binds to the sliced text by construction on every side.

## PurRDF law

The crate mints no vocabulary. Every class, predicate, and datatype IRI, and the base under which node IRIs are minted, is caller-supplied through a `Vocabulary` (with `Vocabulary::under(base)` deriving the full set from one base IRI), and the profile name is caller-supplied too. The first consumer's profile reproduces its existing contract id and its claims byte for byte, proven by two goldens in the vectors.

## Scope

One dependency (`purrdf-core`), `purrdf-rdf` as a dev-dependency for the parse-and-canonicalize vector; no I/O, no Cargo features, rlib only, wasm-clean by construction (the same std subset as `purrdf-text`). Registered as a release crate and in the bootstrap ledger; the publishable count moves to twenty-two.

A PR follows.

paudley:
Closes #284.

## Summary

Adds `crates/markdown` (`purrdf-markdown`): a structural Markdown-to-RDF 1.2 slicer. A document becomes a graph of its own structure — document, sections (ATX headings and movement markers), units (numbered verses and paragraphs) — each unit carrying exactly one verbatim `xsd:string` literal of its byte span, plus ordinal, byte span, scalar span, content digest, section, and heading lineage. A `## Concordance` table lifts into RDF 1.2 reified citations. Output is a pure deterministic function of the bytes and a declared profile whose parameters live inside its identity.

The crate ships a standalone specification, [`crates/markdown/SPEC.md`](./crates/markdown/SPEC.md), offered to the Markdown and RDF communities. We surveyed the standards that map documents to RDF and found none that combines determinism, byte-span addressability, and law-content-addressed identity for Markdown; if one exists we would rather hear of it than claim it does not.

## The model is the product

`analyze()` returns a typed stand-off model (`Document`/`Section`/`Unit`/`Citation`) — untouched source bytes plus a typed layer of range annotations — and `render()` is one projection of it into claims. The model carries what the flat triple projection cannot: unflattened concordance rows, malformed- and unmatched-row accounting, containment-lattice queries (`unit_at`, `covering`, `units_under`), the thirteen Allen span relations, and content anchors for deterministic re-anchoring. The Markdown dialect reader sits behind a documented seam; splitting, identity, rendering and the lattice are dialect-independent.

## Vocabulary

The processing core mints no vocabulary; this is an import surface and is deliberately opinionated. `Vocabulary::standard()` mints the specification's designated namespace `https://w3id.org/purrdf/markdown#`, and `Vocabulary::under(base)` remains for deployments that want their own. Every vocabulary IRI, the source id, the canon base, and every minted anchor IRI is validated under the workspace IRI law (reached through a widened `purrdf-core` re-export — no new dependency edge); anchor minting is containment-checked by a relativize round-trip, so path-traversal and authority-extension escapes refuse with the concordance row's verse range while CJK anchors lift, because the law is IRI-lawfulness, not ASCII.

## Identity

Node IRIs are content-addressed. The profile's contract id is derived over a stage description that states the complete law — dialect grammar, split law, concordance law, emission law — so any change to any rule re-mints rather than drifts. The profile also yields a PURREMB applied-stage identity whose parameters are that same description verbatim, so the two identities cannot move independently; `Unit::text_chunk_target` and `verify_unit` bridge to the kernel's own verification law.

## Refusals, each with its valid neighbour

Bounds under the widest scalar, an overlap that swallows its window, control characters in a profile name, empty or relative source ids, empty or unlawful canon bases, and unlawful minted anchors all refuse at the entry seam. A leading byte-order mark is *not* refused — it is skipped for structure detection only, spans and literals staying verbatim over the original bytes. A concordance row naming an enormous verse range does not hang: lifts are decided by walking the document's own verses, and unmatched verses are stated as maximal runs bounded by the lifts.

## Registration

Workspace member; fifteenth entry of `PURRDF_RELEASE_CRATES` (after `purrdf-rdf`, its one dev-dependency); the bootstrap ledger; `docs/RELEASE.md`; the publishable count 21 → 22 in the Makefile and CI; both README crate maps and `AGENTS.md`; re-exported by the umbrella facade as `purrdf::markdown`, with a facade test that slices a document end-to-end through the facade alone.

## Verification

`cargo test -p purrdf-markdown`: 119 green (19 unit, 100 vectors) plus an explicitly-ignored golden-regeneration writer. `cargo check -p purrdf` and `-p purrdf-core` green; `RUSTDOCFLAGS=-D warnings cargo doc` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo fmt --all --check` clean; `cargo bench -p purrdf-markdown -- --test` compiles and runs all six benchmarks (report-only, no asserted speedups); `make check` green end to end; `make conformance` GREEN at 4488 with every scoreboard row unmoved. Counts and gate results in this paragraph are as of the branch head, re-run rather than carried forward from an earlier stage.

The `make check` run above is end to end, its terminal `$(MAKE) wasm` step included — that step built all twenty-two release crates for `wasm32-unknown-unknown` and finished clean, so the wasm leg was run locally rather than left to CI. CI runs the same gates independently and is authoritative for the record.

## Vectors

Determinism; the node model including movements as siblings; exactly one `xsd:string` per unit; lineage and spans; the split law at its narrowest bounds (newline-preferred cuts, the scalar-boundary fallback, the continuation snap law, overlap-aware coverage, and the strict bound); input shapes (CRLF, empty document, headings at EOF, empty titles, tab separators, the `u64` verse bound, byte-order marks); concordance lifting with and without a canon base, with row pairing preserved through reification; every claim parses and canonicalizes under `purrdf-rdf`, reified claims included; identical paragraphs yield distinct IRIs; substitution and insertion signatures; the two identities; kernel-verified chunk targets; scalar↔byte conversion; and the specification's own clauses.


paudley:
## Stage 1 complete — what landed, in ten commits

The branch was rebuilt from its base as a single clean commit and then advanced one task at a time:

| Commit | What it settled |
|---|---|
| `a9b19ceb` | the slicer, with a neutral synthetic corpus and its goldens |
| `aeb808c0` | entry-seam refusals: bounds, overlap, profile name, source id, byte-order marks |
| `038ba07b` | IRI seams under the workspace IRI law, with containment-checked anchor minting |
| `b1fece24` | the split law proven strict and pinned at its narrowest bounds |
| `c51b9317` | the typed stand-off model, the dialect seam, and the specification |
| `7937c858` | the emission law: reified citations, scalar offsets, digests, canonical escaping |
| `ba1fd737` | two identities, one law, and spans the kernel can verify |
| `dd2a5616` | the bench, one-pass section assembly, and the bounded verse walk |
| `833fa149` | the umbrella facade, the front page, and the corrected crate maps |
| `db03673b` | the zh-Hans mirror |

### Two defects found and fixed that were not in the original scope

**The concordance lift destroyed row pairing.** Anchors and sources from different rows merged into one per-unit bag and `dedup` collapsed multiplicity, so a verse covered by two rows could not say which source belonged to which anchor. The golden diff shows the repair directly: a source named by two overlapping rows is now stated once per row, through content-addressed citation nodes carrying `rdf:reifies` triple terms.

**A concordance row could hang the slicer.** Lifts were decided by walking the row's declared verse range, which a document controls; a row spelling the full `u64` range would not terminate. Lifts are now decided by walking the document's own verses, and `Citation::unmatched()` states maximal runs bounded by the lifts rather than one verse at a time — the accurate long-form answer to that row is ~1.8×10^19 entries, so the shape itself was the second half of the hazard. The bench now generates a full-range row, so a regression cannot complete.

### Least confident, and why

The **anchor-containment rule for fragment-style canon bases**. `BaseIri::relativize` accepts `#../x` as contained under `…/canon#` — correct, because a fragment climbs out of nothing — while refusing the same anchor under a path-style base, where it escapes. That asymmetry is right, but it is subtle, and it is the kind of rule a future reader may "simplify" into over-refusal. It is stated normatively in the specification and pinned by paired vectors on both sides; if it is ever wrong, it will be wrong for fragment bases specifically.

Second: the **specification's conformance clause** names this crate's vector suite as the reference tests. That is honest for a first publication, but it means the suite is now a public contract surface, not just an internal gate.

### What you may not know but should

- **The two goldens are frozen law guards** from the emission-law commit onward. Any change that alters them altered the law; the correct response is to stop, not to regenerate.
- **`Profile::stage_bytes()` is public API in the strongest sense**: its exact bytes are the preimage of every contract id in the world. Editing that prose re-mints every downstream identity. This is the design working, but it means the string is not a comment.
- **The `w3id.org` namespace does not resolve yet.** Minting the IRI does not wait on the redirect (RDF IRIs need not dereference), but the redirect PR to `perma-id/w3id.org` is an out-of-repo act — snippet below.
- **`Vocabulary::under` gives `heading`/`dt_heading` and `lineage`/`dt_lineage` one local name each in two roles.** Legal RDF, deliberate term economy, documented — deployments building OWL or SHACL over the output should set those fields explicitly.
- **The specification version line reads 2.0.0-draft**, tracking the workspace's next release rather than the 1.2 originally planned.

### The w3id redirect (for the namespace to resolve)

A PR to `perma-id/w3id.org` adding `purrdf/.htaccess`:

```apache
# https://w3id.org/purrdf/markdown# — PurRDF structural Markdown slicing law
RewriteEngine on
RewriteRule ^markdown/?$ https://github.com/Blackcat-Informatics/purrdf/blob/main/crates/markdown/SPEC.md [R=302,L]
RewriteRule ^markdown/(.*)$ https://github.com/Blackcat-Informatics/purrdf/blob/main/crates/markdown/SPEC.md#$1 [R=302,L]
```

Adjust the target once the specification has a published home (a DOI landing page would be the natural one).

paudley:
# Stage 2 — Gap Remediation for Issue #284 / PR #285

## Context

Stage 1 landed `purrdf-markdown` in ten commits on `feat/markdown-slicer`. Stage 2
ran three independent audits — a context pass over the issue and every review
thread, an adversarial gap analysis, and the completion adversary (which built
**external consumer crates** and drove the public API and the `purrdf::markdown`
facade rather than the test harness).

**Result: 12 of 14 acceptance criteria are DEMONSTRATED-IN-PRODUCTION**, including
rebuilding both committed goldens byte-for-byte from outside the test suite. Two
adversary findings are open and binding, and the gap analysis found a third
merge-blocker plus a set of silent-drop defects. Deferral hunt: **zero hits**
anywhere (diff, commits, PR, issue comments); `.deficiencies` is clean. The
failures are the inverse of laundering — things described as done that the code
does not deliver.

Three environment artifacts were disproven and are **not** gaps: the
`redundant_clone` clippy errors and the wasm32 `E0463` failure I first saw came
from a shell whose default nightly and `RUSTC_WRAPPER=sccache` shim leaked past
the `rust-toolchain.toml` pin. Under the pinned toolchain, `purrdf-validate`
clippy is clean and `cargo check -p purrdf-markdown --target
wasm32-unknown-unknown` finishes. `make doc` and `make conformance` are GREEN
(conformance: all suites pass or ledgered).

### User decisions taken during this analysis (binding)

1. **`is_absolute_iri` needs hardening** — "definitely needs hardening".
2. **Trim law: tolerate leading spaces** — CommonMark-style, up to 3 spaces before
   a marker; 4+ spaces is not a marker.
3. **Identity: split the two preimages** — chunking identity and emission identity
   become separate content addresses.
4. **gmeow**: the ontology is at `~/Active/gmeow-ontology`; the concept is
   `logic:Plan` = `https://blackcatinformatics.ca/logic/Plan`, defined by
   `<https://blackcatinformatics.ca/gmeow/slices/logic>`. **"gmeow:logic is fully
   provable"** — §10 must say that the recommendation rests on a formally provable
   layer, not merely a richer vocabulary.

## Gap list

### CRITICAL — merge blockers

**G1 — `render()` walks around every entry-seam refusal.** `Document`
(`crates/markdown/src/model.rs:737`) stores no binding to the profile it was
analyzed under, so `render(&Document, &Profile)` accepts any profile.
`analyze()` with no canon base (anchors never walked, `lib.rs:335`) followed by
`render()` *with* a canon base emits the exact traversal anchor §4.2 exists to
refuse — proven:
`<…/cites> <https://example.org/canon/../../etc/passwd>`, including inside a
reified triple term. This defeats the containment apparatus and its three paired
vectors, and silently re-opens two CodeRabbit findings the PR reports as closed.
**Fix:** `Document` retains its validated profile; `render(&Document) -> Vec<Claim>`
takes one argument. Vector: emitted anchor IRIs are exactly those `analyze`
admitted; the bypass sequence no longer compiles.

**G2 — `make check` is RED on this branch and green on `main`.**
`scripts/check-iri-resolver-singleton.py` (step 4, `Makefile:54`) flags
`crates/markdown/src/profile.rs:603`. Verified: exit 1 here, exit 0 on main. The
gate's delegation detector is `DELEGATES = re.compile(r"\bpurrdf_iri\b|\bBaseScope\b")`
and the body reads `BaseIri::parse(...)` through the *new* `purrdf-core` re-export,
so the token never appears. **CI never runs this script** (`grep
check-iri-resolver-singleton .github/workflows/` → no hits), which is why PR #285
is green. **Fix:** widen `DELEGATES` to recognise `BaseIri` (protects every future
producer routing through the kernel re-export) **and** add the script to the CI
`workspace` job so a `make check` step no CI job runs cannot rot again.

**G3 — identity does not answer for the bytes emitted.** Three profiles differing
only in `canon_base` produced *identical* `contract_id`, `stage_bytes`, and
PURREMB stage id, with different graphs. Worse, `citation_iri` hashes the row line
and unit id but **not the minted anchor**, so one citation node IRI asserts three
mutually exclusive reified edges, and the `cites` object silently changes kind
(typed literal ↔ IRI). Both committed goldens carry the same pinned id while
differing by 320 bytes. This contradicts the issue's criterion, SPEC §5's MUST
("two runs that agree on the id MUST agree on every byte they emit"), and
`stage_bytes`' own doc.
**Fix (user-chosen: split the preimages).**
- `chunking_bytes()` — dialect grammar + split law + constants only. Feeds
  `purremb_chunking_stage()`. Adding a canon base or renaming a predicate then
  does **not** invalidate a consumer's embeddings (today it does, because the one
  preimage carries the vocabulary too).
- `emission_bytes()` — `chunking_bytes()` + vocabulary + **`canon_base`** +
  emission law. Feeds `contract_id`, the id written into the graph as
  `sliceProfile`.
- `citation_iri` preimage gains the minted anchor.
- Replace `tests/slicer.rs:1181` ("ignores the canon base") with a vector
  asserting the new law, plus one proving equal contract ids ⇒ byte-identical
  graphs, and one proving a canon-base change moves `contract_id` but **not** the
  chunking id. Re-mint goldens; update SPEC §5 (delete line 348) and §11.

**G4 — SPEC §2 states a recognition law the code does not implement, and that
clause is inside the contract-id preimage.** §2:98 says recognition operates on
*trimmed* text; only the trailing edge is trimmed, so `" # Title"` yields zero
sections, `"  1. one"` is a paragraph, `" ⁂ *m*"` is a paragraph. An independent
implementor following the spec derives the same id and emits a different graph —
the failure §5 promises cannot happen.
**Fix (user-chosen: tolerate leading spaces).** Up to 3 leading spaces before
`#`, `⁂`, or a verse number are ignored for recognition; 4+ spaces is not a
marker. Spans stay verbatim over the original bytes (the leading spaces remain
inside the unit's literal, exactly as the BOM rule works). Update
`dialect/markdown.rs` (`atx_heading`, `movement`, `verse_number`), the
`stage_bytes` dialect line, and SPEC §2. Vectors for all four rows above plus the
4-space negative case and a 3-space positive case. Re-mint.

### HIGH

**G5 — a concordance organised into subsections silently loses every row.**
`in_concordance()` (`dialect/markdown.rs:198`) tests `stack.last()` only, but §4
defines "inside" as containment. A `### Book one` under `## Concordance` makes
every row vanish — not refused, not malformed, not unmatched, not reported.
**Fix:** walk the stack; true if any open section is the concordance heading.
Vector: subsectioned concordance lifts what the flat form lifts.

**G6 — a row with sources and no anchors mints an unreachable orphan node.**
`claims.rs:279-299` emits `<citation> <canonSource> "…"` with no `rdf:reifies`,
no back-edge, no `rdf:type`. The only documented traversal (§11) never reaches it.
**Fix:** state the case normatively and either emit `canonSource` only when the
row lifted an anchor, or give every citation node an unconditional back-edge to
its unit and a type. Vector either way.

**G7 — a GFM-escaped pipe silently destroys a row's content.** `cells()`
(`dialect/markdown.rs:302`) splits on bare `|`. `` | 1 | `a\|b.ttl` | `x` | ``
lifts onto verse 1 with **empty** sources and anchors — not malformed, not
unmatched, no report. `\|` is ordinary GFM.
**Fix:** honour `\|` (split on unescaped pipes, unescape cells). Vector plus a
valid neighbour.

**G8 — the designated namespace ships two IRIs that are both predicate and
datatype.** `LOCAL_NAMES` lists `heading` and `lineage` twice. Defensible for
`under(base)`; not for a namespace SPEC §6 tells the world to exchange under —
no OWL/SHACL description can declare one IRI as both `owl:DatatypeProperty` and
`rdfs:Datatype`. (I declined this in stage 1 as term economy; the directive that
made this namespace canonical changed the calculus.)
**Fix:** distinct datatype local names (`headingText`, `lineagePath`). Re-mint.

**G9 — `docs/RELEASE.md:123` states a false release-lane fact.** It says nothing
in the workspace depends on the crate "so the stop is at the end of the set"; the
umbrella now depends on it and `purrdf-wasm` depends on the umbrella, so the stop
is at the flagship crate. **Fix:** rewrite naming `purrdf` as the first dependent.

**G10 — no CHANGELOG entry** for a new publishable crate, a new umbrella module,
and three new `purrdf-core` re-exports. **Fix:** `[Unreleased] ### Features`
entry.

**G11 — stranded doc comments in the umbrella** (`crates/purrdf/src/lib.rs:286`):
inserting the facade test split an existing doc comment, so both
`facade_exposes_markdown_slicer` and `facade_exposes_full_text_search` now carry
the wrong description. My error in stage 1. **Fix:** restore both.

### MEDIUM

**G12 — `is_absolute_iri` conflates "unparseable" with "scheme-less"** (user-
directed hardening). `profile.rs:603` is `BaseIri::parse(c).is_ok()`, discarding a
rich `IriError`; `"http://[not-an-ipv6"` and `"https://example.org/%"` both report
`RelativeSourceId`, whose `Display` says "has no scheme" — false. Three call
sites each invent their own single explanation.
**Fix:** fallible `absolute_iri(&str) -> Result<Iri, IriError>`; genuinely
scheme-less keeps `RelativeSourceId`; malformed gets a distinct variant carrying
the `IriError` as cause (the pattern `TamperedUnit` already uses). Vectors for
truncated percent-encoding, bad scheme, disallowed char, bad authority, the
genuinely relative case, and a valid neighbour. Pairs with G2.

**G13 — the "specification's own clauses" vector is a substring-presence test**
(`tests/slicer.rs:3179`), 30 `SPEC.contains(...)` assertions; it cannot detect G3
or G4 and the PR's Vectors list implies behavioural coverage. **Fix:** rename it
honestly as an anti-rot check and add real clause-to-behaviour vectors for §2,
§4.2 and §5.

**G14 — `section_iri` has no in-crate consumer** (the projection mints via
`node_iri_of_digest`). The adversary proved it mints the right IRIs, so it is not
dark, but it is a second unpinned path to section identity. **Fix:** add the
`section_iri` ⇄ `node_iri_of_digest` equivalence vector (mirroring
`identity.rs:156` for units).

**G15 — SPEC §10 cites gmeow with no resolvable IRI.** **Fix:** cite
`logic:Plan` (`https://blackcatinformatics.ca/logic/Plan`, defined by
`<https://blackcatinformatics.ca/gmeow/slices/logic>`) and state why it is
recommended over `prov:Plan`: it is a transaction program with action schemas,
goal-expression success criteria and an explicit success quantification, **and it
lives in a formally provable layer**, so a recorded slicing plan carries
machine-checkable structure rather than descriptive annotation. **Recast the
PROV-O sentence**: the current "PROV-O MAY be used instead or as well" frames
them as rivals, which is wrong — gmeow carries `owl:sameAs` bridges to PROV
where interop is wanted (the ontology already ships bridge views of the
plan/enactment spine onto OPMW and OBI), so a deployment loses no PROV
compatibility by recording the provable form. Say that, and note honestly that
the gmeow namespace is not yet publicly dereferenceable.

**G16 — `https://w3id.org/purrdf/markdown#` does not resolve.** Out-of-repo act
in a third-party repo; the `.htaccess` snippet is drafted on the PR. Tracked as a
**stop-and-ask**, not a code gap — it does not block the other fixes.

### LOW

- **G17** — `iri_forbids` rejects `≤ 0x20` while `purrdf-core` also escapes the C1
  block; the gap is closed only incidentally because `ucschar` starts at U+00A0.
  Pin with a vector.
- **G18** — `triple()` writes the predicate bare while the subject goes through
  `emit_term`; consistent with `purrdf-core` but two spellings of one IRI.
  Document or unify.
- **G19** — the heading-stack law lives in both `Walker::open_section` and
  `assemble_sections`; they agree today. De-duplicate or document the one path.
- **G20** — the zh-Hans back-translation transcript is not on the PR; post it.

## Order of work

Dependencies and re-mint economy drive the sequence. Commit and push after each.

1. **G2** (gate + CI wiring) — unblocks `make check`.
2. **G12** (`absolute_iri` hardening) — pairs with G2, same file.
3. **G1** (`render(&Document)`) — structural; emission unchanged, goldens stable.
4. **G5**, **G7** (dialect silent drops) — fixture gains the shapes; re-mint if it does.
5. **G4** (leading-space tolerance) — dialect + `stage_bytes` + SPEC §2; re-mint.
6. **G6** (orphan citation node) — emission law; re-mint.
7. **G3** (preimage split + `citation_iri` anchor) — re-mint.
8. **G8** (distinct datatype local names) — re-mint.
9. **G9**, **G10**, **G11** (docs, changelog, umbrella doc comments).
10. **G13**, **G14**, **G15**, **G17**–**G20**.
11. Re-run `make check`, `make wasm`, `make doc`, `make conformance` — all four to
    exit 0, unpiped, under the pinned toolchain.
12. **Re-dispatch the completion adversary** as the exit gate. Stage 2 does not
    complete until it returns PASS or every residual item carries verbatim human
    sign-off.

## Files that will change

- `crates/markdown/src/`: `profile.rs` (identity split, `absolute_iri`,
  `LOCAL_NAMES`), `model.rs` (`Document` retains its profile), `claims.rs`
  (`render` arity, citation emission), `identity.rs` (`citation_iri` preimage),
  `dialect/markdown.rs` (leading spaces, `in_concordance`, `cells`),
  `error.rs` (new variants), `lib.rs` (seam), `SPEC.md` (§2, §5, §6, §10, §11),
  `tests/slicer.rs`, `tests/fixtures/*` (re-mint), `benches/slicer.rs` if the
  generator needs the new shapes.
- `scripts/check-iri-resolver-singleton.py`, `.github/workflows/ci.yaml`.
- `crates/purrdf/src/lib.rs` (doc comments), `docs/RELEASE.md`, `CHANGELOG.md`.

## Verification

Per fix: `cargo test -p purrdf-markdown`, `cargo clippy -p purrdf-markdown
--all-targets`, `RUSTDOCFLAGS=-D warnings cargo doc -p purrdf-markdown --no-deps`,
and `git diff -- crates/markdown/tests/fixtures/` empty except where the fix
deliberately re-mints (each such commit states the diff's shape).

End to end: all four `make` gates green under the pinned toolchain; the external
consumer demonstrations re-run (the adversary's `/tmp` consumer crates drive the
public API and the facade, including rebuilding both goldens from outside the
test harness); CI green on the PR; completion adversary returns PASS.

## Out of scope, explicitly

Nothing from the issue is deferred. G16 is the only item this branch cannot
close, because it is a pull request against a repository this project does not
control; it is surfaced to the user as a decision, not logged as a deficiency.

paudley:
## Gap G2 closed — `make check` was red on this branch, and a whole class of gates ran nowhere

**The gap.** `make check` step 4, `scripts/check-iri-resolver-singleton.py`, failed on this branch and passed on main. It refused `crates/markdown/src/profile.rs` for re-implementing RFC 3986 resolution — but the code genuinely delegates to the shared IRI law. The gate tested for the *crate name* `purrdf_iri`, which this crate structurally cannot write: its only runtime dependency is the kernel, which re-exports the law. A producer routing correctly through the kernel could never satisfy the test.

**Why CI was green anyway.** No workflow ran that script. A `make check` step no runner runs is a gate that rots — so the branch could be green on the PR and red on a contributor's machine.

**The fix.** The detector now recognises the shared layer's own types (`BaseIri`, `Iri::resolve`), each of which has exactly one definition in the workspace. `parse_iri` deliberately stayed out of that alternation — three crates own a local lexer method by that name, and a bare token would let a hand-rolled predicate launder itself; it is instead recognised two-condition (imported from the shared layer AND called free-standing, so `self.parse_iri()` never counts). Every rule that matches the resolution *arithmetic itself* — `remove_dot_segments` by name, the section 3.1 character class, the dot-segment popping loop — stays unconditional, so naming a shared type buys no amnesty. Proven with seven probes: three hand-rolled shapes still fail, four legitimate delegations pass.

**The sweep.** Rather than fix the one instance, every command `make check` runs was diffed against every command the workflows run. Four gates ran in neither place or only on tag pushes, and are now wired into CI: the resolver singleton, publish order, the core hygiene target, and the hand-written-numbers gate — which nothing ran at all, despite this branch having edited it. That last one also joins `make check`, so the local gate stays predictive of CI rather than creating the mirror asymmetry.

Remaining asymmetries are legitimate and documented: the wasm job inlines the identical build (crate lists compared mechanically, 22 each, identical), and the oracle, conformance and binding gates need toolchains `make check` deliberately omits.

21 of 21 hygiene gates pass. Commit `cdc27f4b`.

paudley:
## Gap G12 closed — one answer was standing in for five questions

**The gap.** `is_absolute_iri` asked the IRI law a yes-or-no question and threw away everything else it said. Three seams — the source id, every vocabulary field, and the canon base — each had only one word left to explain a refusal, so a truncated percent-encoding (`https://example.org/%`) and an unterminated address literal (`http://[not-an-ipv6`) were both reported as carrying no scheme. That is a false statement about strings that plainly carry one, and it is the shape of refusal a corpus owner cannot act on.

**The fix.** One classifier now owns the distinction for the whole crate: a reference the grammar read whole but found scheme-less refuses as relative, and anything the grammar could not read at all refuses as malformed, carrying the law's own finding verbatim. Three new error variants (`MalformedSourceId`, `MalformedVocabulary`, `MalformedCanonBase`) carry the cause rather than restating it — the same pattern the tampered-unit refusal already used to carry the kernel's finding. The classifier also hands back the parsed base, so the profile seam no longer parses the canon base twice.

The mapping is principled rather than case-by-case: the base parser is *parse, then check the scheme*, so exactly the two findings about a string the grammar read whole mean "relative"; everything else — including the empty string, which is not a scheme-less reference but not a reference at all — means malformed.

**Proof it is not one message wearing five hats.** Each vector asserts a distinct diagnostic code, that the rendered message never claims "no scheme" for a string that has one, and that the message *ends with the law's finding verbatim* — so no seam can quietly reword the IRI law. Every refusal ships with its valid neighbour; the declared vocabulary, an absolute source id, and a percent-encoded canon base all still slice and lift.

80 tests green. Goldens byte-identical — this fix touches refusal paths only, so no lawful document's emission moved. Commit `8cd6dd7f`.

paudley:
## Gap G1 closed (merge blocker) — the projection could be handed a law the document never passed

**The gap.** The crate's load-bearing safety claim, stated in three places including the specification, was that projecting claims is infallible *because a model cannot exist without having been admitted*. The model carried no record of the law it was admitted under, and the projection took a second law as an argument. So:

1. admit a document with **no canon base declared** — anchors are never walked, because there is nothing to mint them against;
2. project that model **under a base** — and the anchor is minted with no containment check.

A concordance anchor of `../../etc/passwd` reached the graph as `<https://example.org/canon/../../etc/passwd>`, including inside a reified triple term — the exact shape the containment law exists to refuse, past an apparatus with three paired vectors defending it.

**The fix.** The model now carries the law it was admitted under, and the projection takes the document alone. The invariant is structural rather than documentary: the only constructor validates the profile and walks every anchor under it before assembling, and the projection reads the law off the document with no parameter through which another could arrive.

An owned profile rather than a borrowed one, deliberately: the model borrows only the source bytes, so a borrowed profile would tie its lifetime to that borrow and force lifetime gymnastics on any consumer that builds a profile locally. A partial handle was rejected for a sharper reason — consumers legitimately want the whole profile, and handing back a fragment would send them to their own copy, which is precisely what caused this bug.

**Proof.** The central vector admits the hostile anchor with no base, projects it, and asserts the minted IRI appears in **no** claim's turtle — neither asserted nor reified — for either base shape, while the anchor remains a typed literal as the no-base law requires. Its valid neighbour proves lawful anchors, including a CJK one, still lift identically.

83 tests green. Goldens byte-identical: for lawful input nothing about emission moved. Commit `43221e2c`.

paudley:
## Gaps G5 and G7 closed — two ways a concordance row vanished without a word

Both are the failure the repo's law and the specification forbid by name: a row that is not refused, not reported malformed, not reported unmatched, and not lifted. It simply ceases to exist.

**A row inside a subsection of the concordance was lost.** The reader asked only the innermost open section whether it was the concordance, so a `### Book one` beneath `## Concordance` swallowed every row under it — `citations=0 malformed=0 unmatched=0`. The specification defines "inside" as containment, so the reader now asks the whole open-section stack, which is the idiom the movement-level lookup already used. The other half of the law survives and is pinned: a table after the concordance closes still lifts nothing, so this did not become over-refusal in the other direction.

**A cell holding an escaped pipe was split at that pipe.** `` `a\|b.ttl` `` is ordinary GitHub-flavoured Markdown, and it produced a row that still counted as lifted while both of its cells were silently emptied. The reader now honours exactly two escapes, which close each other: an escaped pipe is a cell value and never a delimiter — not even the row's leading or trailing one — and an escaped backslash is a backslash, which must be recognised or `\\|` would be misread as an escaped pipe. Every other backslash is content and keeps itself, so the law is total and no third rule can drift in later.

**Both fixes were proven load-bearing** by reverting each in turn and confirming the intended vectors fail — a fix whose test passes with the fix removed is not a fix.

**On the goldens.** Both laws live inside the contract-id preimage, so this re-mints the identity deliberately — the first such re-mint in this series. Because the contract id is a field of every node preimage, every unit, section and citation IRI moves with it; normalising the hex away leaves 36 structural lines, and each is the fixture's new shapes: the concordance subsection, its prose unit, and the five new lifts with their reified citations and source paths.

89 tests green. The fixture and the bench generator both carry the new shapes, so the corpus and the performance harness exercise the laws rather than merely the unit tests. Commit `45f33cd8`.

paudley:
## Gap G4 closed — the specification described a law the reader did not implement

**The gap.** The specification said structure is read on a line's *trimmed* text, and the reader trimmed only the trailing edge. So `" # Title"` opened no section, `"  1. one"` was not a verse, and `" ⁂ *m*"` was not a movement — all of them ordinary prose.

That would be an ordinary documentation bug except for where the clause lives: inside the stage description, which is the preimage of every contract id. An independent implementor who read the specification, implemented trimming, and derived the same identity would have emitted a **different graph under an identical id** — the one failure the identity section promises cannot happen. A published standard cannot carry that.

**The law now implemented.** A marker is read behind a leading run of up to three spaces. Four or more — or a tab anywhere in that run — opens no marker and leaves ordinary content. The tab clause is stated on the line's own bytes rather than by column arithmetic: a run of up to three spaces followed by a tab reaches the fourth column under the usual tab-stop rule, so declaring a tab in the run to be no marker means **no implementation ever has to expand a tab** to decide. Four-or-more is explicitly *not* a code block — this dialect defines none; it simply declines to read a marker that far in.

**The tolerance is recognition only.** The leading run stays inside the unit's span, inside its verbatim literal, and inside a section's heading span; a title is what it would have been had the run not been written. This follows the byte-order-mark precedent, and the specification now states exactly where the two part company: the byte-order mark falls *before* every span and so enters no digest, while the leading run is *inside* the span and therefore inside the content digest. A vector asserts that.

**The corpus discriminates.** The fixture now carries a movement behind two spaces, a heading behind one, a verse behind three — and a four-space line that would open a level-one heading under the wrong law, but is content under this one. One count moved upward and was updated rather than relaxed; one existing assertion was strengthened to check the section span opens at the line's first byte, run included.

95 tests green. Goldens re-minted deliberately: same line count, every identity moved, spans shifted by the fixture's new bytes. Commit `04fac48a`.

paudley:
## Gap G6 closed — a citation node is never an orphan

**The gap.** A concordance row that names source paths but no anchors minted a citation node carrying its source and nothing else: no type, no reification, and no edge from the unit it belonged to. Nothing in the graph pointed at it. The only traversal the specification documents starts at a unit, so that node was unreachable — the row's source was formally present in the graph and practically lost to anyone walking it.

**The choice.** Two shapes were available: emit the source only when the row lifted an anchor, or give every citation node a type and a back-edge unconditionally. The first was rejected because it makes the graph state *less* than the document does — the row's canon source would vanish entirely, leaving the typed model as the only place a fact of the document lives, which is exactly the projection weakness the specification already flags. The second makes the invariant structural instead of conditional: no shape of row can mint a node the graph cannot reach. The back-edge direction matches the shape already used for a unit's document and section.

**The cost is stated, not hidden**: two triples per citation node on every document, including documents whose rows all name anchors. That sentence is in the specification, the crate README, and the module docs.

**Proven by a walk, not a count.** The new vector runs a breadth-first search over every emitted triple, taking edges from plain IRI objects *and* from IRIs inside triple terms, while deliberately excluding datatype IRIs — otherwise the walk would say yes to anything. It finds citation nodes structurally, as subjects no claim is about, rather than by name, then asserts every one of them sits exactly one step from a unit, under three different profiles. A future change that reintroduces the orphan shape fails that test rather than passing quietly.

The specification gained the guarantee normatively, including the sentence a conforming implementation must satisfy: its emitted graph contains no citation node that no unit reaches. The fixture gained a row with an empty anchor cell, so the golden corpus and the performance harness both carry the shape.

98 tests green. Goldens re-minted deliberately. Commit `a2305d06`.

paudley:
## Gap G3 closed — two preimages, because a chunk is not a claim

**The gap.** One preimage fed both the identity written into the graph and the identity a consumer keys embeddings on, and it was wrong in both directions at once.

It omitted the canon base. Three profiles differing only in that field emitted three different graphs and reported the same contract id — and because a citation node's identity did not carry what it reified, one node asserted three mutually exclusive reified edges, with the cited object silently changing kind between a typed literal and an IRI. A consumer merging two stores that both truthfully reported that id got a contradiction.

It also carried the vocabulary and the emission law. So renaming a predicate moved the identity a consumer keys embeddings on, though not one chunk boundary had moved.

**The split.** A narrow preimage states only what decides where a unit starts and ends — the dialect grammar, the split law, the constants. A wide one is that narrow preimage **verbatim as a prefix**, then the vocabulary, the canon base stated present or absent, and the emission law. The prefix relation is asserted, and it is what makes the specification's MUST provable rather than merely intended: no clause that can move a boundary sits outside the wider identity, so the contract id can never stand still while a boundary moves.

Deciding which concordance rules belong in the narrow preimage was traced rather than assumed: a table row closes the open unit before the reader ever asks whether a concordance is in force, so exactly one clause there can move a boundary. It is stated explicitly, which turns "adding a concordance heading cannot re-cut the text" into a fact of the identity rather than a hope.

**A node carries the narrow id.** A node's identity answers what it *is* — this span of these bytes, cut by this law — not what is *said* about it. Had the wide id been the field, a consumer who merely declared a canon base would re-mint every unit and section of their corpus and orphan every stored reference, and two graphs over the same text under two vocabularies could never merge. The goldens show the result directly: units and sections are shared 22-of-22 between the canon and no-canon corpora, while citations share 1 of 12 — and the one still shared is the row that reifies nothing, which must not move. The fix re-mints exactly the nodes whose reified content changed and not one more.

**A citation carries the whole term it reifies** — subject, predicate and object, rendered once and used both as the node's identity and as the object of its reification, so what a node is and what it says cannot part company. The anchor alone would have left the predicate outside, and a profile renaming that predicate would still have collided; the suite constructs exactly that profile.

104 tests green, including a property over 66 profile pairs asserting that equal contract ids imply byte-identical output, with both halves proven non-vacuous. Commit `f75a517c`.

paudley:
## Gap G8 closed — and it exposed a second identity hole underneath

**The gap.** Two IRIs of the term set were written in two roles at once: one served as both the predicate carrying a section's heading and the datatype of that heading's literal, and the same for a unit's lineage. For a deployment minting its own namespace that is defensible term economy, and the specification said so honestly. It is not defensible for the namespace this specification *designates for interchange*, because no description of that namespace can declare one IRI to be both a datatype property and a datatype — a reasoner or shape validator will not accept it. Pushing the problem onto deployments did not work either: a deployment that sets those fields explicitly is no longer using the designated namespace, which was the point of designating one.

This was declined in stage 1 as deliberate term economy. That was right at the time and wrong afterwards — the directive that made the namespace canonical changed what the decision costs. Re-minting now is cheap; after publication it is not.

**The fix.** The datatypes take their own names, so classes, predicates and datatypes are three disjoint sets. The rule is stated normatively rather than described, and proven by a property over the entire term set — including over the *emitted bytes* of five profiles, with a negative control that hand-builds a dual-role vocabulary and confirms the check catches it. A future term that doubles up fails a test instead of shipping.

**The second defect, found by the audit rather than by the report.** Changing those datatype IRIs did not re-mint the contract id. The emission preimage named only the vocabulary's *base* — and a base names a namespace, not a term set — so the emitted bytes moved while the identity stood still. That is the one thing the contract id promises cannot happen, and it made a claim in the specification false for every derived vocabulary. The preimage now states the term set field by field. A renamed term re-mints the contract and leaves every unit and section exactly where it was, which is the split from the previous gap working as designed.

The audit was run over the whole set rather than the two names reported: only those two wore two roles. Near-misses were checked and cleared rather than assumed — class and predicate pairs distinguished only by case are genuinely distinct IRIs, and the non-vocabulary terms the crate emits are disjoint from the set and from each other.

107 tests green. Goldens moved by exactly 24 lines each: the renamed datatypes and the new contract id. No node IRI changed. Commit `8e63d9b0`.

paudley:
## Gaps G9, G10, G11, G15 closed — four documentation defects

**The release lane stops two crates short of where the doc said.** The bootstrap note claimed nothing in the workspace depended on the new crate, so the stop fell at the end of the set. The umbrella now takes it as a normal dependency and the wasm binding takes the umbrella, so the publish lane stops at the flagship and never attempts the binding behind it. The dependency graph, the publish order, and the script's actual stop computation were each verified before the paragraph was rewritten, rather than inferred from the gap report.

**An entire new publishable crate had no changelog entry** — nor did the new umbrella module or the widened kernel re-export. All three are now recorded under the unreleased section in its existing convention.

**Two facade tests carried each other's descriptions**, the result of inserting one test into the middle of another's doc comment in stage 1. Each now describes itself: one slices a document through the facade and asks the returned model a containment question, the other builds a real index and ranks a retrieval out of it — both written so the failure they catch is a re-export that compiles while exposing nothing usable.

**The provenance guidance named a concept with no resolvable identifier** and framed PROV-O as its rival. It now cites the concept and the document that defines it, and gives the two reasons it is recommended: the structure it carries (a transaction program with ordering, guards, branching and an explicit success quantification) and the fact that it lives in a formally provable layer, so a recorded plan is machine-checkable rather than trusted.

**One correction worth recording.** The guidance was going to say the ontology bridges to PROV with `owl:sameAs`. It does not — the mappings use `skos:closeMatch`, each row annotated with what the projection loses, including that PROV-O leaves plan structure opaque by design. Writing "sameAs" would have put a false interoperability claim into a specification offered to the community, so the section says what the bridges actually are. The substantive point is unchanged and better founded: a deployment recording the provable form gives up no PROV interoperability.

The section's clause pins were also replaced — they previously pinned the two bare words "gmeow" and "PROV-O", which pinned nothing normative; they now pin the seven actual sentences.

Commit `8bfe536c`. No goldens moved; the specification text is not part of the identity preimage.

paudley:
## Gaps G13, G14, G17, G18, G19 closed — holding true what was only true

Five places where something was correct today with nothing keeping it correct tomorrow.

**The specification vector was a presence pin wearing a behavioural name.** It asserted that thirty-odd clauses still appear in the text — genuinely useful against a clause being deleted, and blind to the failure that actually happened twice this stage: a clause describing behaviour the code does not implement. Both the trimmed-text clause and the identity clause that misstated the canon base passed it on every run. It is renamed to say what it is, and joined by clause-to-behaviour vectors that drive the public API: the dialect grammar checked rule by rule with each rule's neighbouring negative; the anchor lift pinned as concatenation *against* RFC-3986 resolution, which would give a different IRI; and an identity check that rebuilds the preimage **from the specification's own words** and mints every node the graph must already carry — including proving field three is the chunking id by showing the same formula under the other id names a node the graph does not have.

**A second path to section identity agreed with the projection by coincidence.** The unit equivalent was pinned; this one was not. It is now, end to end over every section of the corpus, plus the sharper fact that the identity is minted over the heading *line* rather than the section — minting over the whole section names a node that does not exist.

**A forbidden-character set closed only by accident.** The crate's own character check stops at the ASCII range while the kernel's writer also escapes the C1 block, and the gap between them is covered only because the IRI grammar opens its permitted range one code point above it. Nothing said so. Now the reason is documented where a future reader would otherwise "simplify" the two sets into disagreement, and proven at all four seams: C1 characters refused carrying the law's own finding, the very next code point admitted and minting.

**The predicate was written by a different path than its subject** — consistent with the kernel, but the crate's own specification warns against exactly that. Unified rather than documented, because the asymmetry's only defence was that the kernel does it too, and the kernel does it *inside* the canonical writer, which is the thing implementations are told to reach through rather than to reimplement beside.

**The heading-stack law lives in two files.** De-duplication would have damaged the seam: the reader needs its stack for facts that are only knowable while lines go past, and having it report a parent would push a containment fact back into the reader, which the seam contract forbids. So the law is stated once and the two implementations are proven to agree over every one-, two- and three-section sequence across six levels and the movement marker, plus longer generated ones — 463 documents.

117 tests green; no golden moved, because none of this changes behaviour. Commit `a1250c14`.

paudley:
## An added gap, found while writing a control-character vector — and fixed here by maintainer direction

**The defect.** The SPARQL lexer ended an IRI reference at any Unicode whitespace. The production it implements excludes only the range through the ASCII space and nine reserved delimiters — every one of them ASCII. A non-breaking space is lawful inside an IRI reference, and so are the other non-ASCII spaces.

The consequence was a broken round trip in the workspace's own hands: the serializer wrote such an IRI correctly, because *its* escape table already matched the production, and the reader then refused what the writer had just produced. Serialize-then-parse was not an identity for any IRI carrying non-ASCII whitespace.

**Why it hid.** The one vendored W3C test that carries a non-breaking space in an IRI ships it as a ` ` escape, which takes the decoding path rather than the verbatim one. The conformance suite could not see this.

**The fix reuses a spelling the workspace already had.** Two sibling lexers — ShEx and CDT — already implement the production correctly, character for character. The SPARQL lexer was the lone third spelling, and it now uses the same predicate rather than a fourth. The fast path became a byte scan, which is exact here because every excluded code point is ASCII, and drops a per-character decode from the hot path.

**Restraint where restraint was right.** Two nearby uses of the same wrong predicate govern inter-token whitespace and the anonymous-node production. They are over-lenient against the grammar — the opposite defect class — and tightening them would *add* refusals. They are reported and left alone rather than changed on the way past.

**Proven, not asserted.** The vectors were checked by restoring the old predicate and confirming four of them fail. The round trip is proven on both the verbatim and escape-decoding paths, and the ASCII space is still refused, which is the neighbouring case that keeps the fix from becoming over-refusal.

`make conformance` is GREEN — 4488 passing, and all twenty-one scoreboard rows compared cell by cell against the committed scoreboard: nothing moved. That flatness is the expected result, not a disappointment: the defect only fires for IRIs no vendored suite ships raw.

Commit `9677175`. Nothing under the markdown crate was touched, so this change is separable from the rest of the branch.

paudley:
## Stage-2 exit gate: the completion adversary returned FINDINGS-OPEN

It rebuilt every demonstration from an external consumer crate driving only the public API and the umbrella facade — not the test harness — and **both prior findings are discharged**:

- The resolver-singleton gate passes, and it is not hollowed: a hand-rolled resolver that merely *names* the shared type was planted and still fired. All four newly-wired gates run green, unconditionally, inside the CI workspace job.
- The three-profile identity repro now yields three contract ids and one chunking id, with the narrow preimage a verbatim prefix of the wide one and every unit and section IRI unchanged across the three. Citation IRIs no longer collide across profiles.

Fourteen acceptance criteria were demonstrated in production, including the bypass that started this: admitting a document with no canon base and then projecting it under one no longer leaks the hostile anchor, and the model's fields are private with no setter, so the bypass is closed by construction rather than by convention.

**Three findings remain, and the first is mine.**

### The specification contradicts itself on citation cardinality

Section 5 — which the document itself designates as the law — says a citation node "is a reifier: it identifies **one** occurrence of **one** triple." Section 11.1 mandates one node per row-lifting, reifying *each* anchor that lifting produced. A two-anchor row emits one node carrying two reified edges, so the code implements one section and violates the other.

That paragraph was **introduced by the identity split in this stage** — my change, arguing against the shape the crate ships. It matters beyond prose: a second implementer reading section 5 mints one node per anchor, reading section 11.1 mints one per lifting, and the two disagree on every citation IRI. That is precisely the interoperability failure content-addressed identity exists to prevent, and the failure that paragraph invoked to justify itself.

Being fixed by rewording section 5 to the law actually in force, not by changing emission: one node per reified triple would mint **zero** nodes for a row naming sources and no anchors, re-opening the orphan-citation defect closed earlier this stage. The argument the paragraph carried — why the reified terms belong inside the preimage — is preserved and restated correctly, and the cardinality law becomes an executable vector rather than a sentence.

### Two tautological identity tests

Two vectors assert that a convenience constructor equals the general one, but the convenience constructor's entire body *is* that call — after inlining, the assertions read `f(x) == f(x)`. They cannot fail except by deleting the delegation they exist to check, and a doc comment beside them claims the two routes are "held to one answer by the vectors rather than by resemblance," which at that seam is not true. Being rewritten against a hand-built preimage, with the doc comment corrected to claim only what is proven. The independent proof does exist elsewhere — a vector that rebuilds the preimage from the specification's own words — and the comment will name it.

### The lexer disclosure had no owner

A PR comment honestly disclosed two over-lenient whitespace sites left unchanged. The disclosure was right; leaving it with no owner was not. Filed as its own issue with the demonstration, because tightening it *adds* refusals and belongs with a conformance re-run rather than riding along here.

### Also corrected

The PR body's test count was stale from stage 1 — it says 76; the real count at head is 117.

paudley:
## The toolchain pin argued for itself on two grounds, and both inverted under reading

This began as an environment puzzle and ended as a defect. Recording it because the reasoning is the useful part.

**Ground one: "nightly buys a sharper analysis surface, so a finding here is a real finding."** The stated purpose is to catch more. Freezing the date does the opposite — it welds that surface to one day, and every lint sharpened afterwards stops being a finding and becomes debt accruing behind a green gate. Nine redundant clones were sitting underneath it: two in library code, seven in the conformance corpora's own walks, each one a value the program copied and then dropped unused. A recent default borrow checker is also markedly stronger, which a date forfeits as well. The file argued for rigor while implementing preservation.

**Ground two: "a floating channel is incompatible with byte-deterministic contracts."** This inverts the argument it is making. Determinism here is a property of the *code*: output sorted, deduplicated and explicitly ordered, identities content-addressed over this workspace's own declared law and never over anything a compiler chooses. Sorted output is sorted whoever compiled it. Determinism that survives only under one compiler build is not determinism — and a golden that moved under a bump would be a serializer defect the pin was concealing, not a reason to stop bumping.

The gate that enforced the date now **rejects** it and requires the floating channel, while keeping every check it already made: each workflow must still name the same channel, a toolchain action or environment override that disagrees is still drift, and the MSRV job keeps its deliberate stable pin exactly as before. Its self-test proves the new rule in both directions — a dated or exact channel is refused, a floating one with agreeing workflows passes.

Ten workflow sites moved, and the rationale was rewritten wherever it was repeated in prose: the contributor guide, both front pages, the agent contract, the design rules, the release notes, and the shapes crate's README. The zh-Hans passage went through blind back-translation; every claim survived the round trip.

**Workspace clippy is clean on the floating channel** — which is the evidence that the nine defects were the whole of what the date was hiding, rather than the first of many.

Commits `43993078` (the nine clones) and `6e77d7a1` (the float and its rationale).

paudley:
feat(markdown): a document becomes a graph of itself

Adds `purrdf-markdown`, the twenty-second publishable crate: a structural
Markdown-to-RDF 1.2 slicer that turns a written document into claims along
the structure its author already put there. A document node; a section per
ATX heading and per movement marker; a unit per numbered verse and per
blank-line paragraph, each carrying exactly one plain literal holding the
verbatim bytes of its span, beside its ordinal, byte span, scalar span,
content digest, section and heading lineage. A concordance table lifts into
RDF 1.2 reified citations. Output is a total, deterministic function of the
source bytes and a declared profile.

The crate ships a specification, `crates/markdown/SPEC.md`, offered for use
rather than merely documenting an implementation, with a designated
vocabulary namespace `https://w3id.org/purrdf/markdown#` and a related-work
section naming the twenty-one standards the absence claim was measured
against.

## The model is the product

`analyze` returns a typed stand-off model — untouched source bytes plus a
layer of typed range annotations — and `render` is one projection of it into
claims. The model carries what a flat triple projection cannot: concordance
rows unflattened, malformed- and unmatched-row accounting as data rather than
silence, containment-lattice queries, the thirteen Allen interval relations,
and content anchors for re-anchoring across revisions. The Markdown dialect
reader sits behind a documented seam; splitting, identity, claim rendering and
the lattice are dialect-independent, which is where a second format attaches.

## Identity: two preimages, because a chunk is not a claim

A narrow preimage states only what decides where a unit starts and ends — the
dialect grammar, the split law, the constants — and feeds the embedding stage
identity. A wide preimage is the narrow one verbatim as a prefix, plus the
vocabulary term set, the canon base stated present or absent, and the emission
law; it feeds the contract id written into the graph. The prefix relation is
asserted, which is what makes the specification's MUST provable: no clause
that can move a boundary sits outside the wider identity.

A node's identity carries the narrow id, because a node answers what it *is*,
not what is said about it. Declaring a canon base or renaming a term therefore
re-mints the graph's label and leaves every unit and section where it was —
proven by the goldens, which share 22 of 22 unit and section identities across
the canon and no-canon corpora.

## Requirements satisfied

Every criterion of the issue, as modified by six recorded human directives:
the crate and its node model; one plain literal per unit over its verbatim
span; ordinal, byte span, section, lineage; the concordance lift; the split
law with its snapped overlap, never inside a scalar and never across a
heading; determinism; profile parameters inside the identity via
`derive_chunking_contract_id`; caller-supplied vocabulary via
`Vocabulary::under`; a sole runtime dependency on `purrdf-core` with
`purrdf-rdf` dev-only; no I/O; no Cargo features; rlib only; wasm-clean;
registration as a release crate and in the bootstrap ledger; the publishable
count moving to twenty-two.

Six directives modified the issue's text, each recorded verbatim in the PR
rather than applied silently: a private corpus was replaced by a neutral
synthetic one; the contract-id reproduction criterion was dropped as
greenfield, replaced by stability vectors; the mints-no-vocabulary law was
scoped to the processing core so this import surface mints a designated
namespace; the umbrella re-exports the crate; a SPARQL lexer fix was folded
in; and the workspace moved off a dated nightly pin.

## Standing constraints

RDF 1.2 only — citations are reified through minted-IRI reifiers carrying
parenthesized triple terms, emitted by the kernel's canonical writer.
Hard fails, no optionality — every refusal is a typed error at the entry
seam, and every refusal ships with an executed valid-neighbour case proving
it did not widen into over-refusal. Rust-first, wasm-clean, no features, no
member-crate version pins.

## Defects found and fixed beyond the issue

The concordance lift destroyed row pairing: anchors and sources from
different rows merged, and deduplication collapsed multiplicity, so a verse
covered by two rows could not say which source belonged to which anchor.

A concordance row could hang the slicer: lifts were decided by walking the
row's declared verse range, which the document controls, so a row naming the
full `u64` range did not terminate. Lifts now walk the document's own verses.

The projection could be handed a law the document never passed: analyzing
with no canon base and then rendering under one minted the traversal anchor
the containment law exists to refuse, into a reified triple term. The model
now carries the law it was admitted under and the projection takes one
argument.

A row inside a subsection of the concordance was silently lost; an escaped
pipe silently emptied a row's cells; a row naming sources but no anchors
minted a citation node nothing pointed at. Each is now stated law with a
vector, and citation-node reachability is proven by walking the emitted graph
rather than by counting triples.

Nine redundant clones — two in library code, seven in the conformance corpora
— stood because a dated toolchain pin froze the lint surface. A SPARQL lexer
ended an IRI reference at any Unicode whitespace, so a lawful IRI carrying a
non-breaking space serialized correctly and refused to re-parse; it now uses
the predicate two sibling lexers already spelled.

Four hygiene gates that `make check` ran and no workflow did are now wired
into CI, including one that nothing ran at all.

## Completion-adversary: PASS

Five rounds. Every acceptance criterion was demonstrated on the production
surface from external consumer crates driving the public API and the
`purrdf::markdown` facade — not the test harness — including rebuilding both
committed goldens byte-identically from outside the repository.

Rounds one and two closed code findings: a red gate invisible to CI, an
identity that did not answer for the bytes it emitted, and the render bypass.
Rounds three through five closed findings in the *claims about* the work: a
designated namespace that does not resolve with the disclosure omitted, a
false changelog bound, a stale test count in the artifact that becomes this
commit message, a retracted claim surviving in that same artifact, a summary
contradicting its own table, and nine ranking claims in the related-work
section that its own entries falsified. The final round hunted specifically
for a relocated instance and found none.

## Deferral gate: clean

Zero keyword hits on added lines. Two PR-text hits adjudicated: CodeRabbit's
out-of-scope Makefile warning is a proven false positive — the branch's entire
Makefile diff is four hunks and touches neither native-import targets nor the
shared-shapes test, which appear only as unchanged context — and the w3id
redirect is an act in a repository this project does not control, now filed as
`perma-id/w3id.org` PR 6683 with the target serving live.

`.deficiencies` holds its notice and marker, with nothing below the line.

## Verification

`make check` green end to end at the merge commit, its terminal wasm step
included; `make conformance` GREEN at 4488 with every scoreboard row unmoved;
`cargo test -p purrdf-markdown` 119 green; workspace clippy clean under
`-D warnings`; rustdoc clean; goldens byte-stable since the law settled.

## For whoever comes next

The goldens are law guards. A change that moves them changed the law; the
correct response is to stop, not to regenerate.

`Profile::emission_bytes()` and `chunking_bytes()` are public API in the
strongest sense: their exact bytes are the preimage of every identity in the
world. Editing that prose re-mints. That is the design working, not a hazard,
but the strings are not comments.

The designated namespace is a stable identifier that does not dereference
today. The redirect is filed; nothing in the law depends on it.

The zh-Hans translations were produced under blind back-translation and the
project's glossary gate, but have not had a native-reader review.

PR #290, opened 2026-09-11
fix(scanners): a scanner's character classes decide token boundaries, so make them exact

### LINKED ISSUE #287: fix(sparql): non-ASCII whitespace is absorbed into names and silently changes query results
## Summary

The text stack's shared lexer tests `char::is_whitespace` — the full Unicode `White_Space` property — where the grammar's `WS ::= #x20 | #x9 | #xD | #xA` admits four code points. Two named sites in `crates/sparql-algebra/src/lexer.rs`: the inter-token skip, and the `ANON ::= '[' WS* ']'` scan.

That much is over-acceptance. **The same root cause also produces silent wrong answers**, which is the more serious half and the reason this is not merely a conformance nicety.

## The silent misparse

`is_pn_chars_base` and `is_varname_char` approximate their productions as "any character above U+007F". Non-ASCII whitespace is above U+007F, so the greedy name scan absorbs it *before* the whitespace skip is ever consulted:

```
?s<NBSP>?p        →  [Variable("s\u{a0}"), Variable("p")]     not two variables named s and p
ex:a<NBSP>ex:b    →  [PrefixedName("ex", "a\u{a0}ex:b")]      one term, not two
```

End to end through the query parser:

```
SELECT ?s WHERE { ?s<NBSP><urn:ex:p> ?o . ?s <urn:ex:q> ?z }
  →  ACCEPTED, four distinct variables: "s\u{a0}", "o", "s", "z"
```

The author wrote three. One invisible character turns a join into a cross product and projects a variable the first pattern never binds. Exit status zero, no diagnostic, wrong rows. `VARNAME` and `PN_CHARS_BASE` both exclude U+00A0, so this is not a grey area.

Tightening the whitespace skip alone does not fix this, because the skip never sees that character.

## Scope

This lexer is the shared front end for SPARQL queries and updates, Turtle, TriG, N-Triples and N-Quads, and therefore for the CLI, the Python, WebAssembly and C surfaces. The same defect exists in the N-Triples line path (`raw.trim()` is the Unicode class) and in the ShEx lexer and ShapeMap parser, both of which inherit Turtle's `WS`.

## Acceptance criteria

- [ ] The four `WS` code points, spelled explicitly. `is_ascii_whitespace` is not the fix — it admits U+000C, which the production does not.
- [ ] The name-character predicates restrict to their real productions, so an unlawful character reaches the lexer's error arm at the right offset instead of being absorbed into a name.
- [ ] Both lexer sites land together: changing the `ANON` scan alone flips `{ [<NBSP>] }` from refuse to accept.
- [ ] The N-Triples line path trims the four, not the Unicode class, with column reporting moved in step.
- [ ] The ShEx lexer and ShapeMap parser corrected the same way.
- [ ] Every newly refused code point has an executed vector, each paired with the ASCII neighbour that must still parse.
- [ ] U+1680 OGHAM SPACE MARK still parses as a name character — it is simultaneously Unicode whitespace and inside `PN_CHARS_BASE`, and is the one genuine over-refusal risk.
- [ ] A lawful IRI containing U+00A0 still parses. That is a recently fixed defect in the same file, in the opposite direction, and must not regress.
- [ ] `make conformance` re-run; any moved row explained; no frozen corpus or baseline budget edited to accommodate the change.

paudley:
Closes #287

## The defect

`crates/sparql-algebra/src/lexer.rs` is the shared front end for **SPARQL queries and updates, Turtle, TriG, N-Triples and N-Quads**, and therefore for the CLI, Python, WebAssembly and C surfaces. It approximated its grammar's character classes with convenient predicates: `char::is_whitespace` where `WS ::= #x20 | #x9 | #xD | #xA`, and "anything above U+007F" where `PN_CHARS_BASE` and `VARNAME` have exact range tables.

Non-ASCII whitespace is above U+007F, so the greedy name scan swallowed it before the whitespace skip was ever consulted:

```sparql
SELECT ?s WHERE { ?s<NBSP><urn:ex:p> ?o . ?s <urn:ex:q> ?z }
```

was **accepted, with four distinct variables** where the author wrote three. A join silently became a cross product and projected a variable the first pattern never binds. Exit zero, no diagnostic, wrong rows. The Turtle-shaped mirror, `ex:a<NBSP>ex:b`, minted **one** IRI instead of two and wrote it into a data file, where it outlives any query answer.

I filed the issue describing over-acceptance — a conformance nicety. That was wrong, and the issue was rewritten: this is a silent wrong-answer bug.

## Why it is a class and not an incident

These predicates run inside a **maximal-munch** scan, so a superset character class does not merely widen the accepted language — it changes the **tokenization of documents both parsers accept**. `?s` NBSP `<urn:ex:p>` is lawful input to both a conforming parser and this one, and they disagree about what it means.

The class also had a *written authorization* in the workspace. `crates/rdf-core/src/blank_label.rs` stated that a lexer may safely over-accept because it is "sound on ingress — a superset lexer can only accept documents a conforming parser would also accept, never reject one". That lemma is false for any predicate deciding a token boundary, which is every predicate in a scanner. Fixing the instances while leaving that paragraph in place would have authorized the next one. It is replaced with the corrected law, and `AGENTS.md` carries it as a hard constraint.

## What landed

| | |
|---|---|
| `crates/iri/src/terminals.rs` | every W3C terminal, spelled **once**, production cited, ranges asserted at compile time. `purrdf-iri` stays a zero-dependency wasm-clean leaf. |
| `sparql-algebra` | `WS` at both sites; `PN_CHARS_BASE`/`PN_CHARS`/`VARNAME` by production; invisible-character diagnostics |
| `shex` + ShapeMap | seven sites, each cited from **ShapeMap's own** grammar |
| `rdf` | the three line paths, and the BOM decision |
| `cdt`, `geo` (WKT + GeoJSON), `sparql-results` | three more scanners found by the new gate |
| `scripts/check-terminal-predicates.py` | the gate, in `make check` |
| head classes | `PN_PREFIX`, `PN_LOCAL` and `BLANK_NODE_LABEL` each restrict their FIRST scalar, and no two of the three agree. `ex:-a`, `_:.a` and `_:\u{300}a` lexed as single names in both entry points, so our parsers accepted labels our own egress validator rewrites |
| `PN_LOCAL_ESC`, `IRIREF` content, `PERCENT` | the last forks collapsed. `PERCENT ::= '%' HEX HEX` was checked in `shex` and not in `sparql-algebra`, so `ex:%zz` was one name in one crate and two in the other |

**The workspace held four independent transcriptions of `PN_CHARS`** — `iri`, `rdf-core`, `shex` and `cdt`. Three are deleted. `rdf-core`'s is deliberately kept as an independent *egress* oracle and is the gate's one reasoned exemption, valid only while an all-scalar equivalence test proves the two agree.

## Over-refusal is the whole risk direction

This change adds refusals, so the mirror bug — rejecting input that is actually valid — is the live danger, not a theoretical one. Three characters look like garbage and are **named by the production**; each is pinned positively so a later "cleanup" cannot manufacture an over-refusal:

- **U+1680 OGHAM SPACE MARK** is simultaneously Unicode `White_Space` *and* inside `PN_CHARS_BASE`, so `?s<U+1680>?p` is lawfully **one** variable — visually identical to the headline bug and exactly its opposite.
- **U+200C / U+200D** are named by `PN_CHARS_BASE`, so a diagnostic table built by sweeping "invisible format characters" would tell authors their valid queries are wrong.
- **U+FEFF** is inside `[#xFDF0-#xFFFD]`, so a byte-order mark is a lawful **name** character. Two agents' first drafts got this wrong.

Being an offender is a property of the **position**, never of the character.

The claim the whole guard rests on is now **derived, not asserted**: intersecting `char::is_whitespace` with each of the five name classes over all 1,114,112 scalars yields exactly `{U+1680}`, and no `WS` member is a name character in any position.

## Evidence

**What the boundary sweep proves — and what it does not.** Four scan entry points are swept (SPARQL `tokenize`, `tokenize_turtle`, the ShExC lexer, ShapeMap's) at `(lo−1, lo, hi, hi+1)` for every range of every production, plus the whole `White_Space` property, the invisible format characters and the `IRIREF` delimiters — 3,630 probes, **+0.6 s** on `make check`, against 35.7M in an `#[ignore]`d lane that also passes.

An earlier draft of this section claimed that made it complete against a wrong table. **It does not, and that was measured rather than argued.** For the six classes backed by the shared module the expectation *delegates to the predicate under test*, so the sweep is a scanner↔predicate **composition** test: widening `PN_CHARS_BASE` by one scalar leaves all four sweeps green across 35.6M probes. Only the derived-ranges snapshot fails.

It is not composition-only throughout. `IRIREF content` and `STRING_LITERAL2 content` are each spelled **twice** — once by the scanner, once independently by the harness — so for those two the sweep genuinely *is* a table oracle.

**The tables are guarded, in three places that all caught the injected mutations:** an independent `matches!` transcription in `terminals.rs` checked against the range tables over every scalar; the `terminal!` macro's `cardinality` / `ranges_all_ascii` / `ranges_sorted_disjoint` const assertions (`WS += U+00A0` is literally uncompilable); and the tracked `insta` snapshots. ShEx now snapshots the one production it transcribes itself, which was rendered nowhere before.

Refusal sets are **generated**, never hand-picked: an early draft omitted U+205F and U+2029, which is the argument in one line.

**The sweep found three defects fifteen hand-written vectors missed**, all in ShapeMap: `BLANK_NODE_LABEL` had no first-scalar class; `_:a.` minted the label `"a."`; and `char::is_control` stood in for two content classes, so `<urn:ex:a b>` **named an IRI containing a space**.

**Census** (`--census`, glob recorded in the script): 4,327 files, **zero** bare tokenization-affecting occurrences; the single hit is `<urn:ex:\xa0>` in the frozen W3C RDFC-1.0 fixtures — a *lawful* IRI a careless tightening would have broken. `--census-refusals` measures the populations the other refusals moved: **0** comment-emptied bracket pairs, **0** leading BOMs, **1** head-class hit — a negative-syntax vector whose purpose is to be refused.

A corpus with no offenders is evidence of no observed regression; it is never evidence of no over-refusal, which by definition lives in input the corpus does not contain.

## Verification

| Gate | Result |
|---|---|
| Hygiene scripts | **17/17** exit 0 |
| Workspace clippy `-D warnings` | exit 0 |
| Workspace `cargo check --lib --tests` | exit 0 |
| Tests | **8,342 passed, 0 failed** — every workspace member |
| `rdf-core-hygiene`, `wasm`, `doc` | all exit 0 |
| `make conformance` | **VERDICT: GREEN** — 21 rows, 0 RED, 0 FAIL |

**No conformance row moved, and no corpus, fixture, baseline budget or xfail ledger was touched.** Syntax codecs 264/0 · SPARQL 1.1/1.2 evaluation 862 pass / 5 ledgered · SPARQL CDT 658/0 · SHACL Core + SPARQL 129/0 · ShEx validation 1105/0 · rdflib drop-in 81 pass / 5 ledgered · Python bindings 796 pass / 4 ledgered.

### Bench — reported, not asserted

`turtle_4k` 1.20 ms / 410 MiB/s · `sparql_4k` 1.80 ms / 274 MiB/s · `modifier_query` 439 µs / 225 MiB/s. Recorded because repo law wants a bench behind a perf-touching change, **not** offered as evidence of a speedup: this machine is not quiet, and the change pushes both ways — the inter-token skip loses a UTF-8 decode and a property lookup, the name scan gains a range tree. Predicates are ordered ASCII-first so the common path stays one compare.

## What review found after the first round

Three audits — a gap analysis, a completion adversary, and CodeRabbit — accepted the named scope as delivered and attacked what *"everything"* meant. Thirteen further gaps were closed. The ones worth a reviewer's attention:

- **A second silent data-drop in the same file**, worse than the first: the line parser called `expect_dot` and returned **without checking the cursor was exhausted**, so every token after the terminator was discarded. And a *missing* terminator parsed clean. Both are now hard errors.
- **The RDF/XML writer emitted predicates it could not read back.** Missing the name class was not a degradation: the fallback declared the whole IRI as a namespace and emitted `nsN:property`, which a reader concatenates to `<whole_iri>property` — a **different predicate**, at exit zero. `http://example.org/123` round-tripped as `…/123property`. The fallback is deleted; the split now walks back over `NCNameChar` to the longest lawful suffix, so a stricter class routes **zero** IRIs into a lossy path. A third defect in the same seam: a predicate inside a quoted triple was emitted bound to an **undeclared prefix**.
- **The gate could be silently blinded by an ordinary `use` line.** A bodyless `#[cfg(test)]` item made the test-module blanker consume the next item's brace and erase the scanner. Live in five modules.
- **Its blind spot was larger than disclosed, and my stated reason was false.** I wrote that widening the cursor pattern would "sweep in every config and rendering path". Measured: one further file, zero new offenders — and that file was a real scanner the whole time.
- **`AGENTS.md` and the gate both asserted a false CommonMark claim** that licensed `char::is_whitespace` in the Markdown dialect. CommonMark's blank line is "only spaces (U+0020) or tabs (U+0009)"; its Unicode whitespace class governs emphasis flanking alone. The unit of judgement is the **clause**, not the specification.
- **Five escape transcriptions, not three** — collapsed, with no emitted byte changed over all 1,114,112 scalars.
- The already-fixed defect standing in sibling codecs (`hextuples`, `trix`), XSD lexical spaces trimmed with the Unicode class in a SHACL validator and the entailment parser, `xsd:token` wrong in *both* directions at one site, `xsd:NMTOKEN` accepting `!!!`, and the XPath `x` regex flag mapped onto Rust's `ignore_whitespace` — which removes 26 code points instead of 4, treats `#` as a comment, and strips inside character classes.

## Also carried — withdrawn

This section previously claimed a one-line `crates/rdf-core/src/ir/pack/bits.rs` fix: `assert_eq!(decoded, [] as [_; 0])`, which a newer nightly's tightened inference stopped accepting, leaving `origin/main` unable to `cargo check --tests` with no commit having touched it.

That was true when written and is no longer. `main` fixed it independently while this branch was in review, so the file is **not** in this diff — `git diff origin/main...HEAD -- crates/rdf-core/src/ir/pack/bits.rs` is empty and both sides now read `[u64; 0]`. The claim is withdrawn rather than quietly deleted, because this body becomes the squash commit message and a reader comparing it against the diff would find a file that is not there.

## Reviewer notes

**What I am least confident about.** The refusals whose justification is an *absence* argument. The BOM refusal is safe because N-Triples has no bare-word position anywhere, so the alternatives are exhaustible — that reasoning is sound but it is the kind that stops being sound when a grammar gains a production. Second: the gate's scanner detection is per file, and it **missed** `geo/src/geojson.rs`, whose cursor lives in `json.rs`. That blind spot is documented rather than papered over, because widening the net would sweep in every config path those crates own, and a gate that cries wolf teaches authors to route around it.

**What you may not know.** `sh:pattern` is compiled with the Rust `regex` crate, whose `\s` is Unicode `White_Space`, not XSD's four — `crates/shapes/src/constraints.rs` already declares this a "Deviation from XPath 2.0". It means every `sh:pattern` in the repo evaluates more leniently than it is specified. It is the same *shape* of defect as this PR but a different layer — a regex dialect, not a scanner terminal — with its own conformance surface. Not touched here; flagged deliberately.

`cdt`'s `parse_blank_node_label` shares the head-class defect and is covered by the in-flight fix.


paudley:
## Implementation plan as approved

_Local worktree mechanics removed; this is the technical record._

# Implementation Plan for Issue #287 — the scanner's character classes must be exact

## Context

`crates/sparql-algebra/src/lexer.rs` is the shared front end for **SPARQL
queries and updates, Turtle, TriG, N-Triples and N-Quads**, and therefore for
the CLI, Python, WebAssembly and C surfaces. It approximates its grammar's
character classes with convenient predicates: `char::is_whitespace` where
`WS ::= #x20 | #x9 | #xD | #xA`, and "any character above U+007F" where
`PN_CHARS_BASE` and `VARNAME` have exact range tables.

I filed this issue describing over-acceptance — a conformance nicety. **That was
wrong.** The same root cause produces **silent wrong answers**:

```
SELECT ?s WHERE { ?s<NBSP><urn:ex:p> ?o . ?s <urn:ex:q> ?z }
  →  ACCEPTED, four distinct variables: "s\u{a0}", "o", "s", "z"
```

The author wrote three. One invisible character turns a join into a cross
product and projects a variable the first pattern never binds — exit zero, no
diagnostic, wrong rows.

**The mechanism, which is the reason this is a class and not an incident:**
these predicates run inside a **maximal-munch** scan, so a superset character
class does not merely widen the accepted language — it changes the *tokenization
of documents both parsers accept*. `?s` NBSP `<urn:ex:p>` is lawful input to
both, and they disagree about what it means.

**The class has a written authorization in this workspace.**
`crates/rdf-core/src/blank_label.rs:13-26` states the rule that produced six of
these in two weeks: the lexer *"deliberately over-accepts on parse … sound on
ingress — a superset lexer can only accept documents a conforming parser would
also accept, **never reject one**"*. That lemma is false for any predicate that
decides a token boundary, which is every predicate in a scanner. Fixing six
instances while leaving that paragraph in place authorizes the seventh.

**Release constraint (user): before 2.0.** The fix adds refusals, so input that
parses today begins to fail. That is a breaking change and belongs in the major
bump; afterwards it is a semver violation or a deprecation cycle for behaviour
that was never lawful.

**Scope (user): "Everything including ShEx."**

## Standing constraints

`.goals`: greenfield, no compromises; **LOW/NO OPTIONALITY, HARD FAILS**; RDF
1.2 only; maximal utility, performance, portability (wasm-clean).
`.baseline`: adversarial; no deferrals; no issue/PR numbers or process
references in any repo file; see a bug, own it.
Repo law: no Cargo features; no member-crate version pins; every public item
documented; `-D warnings`; **every refusal ships with an executed valid
neighbour**; a perf-touching change gets a bench (report-only, never asserted);
`cargo fmt --all` before each commit.

## Completeness Contract

| Requirement | Task |
|---|---|
| The four `WS` code points, spelled exactly; `is_ascii_whitespace` refused as the fix because it admits U+000C | 2, 3 |
| Both lexer whitespace sites land in one commit — changing `ANON` alone flips `{ [<NBSP>] }` from refuse to accept | 3 |
| Name predicates restricted to their real productions, closing the silent misparse | 4 |
| **`PN_CHARS`'s extra ranges preserved** — `#xB7`, `[#x300-#x36F]`, `[#x203F-#x2040]` — so `ex:col·lecció` and every NFD-decomposed name still parse | 4 |
| **`VARNAME` split into start and continue predicates**, and `'-'` excluded from it though `PN_CHARS` includes it, so `?a-?b` still parses | 4 |
| Blank-node labels covered — `lex_blank_label` calls the same predicate, and label identity is data identity | 4 |
| The false ingress-liberality doctrine deleted from the kernel and replaced with the corrected law | 1 |
| One home for the terminal classes, with its owning crate and export named; no third transcription of tables that already exist twice | 1 |
| The empty bracket pair refused in **both parsers**, which is where the `ANON` over-acceptance actually lives | 3 |
| `shapemap.rs`'s `take_keyword` boundary predicate, same class, found while auditing | 8 |
| Every refusal list generated from the exhaustive sweep rather than hand-picked | 5 |
| An end-to-end Turtle vector through `parse_dataset` — a silently wrong IRI in a data file outlives a wrong query | 10 |
| Exhaustive verification over every Unicode scalar, per production | 5 |
| Diagnostics naming the offending character and its production | 6 |
| N-Triples line paths: the four, not the Unicode class — including whitespace-only and comment lines, not just separators | 7 |
| ShEx lexer and ShapeMap corrected | 8 |
| A lawful IRI containing U+00A0 still parses — the day-old `is_iriref_forbidden_byte` fix does not regress: `crates/rdf/tests/relative_iri_base.rs:493-542` and `lexer.rs:937-990` re-run green after tightening, plus a new end-to-end `<urn:ex:a\u{a0}b>` vector through the public parser | 4, 5, 10 |
| A gate that refuses the seventh instance | 9 |
| The kernel's unchecked sorted-and-disjoint precondition promoted to a compile-time assertion | 1 |
| Each production expressed once as data, emitting predicate, assertions and sweep registration | 1 |
| The hot-path throughput change reported through the existing benches, never asserted | 2, 10 |
| `make conformance` re-run, rows named, nothing edited to accommodate | 10 |

## Tasks


### Task 1 — one home for the terminals, and delete the doctrine that authorized the bug
The exact tables **already exist twice** and neither is cited by the issue:
`crates/rdf-core/src/blank_label.rs:521-582` (range tables with per-range W3C
citations, binary search, adversarial tests) and `crates/shex/src/lexer.rs:882-911`
(the same production as a `const fn`, the cheaper form for a hot path). Writing a
third is the defect repeating itself.

`purrdf-sparql-algebra` deliberately does not depend on `purrdf-core`, but
sparql-algebra, shex and rdf-core **all depend on `purrdf-iri`**, a required
zero-dependency wasm-clean leaf. Add `pub mod terminals` there in the `const fn`
form (adds no dependency, keeps the leaf zero-dep); `purrdf-core` re-exports to
preserve its surface. Lift ShEx's already-correct predicate rather than
retyping it. Keep `blank_label.rs`'s independent range table as the **test
oracle** — implementation versus independent transcription is stronger than
collapsing to one spelling.

Then rewrite `blank_label.rs:13-26`: a scanner's character classes must be exact
in both directions because they decide boundaries, not merely membership; only
*validators* may be liberal on ingress. Cite the misparse as the counterexample.

**And fix the latent bug the same file is carrying.** `blank_label.rs:524-525`
states, in a comment, that its ranges are "sorted and non-overlapping, which
`in_ranges` relies on for binary search" — an unchecked precondition guarding a
binary search that decides the **egress alphabet for every blank node this
workspace writes**. Insert a range out of order and membership is silently
wrong, with no error and no test failure, in the file whose contract promises
two distinct blank nodes never become one. This defect class, one crate down.
Promote it to `const _: () = assert!(ranges_sorted_disjoint(...))` — a `const fn`
`while` loop, stable since 1.46/1.57 — so the comment becomes a compile error to
violate.

**Express each production once, as data, via one `macro_rules!` per terminal**
emitting the `const fn` predicate, its const assertions (sortedness, and the
cardinality of a small set like `WS`), and its registration into Task 5's sweep.
Numeric literals stay in plain source, which matters because Task 9's gate scans
source text: a proc-macro-expanded predicate would be invisible to the mechanism
meant to keep the next author honest.

Keep the `matches!` form on the hot path and the range table as data for the
assertions and as the sweep's oracle. Do **not** put `in_ranges`'s binary search
(~4 unpredictable branches over 13 ranges) into the tokenizer's inner loop; it is
right for egress, once per label, and wrong per character.

### Task 2 — the `WS` predicate, and the byte idiom
Introduce `WS` in `terminals` with its production cited, styled after
`is_iriref_forbidden_byte` (`lexer.rs:873`). **`is_ascii_whitespace` is not the
fix** — it admits U+000C, which `WS` does not. Spell the four, as
`crates/cdt/src/parse.rs:274` and `crates/geo/src/wkt.rs:318` already do — and
copy their *shape*: a byte compare on `peek() -> Option<u8>`, exact over UTF-8
because every lead and continuation byte is ≥ 0x80. Today `skip_trivia` decodes
a `char` per iteration and `lex_bracket_or_anon` does a slice-plus-decode per
character, so the correct fix is **less work per character in the hottest loop**.

This is the inter-token skip: it runs between every pair of tokens, in SPARQL,
Turtle, TriG, N-Triples and N-Quads, on the CLI, Python, wasm and C surfaces.
The fix removes a UTF-8 decode and a Unicode-property lookup from it. Say so in
the PR and report it through the existing `benches/tokenize.rs` and
`benches/tokenize_values.rs` — report only, assert nothing, per repo law and a
machine that is not quiet. The name scan moves the other way (one compare → a
range tree), so order the predicate ASCII-first: the overwhelmingly common path
stays a single compare and only non-ASCII pays the tree.

### Task 3 — both whitespace sites, one commit
`lexer.rs:252` (inter-token skip, which also governs `NIL ::= '(' WS* ')'`) and
`lexer.rs:615` (`ANON`). Together, because `ANON` alone flips `{ [<NBSP>] }`
from refuse to accept.

**The `[ # comment ]` fix is not in the lexer, and must not be put there.**
`lex_bracket_or_anon` is comment-blind by construction and has to stay so:
`[ # c\n ?p ?o ]` is a lawful blank-node property list, so the lexer cannot tell
the empty case from the populated one. Teaching it to skip comments would
*increase* acceptance and is grammar-wrong — name that trap so no one takes it.

Only the parser can decide, and **both parsers accept an empty bracket pair**:
`crates/sparql-algebra/src/parser.rs:2723-2731` (whose rustdoc asserts "An empty
`[]` (SPARQL ANON) is legal"), plus `:2701-2711`, `:2882-2892`, `:2930-2934`,
and `crates/rdf/src/native_codecs/text_parse.rs:1342`. Those arms are reachable
only via the comment case, since a real `[]` lexes to `Anon` — they exist to
paper over this. `BlankNodePropertyListPath ::= '[' PropertyListPathNotEmpty ']'`,
so the grammar-correct fix refuses the empty pair in both parsers. **Both files
are owned by this task**, with vectors and neighbours: `[ ?p ?o ]` and
`[ # c\n ?p ?o ]` must both still parse.

Measure the corpora and the rdflib drop-in fixtures before landing it — rdflib
and Jena accept the comment form and the gate may rest on it — and state the
count. No unmeasured refusal rides along.

**Turtle does not flip, which is worse than flipping.** On the SPARQL side an
`ANON`-only change moves `{ [<NBSP>] }` from refuse to accept. On the Turtle
side `text_parse.rs:1342` accepts the empty pair before *and* after, so the
input stays silently wrong either way. Commit ordering hides the SPARQL flip; it
does not fix Turtle. The parser fix above is what closes both at the root.

### Task 4 — the silent misparse, exactly
Restrict `is_pn_chars_base`, `is_pn_chars`, and the `VARNAME` predicates to
their productions. **Two traps that a naive tightening walks into, both of which
break lawful input:**

- `is_pn_chars` is what `take_pn_prefix`, `take_local` and `lex_blank_label`
  actually call, and `PN_CHARS` adds `#xB7`, `[#x300-#x36F]` and
  `[#x203F-#x2040]` beyond `PN_CHARS_U`. Restricting the base without restoring
  those refuses `ex:col·lecció` and **every NFD-decomposed name** — 113 code
  points, and NFD is what macOS filesystems and many exports emit. `ex:café`
  would parse in NFC and fail in NFD.
- `VARNAME` is **position-dependent** and **excludes `'-'`**, which `PN_CHARS`
  includes. One predicate cannot express it; reaching for `is_pn_chars` breaks
  `?a-?b` and `?x-1`. Needs `is_varname_start` and `is_varname_continue`.
  `is_varname_char` is `pub(crate)` — find every caller before changing meaning.

Expect the refusal to surface as a **different token split** at least as often
as a lexer error: after tightening, `?`+NBSP takes the `Question` branch and
fails later in the parser at a different offset. Assert the property, not the
error arm — the shape `no_leading_iriref` (`lexer.rs:976-1010`) already uses.

### Task 5 — exhaustive at the predicate, boundary-exhaustive at the behaviour
Two sweeps, because they cost differently and prove different things.

**Predicate level, every one of 1,114,112 scalars.** Milliseconds — no
allocation, no tokenizer. It is only meaningful if the two sides are
independently spelled, so: implementation is the `matches!`-shaped `const fn`
(ShEx's form), oracle is the transcribed range table with its per-range W3C
citations (`blank_label.rs:526-550`). Implementation against independent
transcription is real evidence; a predicate checked against the table it was
derived from proves nothing.

**Behavioural level, boundary-exhaustive.** A correct predicate composed with
maximal munch, the `?`-disambiguation peek (`lexer.rs:271`), the separate `ANON`
scan and the trailing-dot pushback can still be wrong — that composition is where
the defect actually lived. But 1.1M `format!`-plus-tokenize calls per production,
five productions, in an unoptimized test profile, is **minutes on every
`make check`**. So: (lo−1, lo, hi, hi+1) for every range across all productions
(~400 scalars), plus the full Unicode `White_Space` set, plus the invisible
format characters (U+200B, U+200E/F, U+2060, U+FEFF, U+00AD), plus the nine
IRIREF delimiters. About 450 cases, sub-second, and complete *at every boundary*
— which is the only place a range table can be wrong. Keep a full behavioural
sweep as an `#[ignore]`d lane for anyone who wants it.

Snapshot the derived ranges with `insta` (already a dev-dep) so a future
predicate edit is a reviewable range diff rather than a green run.

**The sweep does not replace the named vectors; it complements them.** It
satisfies the refusal law's substance mechanically — every accepted scalar is an
executed valid case — but a failure prints "0x2028 disagreed with the table",
which tells the next reader nothing about *why* the boundary sits there. Keep
about ten named vectors as the record: U+1680; U+00B7 and an NFD `é`; U+3000;
**U+200C ZWNJ accepted**, which is the trap for whoever writes the diagnostic
table; `?a-?b`; and the day-old IRIREF pair at `lexer.rs:954-1010`.

**Every refusal list in this plan is generated from this sweep, never hand
written.** My first draft's list omitted U+205F and U+2029, which is the
argument in one line: a hand-picked set of a 26-member class will miss members,
and the missing ones are the ones nobody pictures. Enumerate the Unicode
`White_Space` set, assert the predicate is true for exactly `{09, 0A, 0D, 20}`,
false for U+000B and U+000C — the two-sided `is_ascii_whitespace` trap — and
carve out U+1680 with its reason in place, so the carve-out documents itself
rather than living in a comment someone deletes.

Position matters as much as membership. `PN_CHARS_BASE ⊂ VARNAME`, so
`?s<U+1680>?p` lawfully remains **one** variable after the fix — visually
identical to the headline bug, and the next reader will "fix" it and manufacture
the over-refusal. Pin it positively with the reason. Sweep head and tail
separately, and include the tail positives a head-only predicate would refuse:
`?a·`, `?a\u{0300}`, `?a\u{203F}b`, `ex:a·b`.

This *derives* rather than asserts the fact the whole over-refusal guard rests
on — that **U+1680 is the unique scalar both in Unicode `White_Space` and in
`PN_CHARS_BASE`** — and it makes every accepted scalar an executed valid
neighbour, satisfying the repo's refusal law by construction rather than by my
having imagined the right cases. It is also what catches the `PN_CHARS` and
`VARNAME` traps above if the implementation gets them wrong.

### Task 6 — the diagnostic is most of the user-facing value
The character is invisible; that is why the bug is. Today's arm says
`unexpected character '\u{a0}'`. A table of the `White_Space` set plus the
invisible `Cf` offenders (U+200B, U+200E/F, U+2060, U+FEFF, U+00AD) with names
and a hint:

> `U+3000 IDEOGRAPHIC SPACE cannot separate tokens: WS is #x20 | #x9 | #xD | #xA. Replace it with a space.`

**U+200C and U+200D are in `PN_CHARS_BASE`** and must not be listed as
offenders. Vectors assert the message text for U+00A0 and U+3000.

### Task 7 — the line paths, including the lines that are only whitespace
`text_parse.rs:437` uses `raw.trim()` (Unicode class); `column_in_raw` (`:95`)
uses `trim_start()` and must move with it or columns shift. The real-world
breakage is a layer above separators: **a line consisting only of U+00A0 trims
to empty and is skipped as blank today**, and an NBSP before `#` trims away, so
after the fix such a line is neither blank nor a comment. Vectors on all three
line paths (sequential, chunk-parallel, `LineStreamParser`).

Decide the **leading UTF-8 BOM** here rather than letting it fall out: U+FEFF is
not `is_whitespace` but is above 0x7F, so today it becomes part of a `Word` and
fails with a keyword error naming nothing. Either strip exactly one leading
U+FEFF at each line-path entry, or refuse it with the Task 6 diagnostic naming
`U+FEFF ZERO WIDTH NO-BREAK SPACE`. Quote the deciding spec clause in the
rustdoc at the entry point. Whichever is chosen, ship vectors on all three line
paths asserting the chosen behaviour for a single leading U+FEFF, plus the two
neighbours — a *second* consecutive U+FEFF, and a mid-line U+FEFF — both of
which stay refused under either choice. A decision satisfiable by writing a
sentence is not a decision this plan can carry.

### Task 8 — ShEx and ShapeMap
`crates/shex/src/lexer.rs:213` and `crates/shex/src/shapemap.rs:704`. ShEx's
name predicates are already correct and are the donor in Task 1; only its
whitespace skip is wrong.

**Also `shapemap.rs:731-733`**, where `take_keyword` decides a token boundary
with `next.is_alphanumeric() || *next == '_'` — the same defect class, in the
same file, found while auditing it. See it, own it.

The vector list for this task is **not hand-written**: Task 5's behavioural
boundary sweep runs over **three scanners, not one** — `sparql_algebra::lexer`
(both `tokenize` and `tokenize_turtle`), `shex::lexer`, and `shapemap.rs`'s
scanner including `take_keyword`'s boundary — with the same (lo−1, lo, hi, hi+1)
cases per range, the full Unicode `White_Space` set and the invisible format
characters, asserting per scanner that a probe splits into two tokens **iff** the
cited production says so. Each has its own maximal-munch composition, and the
plan's own argument is that composition is where the defect lives; checking two
of the four scan sites by prose while the others are checked mechanically would
leave exactly the gap this issue is about.

Task 8 ships the sweep's derived range snapshot plus the named positives that
record *why*: a ShEx shape label carrying `·` and an NFD `é`, `START` followed
immediately by a non-`WS` boundary character, and U+1680 remaining a name
character in ShEx as in SPARQL.

Cite the production per site. ShapeMap's grammar is not ShExC's, so "inherits
Turtle's `WS`" needs the ShapeMap production quoted for itself rather than by
association — that would otherwise be the weakest-founded refusal in this plan.

### Task 9 — a gate, so there is no seventh
`scripts/check-terminal-predicates.py`, modelled on
`check-iri-resolver-singleton.py` — whose own docstring describes this exact
history for resolvers. It refuses, in scanner-shaped first-party sources:
Unicode-property predicates (`is_whitespace`, `trim*`, `is_alphanumeric`) in a
file that also defines `lex_*`/`skip_trivia`; `(c as u32) > 0x7F` used as a name
predicate; and a local terminal predicate that does not reach `terminals`. Same
`ALLOWLIST[(path, rule)] = reason` shape, same **stale-exemption** failure. The
five known-correct sites Task 9 would otherwise document in comments become its
initial table, so the audit lands somewhere `make check` runs instead of
somewhere a reviewer must remember. `pn_local_allows_slash` (`lexer.rs:758`) is
the model exemption: a named, cited, justified deviation. The allowlist becomes
the workspace's deviation ledger.

### Task 10 — gates, bench, PR
`make check`, `make wasm`, `make doc`, **`make conformance` in full** (name the
rows: SPARQL evaluation, syntax codecs, ShEx validation, ShEx syntax, Python
binding suite, rdflib drop-in gate), and
`crates/sparql-conformance/tests/cdt_lexical_divergence.rs`, which pins exact
tokenization counts and is the tripwire if any corpus file's tokens move.
Run `benches/tokenize.rs` — report only, assert nothing.

**Re-measure in this branch, per site, with the glob written down.** The
inherited "3,879 files, zero offenders" is second-hand, is a *sparql-algebra*
measurement, and licenses nothing about the line path, ShEx or ShapeMap, which
read different corpora and have no measurement at all. Treat it as unverified.

Run a differential tokenization — old predicate against new — over an explicit
glob (`vectors/**`, `queries/**/*.rq`, `crates/**/tests/data/**`, the vendored
W3C suites, the rdflib and Python fixtures, `*.ttl|trig|nt|nq|rq|ru|shex|smap`),
asserting per-file token-stream identity and **naming every differing file**.
Repeat separately for each entry point, since Turtle uses `tokenize_turtle` and
the line path uses `tokenize`. For the files that do contain such code points,
report *where* each sits — string body, IRIREF body, inside a name, at a token
boundary — because "contains" is not the question. Record the glob beside the
count so the next tightening can re-run it rather than re-invent it.

A corpus with zero offenders is evidence of no observed regression, not evidence
of no over-refusal: the over-refusal lives in input the corpus does not contain.
That is what Task 5's sweep and the range-boundary table are for.

Widen the scan beyond the corpora to where a CJK input method actually reaches:
`docs/book/po/zh-Hans.po` and the rendered zh book (`check-i18n-render.py` runs
the fence parser over it), `docs/playground/examples/`, `bindings/python/tests`,
`crates/rdf-wasm/js`, and the shipped-examples harness that parses SPARQL out of
`--help`, `.pyi`, `.d.ts` and rustdoc.

## Enhancements considered and declined

Closed decisions with technical grounds, not deferrals.

- **A `u128` ASCII bitmask for the runtime `WS` test.** LLVM already lowers
  `matches!(b, b' ' | b'\t' | b'\r' | b'\n')` to the same subtract-compare-bit-test
  sequence, so the mask duplicates a compiler optimization while destroying the
  one property this whole change exists to create — that the production is
  legible at the site that tests it. Its only justification would be throughput,
  and repo law makes benches report-only, so that justification can never be
  established here. The cardinality assertion is kept without the mask.
- **A trait with associated consts per terminal.** There is no polymorphism
  here: every call site knows which production it wants, and W3C terminals do
  not vary. What a trait adds is a *blessed* way to write a sixth spelling of
  `PN_CHARS_BASE` — publishing an extension point against a defect history of
  forked spellings is backwards. It is also half-impossible on stable, since a
  `const fn` trait method needs a nightly feature this repo forbids. Forks are
  prevented by the gate, not by a type. The one axis that genuinely varies is
  dialect, and the codebase already answers it correctly with a config field
  (`options.pn_local_allows_slash`).
- **A proc macro parsing production text.** It would need a new proc-macro crate
  on the build graph of five lexers, with `syn`/`quote`, a crates.io bootstrap,
  and compile cost on every build — before a release-blocking deadline. It buys
  agreement between a doc comment and a table three lines apart that have never
  drifted, and buys nothing against the drift that has happened six times
  (three copies in three crates, which the shared module solves). And it would
  hide the ranges from the source-text gate.
- **Type-state or newtype wrappers.** Type-state models a protocol over time;
  set membership over `char` has no protocol. The bug was the wrong predicate
  applied to the right type, and no newtype distinguishes two values of
  `fn(char) -> bool`.
- **Grammar codegen into `generated/`, and a new `purrdf-terminals` crate.** The
  generator cannot validate a table against the spec, so generated tests would
  only prove implementation-versus-table consistency, which a `const` table
  gives free. A new publishable crate drags the token-then-trusted bootstrap,
  publish-order and version gates onto the critical path of a release blocker;
  a module in `purrdf-iri` is internally identical and costs none of it.
- **`proptest` over the productions.** Random sampling of a finite 2²⁰ alphabet
  is dominated by enumeration, and a test that fails only sometimes is in
  tension with the repo's determinism law. Keep proptest for structural
  properties.

Everything adopted is stable Rust — `const fn` with `while` and slice indexing,
`assert!` in const context, `char` range iteration — so the workspace's
zero-nightly-features law holds, with no new dependency and no wasm impact.

## Risks

**Over-refusal is the whole direction of this change.** The three sharpest, in
order of population: `PN_CHARS`'s combining-mark block (113 code points, every
NFD name); `VARNAME`'s `'-'` exclusion (`?a-?b`); and U+1680, the single scalar
that is both whitespace and a name character. Task 5 catches all three
mechanically; no hand-picked vector list does.

**Regressing the day-old IRIREF fix** in the same file, which went the opposite
way. `is_iriref_forbidden_byte` stays untouched;
`crates/rdf/tests/relative_iri_base.rs:493-542` and `lexer.rs:937-990` are the
neighbour set.

**A CJK input method** is the most likely source of a real U+3000 — including
inside a fenced query in the translated book, which the maintainer cannot read.
Task 6's diagnostic and Task 10's widened measurement are the mitigations.

## Verification

Per task: `cargo test -p <crate>`, `cargo clippy -p <crate> --all-targets`,
rustdoc `-D warnings`. End to end: all four `make` gates, the tokenization
tripwire, the bench, and the headline query driven through the **public parser**
— refusing, and naming the character — rather than a lexer unit test asserting a
token vector, because the defect's whole character is that it survives to the
answer.

**Two public entry points, not one.** The second example in the Context,
`ex:a<NBSP>ex:b` minting one term instead of two, is Turtle-shaped, and Turtle
enters through `tokenize_turtle` with different options (`pn_local_allows_slash`).
A silently wrong IRI written into a data file outlives a wrong query answer, so
the Turtle and TriG case gets its own end-to-end vector through `parse_dataset`,
not merely the line-path coverage of Task 7.

paudley:
# Stage 2 gap remediation — issue #287 / PR #290

## Context

PR #290 delivers the approved stage-1 scope: the terminal classes have one home, the silent misparse is closed, every scanner the plan named is corrected, and `make check` / `make doc` / `make conformance` are green with no conformance row moved and no corpus, budget or xfail ledger touched. The deferral scan is clean — zero hits across the diff, eleven commit messages, the PR body and every comment. `.deficiencies` is empty.

Three independent audits — the gap-analyzer, the completion-adversary, and CodeRabbit's review on the PR — agree that the *named* scope is delivered and that the gaps are in **what "everything" meant**. The maintainer's scope statement was "Everything including ShEx". The gate I installed governs **9 files**, and both audits found character-cursor scanners it cannot see, terminal-class defects in shipped codecs nobody looked at, and one silent data-drop in the very file this issue's line-path task rewrote.

Two findings are corrections to claims I made, and they are the reason this stage is not cosmetic:

1. **The boundary sweep does not prove what the PR says it proves.** It reads both the probe corpus and the expected value off the predicate under test, so it is a scanner↔predicate *composition* test, not a table oracle. Proved by mutation: widening `PN_CHARS_BASE` by one scalar leaves all four sweeps green across 35.6M probes. The tables are genuinely guarded — by `terminals.rs`'s independent `matches!` transcription, the `cardinality`/`ranges_all_ascii`/`sorted_disjoint` const-asserts, and the tracked snapshot — but that is not what the PR credits, and `crates/shex` has no snapshot at all.
2. **The gate's blind spot is larger than I disclosed, and my stated reason for accepting it is measurably false.** I wrote that widening the cursor pattern "would sweep in every config and rendering path those crates own". Widening `self\.pos` → `self\.(pos|position|cursor)` governs exactly **one** more file and produces **zero** new offenders. That file is `crates/cdt/src/parse.rs` — edited by this PR, named in the PR body, invisible to every scanner rule.

Standing constraints: `.goals` — GREENFIELD, NO COMPROMISES, MAXIMAL UTILITY, wasm-clean. `.baseline` — no deferrals, no follow-ups, **when you see it, you own it**, no issue/PR numbers in repo files.

## Gap list

### CRITICAL

**G1 — Lone `#xD` is not a line terminator, and `parse_one_line` silently discards trailing tokens.**
`crates/rdf/src/native_codecs/text_parse.rs`. The file quotes `EOL ::= [#xD#xA]+` at `:554`, then splits on `\n` only (`str::lines()` `:454`, `memchr(b'\n')` `:345,:399`). Separately, `parse_one_line` (`:623-626`) calls `expect_dot()` and returns `Ok(Some(nodes))` **without checking the cursor is exhausted** (`at_statement_end` `:706-707` accepts `None | Dot`). So `<s> <p> <o> .\r<s2> <p2> <o2> .` parses one triple and drops the second — exit zero, no diagnostic. Two independent silent-drop paths. This is the bug class the issue exists to eliminate, in the file its own line-path task owns.

**G2 — The gate can be silently blinded by a two-line source edit.**
`scripts/check-terminal-predicates.py:239-262`. `CFG_TEST` matches bodyless `#[cfg(test)]` items (`#[cfg(test)] use foo;`); `source.find("{", …)` then grabs the *next* item's brace and blanks real scanner code. Reproduced: a scanner with `is_whitespace()` yields `['unicode-whitespace']` unblanked and `[]` after blanking. **Active today** — five `src` files carry a bodyless `#[cfg(test)]` item (`xsd/temporal.rs:55`, `entail/owl_dl/mod.rs:47`, `shapes/report.rs:27`, `rdf/gts_view.rs:30`, `rdf/gts_resolve.rs:32`), and the two unconditional rules cannot see past it in any of them.

**G3 — The sweep is presented as a table oracle and is not one; ShEx has no snapshot.**
`crates/shex/tests/scanner_boundary_sweep.rs` computes `derived_ranges` at `:134` and never renders it — no `insta::` call, no snapshots directory. Fix the coverage gap *and* correct the PR's Evidence prose to describe what the sweep actually proves.

### HIGH

**G4 — Gate detection: close the measured blind spot, replace the heuristic's silent excuse.**
Widen `CURSOR_POS` to `self\.(pos|position|cursor)` (+1 file, 0 new offenders — measured). Add a named `SCANNERS` registry with stale-entry failure for split-file scanners (`geo/src/geojson.rs`, whose cursor lives in `json.rs`), so a file the heuristic misses is an *explicit ledger entry*, not a silent pass. Correct the false rationale in the docstring and PR.

**G5 — XML `NameStartChar`/`NameChar` approximated, on an egress path, twice.**
`crates/rdf/src/native_codecs/rdfxml.rs:1662-1679` uses `is_alphabetic`/`is_numeric`. Over-accepts U+00AA (not in the production) so `…#ªx` is **written as `ns:ªx` — RDF/XML we emit that no XML parser can read back**. Under-accepts `[#x300-#x36F]`, so every NFD local name degrades. Second undeleted transcription at `crates/rdf/src/projections/csvw/table.rs:832-843`, gating `xsd:Name`/`xsd:NCName` on ingress with the same two-sided error.

**G6 — The already-fixed defect, still standing in sibling codecs.**
`crates/rdf/src/native_codecs/hextuples.rs:128` — `line.trim().is_empty()` skips an NBSP-only line as blank. `crates/rdf/src/native_codecs/trix.rs:182,187,295` — Unicode-trims `<uri>`/`<id>` text, silently rewriting a lawful IRI (XML `S` is the four). Both ungoverned by the gate.

### MEDIUM

**G7 — XSD lexical spaces trimmed with the Unicode class.** `whiteSpace=collapse` is the four.
`crates/shapes/src/constraints.rs:1918,1926` (integer/decimal — a SHACL validator accepting an ill-typed literal); `crates/rdf/src/projections/csvw/table.rs:810` (`xsd:token`, over-refusal direction); `crates/entail/src/owl_dl/parser.rs:773,875,932` (boolean/integer feeding `owl:hasSelf` and cardinality — a derived entailment from input the datatype does not admit).

**G8 — `crates/rdf/src/native_codecs/okf/reader.rs:805`** — `ch.is_whitespace()` ends a CommonMark link destination where CommonMark names ASCII space or control.

**G9 — Two dialect scanners never examined, and the allowlist has never been exercised.**
`crates/rdf-core/src/sssom.rs` (13 sites; `:577,607,634` skip NBSP-only lines) and `crates/markdown/src/dialect/markdown.rs` (14 sites). `markdown.rs` may be **correct** — CommonMark names a Unicode class in some contexts — but carries no citation and no `ALLOWLIST` entry. `AGENTS.md` now says the allowlist exists for exactly those cases; it has one entry and this is not it. Cite per site; allowlist the legitimate ones.

**G10 — Three `escape_iri` egress forks.** `crates/rdf-core/src/turtle_render.rs:751`, `crates/rdf-core/src/turtle.rs:175`, `crates/rdf/src/native_codecs/ser_model.rs:378`. Behaviour is safe (over-escaping is lossless), but the PR claims "the last forks collapsed" and these did not.

**G11 — Owed evidence.** The plan required the empty-bracket refusal to ship with a measured corpus count; none appears. Measured: **0** comment-emptied bracket pairs and **0** leading BOMs over 5,014 corpus files; **1** head-class hit, `vectors/shexTest/negativeSyntax/PN_LOCAL-dash-start.shex`, a negative vector that must fail. State these in the PR and extend `--census` to the head-class, bracket and BOM populations so the next tightening inherits a re-runnable measurement rather than a number in a review thread.

**G12 — `sh:pattern` regex dialect.** `crates/shapes/src/constraints.rs:2358-2388` compiles with Rust `regex`, whose `\s` is Unicode. Both audits agree the general case is a **different defect class** (XSD-regex semantics: class subtraction, `\i`/`\I`/`\c`/`\C`, block escapes, `q` flag, `^`/`$`) and correctly not bolted onto a release-blocking scanner fix. But the **`x` flag** at `:2375-2377` maps to `RegexBuilder::ignore_whitespace`, and XPath's `x` removes exactly `#x20 #x9 #xD #xA` — that one line *is* this issue's production. Fix the `x` flag here; open a tracking issue for XSD-regex semantics so it lives in GitHub, not in a PR paragraph.

### LOW

**G13** — `crates/rdf/src/native_codecs/media_type.rs:394` (RFC 7230 `OWS` is SP/HTAB); `crates/rdf/src/projections/csvw/table.rs:748` (`is_ascii_whitespace` for base64 — the exact U+000B/U+000C trap the gate names, in a file it cannot see); `crates/shex/src/semact.rs:129` (inside the crate the maintainer named); `crates/gts/src/openpgp.rs:103,145`, `crates/rdf/src/native_codecs/okf/writer.rs:777`, `crates/gts/src/tar.rs:860`. Audit, then cite or fix each.

### Adjudicated — no action

- **`bits.rs` "out of scope"** (CodeRabbit pre-merge check): keep. One-line test-only fix; `origin/main` currently fails `cargo check --tests` without it, so splitting it blocks all verification for no benefit. Reply on the PR rather than leaving the check unanswered.
- **`OnceLock` probe corpus** (CodeRabbit nitpick): decline with reason on the PR — 0.615 s gated, and caching adds state to a test whose value is that it recomputes from the predicates each run.
- **The `blank_label.rs` allowlist entry**: verified legitimate. `egress_tables_agree_with_the_shared_scanner_terminals` (`blank_label.rs:1403`) is a plain `#[test]`, sweeps all scalars, runs under `make check`.

## Execution

Commit after each gap; never batch. Every fix ships **executed valid neighbours**, not reasoned ones — over-refusal is the risk direction throughout and G5 is the one most likely to produce it (the `[#x300-#x36F]` block is 113 code points).

Parallelised by crate to avoid worktree collisions:

| Wave | Owner | Gaps |
|---|---|---|
| 1 | me, `scripts/` only (no build) | G2, G4, G11 census extension |
| 2 | `crates/iri` | XML `NameStartChar`/`NameChar` into `terminals.rs` (prereq for G5) |
| 3a | `crates/rdf` codecs | G1, G6, G8, G13-partial |
| 3b | `crates/rdf` projections + `crates/shapes` + `crates/entail` | G5-csvw, G7, G12 |
| 3c | `crates/rdf-core` + `crates/shex` + `crates/gts` + `crates/markdown` | G9, G10, G13-partial |
| 4 | `crates/rdf` | G5-rdfxml (after wave 2) |
| 5 | `crates/shex` tests | G3 snapshot |
| 6 | me | PR prose corrections (G3, G4, G11), CodeRabbit replies, tracking issue |

## Verification

- Per gap: `cargo test -p <crate>`, `cargo clippy -p <crate> --all-targets -- -D warnings`, rustdoc `-D warnings`, `cargo fmt -p <crate>`.
- Gate: `python3 scripts/check-terminal-predicates.py` **and** `--self-test` must exit 0, with a new self-test case proving the stripper is bounded (G2) — the current self-test proves the rules fire but not that the blanker stops.
- G1 end-to-end: a two-statement document separated by a lone `#xD`, driven through `parse_dataset` and the CLI on all three line paths (sequential, chunk-parallel, streaming), asserting **both** statements arrive. Plus a leftover-token document asserting a hard error.
- G5 end-to-end: round-trip an NFD local name and a U+00AA local name through the RDF/XML **writer** and read the output back with the reader — the current code emits what it cannot re-read.
- G3: mutate a `terminals.rs` range by one scalar and confirm something now fails.
- Final: `make check`, `make doc`, `make conformance` — foreground and serial; background shells are reaped on this box. Conformance must stay GREEN with no row moved and no ledger edited.
- Re-run the completion-adversary as the stage exit gate; it must return PASS.

paudley:
## Stage 2 complete — completion-adversary returns PASS

Three independent audits ran against this branch: a gap analysis, a completion adversary (three passes), and machine review on the PR. They agreed that the named scope was delivered and then attacked what *"everything"* meant. **Thirteen further gaps were closed**, plus five machine-review findings and three more the widened gate found on its own.

### The nine acceptance criteria — all DEMONSTRATED-IN-PRODUCTION

Each was driven through the CLI, not a unit test:

| | Criterion | Demonstration |
|---|---|---|
| C1 | The four `WS` code points, `is_ascii_whitespace` refused as the fix | U+000C and U+000B refused by name; all four accepted as separators |
| C2 | Name predicates restricted to their productions | The headline query refused at **byte 20** — exactly the NBSP; the ASCII neighbour returns one correct row |
| C3 | Both lexer sites together | `{ [<NBSP>] }` refused; `{ [ ] }` and `{ [\t\r\n] }` both answer `true` |
| C4 | Line path trims the four, column in step | All three lanes, including a 1.58 MB document above the parallel threshold; diagnostics byte-identical |
| C5 | ShEx lexer and ShapeMap corrected | Both refuse U+00A0 and U+000C; TAB/CRLF separators still conformant |
| C6 | Every refusal has an executed vector with its neighbour | **33-point sweep of the whole `White_Space` property + the invisible set: 33/33 match the production** |
| C7 | U+1680 still a name character | `?a<U+1680>b` binds as one variable |
| C8 | A lawful IRI containing U+00A0 still parses | Survives a convert round-trip byte-for-byte (`61 c2 a0 62`) |
| C9 | `make conformance` re-run, nothing edited to accommodate | GREEN; **row-by-row diff against the pre-remediation run is identical** |

### What review found that the first round did not

- **A phantom prefix flipped a validation verdict.** The SHACL text ingest scanned raw document text for `@prefix` directives with Unicode whitespace, so a U+00A0 inside a *string literal* was scanned out as a declaration — turning a hard error into a reported Violation on otherwise identical inputs. Narrowing the class alone traded it for the mirror bug (the junk label survived into the generated `PREFIX` header, breaking every SHACL-AF query in the document), so the label is now held to `PN_PREFIX` as well. The limit that remains is documented rather than implied: a plain-ASCII `@prefix` inside a literal still mints the phantom, because this is a scan and not a parse.
- **Two silent data-drops in the line parser**: a lone `#xD` was not a terminator though the file quoted `EOL ::= [#xD#xA]+`, and the parser returned without checking the cursor was exhausted, discarding every token after the `.`. A *missing* `.` also parsed clean.
- **The RDF/XML writer emitted predicates it could not read back.** The fallback declared the whole IRI as a namespace and emitted `nsN:property`, which a reader concatenates to `<whole_iri>property` — a different predicate, at exit zero. Deleted, not guarded: the split now walks back to the longest lawful `NCName` suffix, so a stricter class routes *zero* IRIs into a lossy path.
- **Nine transcriptions of the same tables**, not the four the PR originally claimed — five `escape_iri`, and four of the XML Name family. All collapsed, each proved byte-identical over all 1,114,112 scalars.
- **A left-padded ustar `SIZE` field read as zero**, silently truncating a member body.
- **An over-refusal this branch itself introduced**: tightening ShapeMap's `IRIREF` content class turned a pre-existing wrong-value bug into `unterminated IRI` for `<urn:ex:a>` — which is what PurRDF's own IRI egress emits. `UCHAR` is now decoded, by extracting the ShExC lexer's existing decoder rather than writing a second one.

### Corrections to claims this PR made

Three things it asserted turned out to be wrong, and the record is corrected rather than quietly adjusted:

1. **The boundary sweep is not a table oracle** for six of its eight productions — it reads both the probe corpus and the expected value off the predicate under test. Measured: widening `PN_CHARS_BASE` by one scalar leaves all four sweeps green across 35.6M probes; only the snapshot fails. It *is* an oracle for the two content classes that are transcribed twice.
2. **The gate's blind-spot rationale was asserted, never measured, and was false.** Widening the cursor pattern governs one further file and produces zero new offenders — and that file was a real scanner the whole time.
3. **`AGENTS.md` and the gate both stated a false CommonMark claim** that licensed `char::is_whitespace` in the Markdown dialect. CommonMark's blank line is "only spaces (U+0020) or tabs (U+0009)"; its Unicode whitespace class governs emphasis flanking alone. The unit of judgement is the **clause**, never the specification.

Also corrected: the `[#x300-#x36F]` combining block is **112** code points, not the 113 stated throughout (113 counts the block plus `#xB7`).

### Deliberately not fixed, on the record

- **`csvw/config.rs`'s prefix validator** stays liberal. It validates JSON-LD *term* keys, and JSON-LD 1.1 §3.1 admits most non-keyword strings — tightening it to `NCName` would be an **unsourced refusal**. Two audits reached that independently, so it is now an `ALLOWLIST` entry quoting the clause, where it cannot be quietly reversed.
- Two issues filed for genuinely separate defect classes, neither reachable from any of the nine criteria: the XML writers' CR/C0 escape defect (pre-existing; fixing it moves emitted bytes, so it is reviewed as such), and the `sh:pattern` XSD-regex dialect gap (which until now existed only as a promise in a source comment).

### Verification

`make check` all stages, `make doc`, `make wasm`, `make conformance` — all green. **8,342 tests, 0 failed.** Conformance GREEN across 21 rows with **no row moved** and no corpus, fixture, baseline budget or xfail ledger touched. `check-terminal-predicates.py` exits 0 with two reasoned exemptions; the gate governs 11 files plus 5 ledger entries where the structure test cannot reach.

`.deficiencies` is empty.

### One note carried forward, non-blocking

The gate's widening check is regex-shaped, so it remains evadable by a bitwise `|`, an `if/else`, or a delegating `match` arm. The adversary enumerated every terminal-named function the gate clears and inspected each cleared body: no such shape exists anywhere in the tree today, and the check's own documentation scopes its claim precisely to "refusing a surviving alternation" rather than to widening in general. A regex gate over Rust cannot be complete without becoming a parser. Worth a future tightening; it hides nothing now.

**Ready for stage 3.**

paudley:
fix(scanners): a scanner's character classes decide token boundaries, so make them exact

Closes #287.

`crates/sparql-algebra/src/lexer.rs` is the shared front end for SPARQL queries
and updates, Turtle, TriG, N-Triples and N-Quads, and therefore for the CLI, the
Python, WebAssembly and C surfaces. It approximated its grammar's character
classes with convenient predicates: `char::is_whitespace` where
`WS ::= #x20 | #x9 | #xD | #xA`, and "anything above U+007F" where
`PN_CHARS_BASE` and `VARNAME` have exact range tables.

Non-ASCII whitespace is above U+007F, so the greedy name scan swallowed it
before the whitespace skip was ever consulted:

    SELECT ?s WHERE { ?s<NBSP><urn:ex:p> ?o . ?s <urn:ex:q> ?z }

was ACCEPTED with four distinct variables where the author wrote three. A join
silently became a cross product and projected a variable the first pattern never
binds. Exit zero, no diagnostic, wrong rows. The Turtle mirror, `ex:a<NBSP>ex:b`,
minted one IRI instead of two and wrote it into a data file, where it outlives
any query answer.

WHY IT IS A CLASS AND NOT AN INCIDENT

These predicates run inside a maximal-munch scan, so a superset character class
does not merely widen the accepted language — it changes the TOKENIZATION of
documents both parsers accept. `?s` NBSP `<urn:ex:p>` is lawful input to a
conforming parser and to this one, and they disagreed about what it means.

The class also had a written authorization here. `blank_label.rs` stated that a
lexer may safely over-accept because it is "sound on ingress — a superset lexer
can only accept documents a conforming parser would also accept, never reject
one". That lemma is false for any predicate deciding a token boundary, which is
every predicate in a scanner. Fixing the instances and leaving the paragraph
would have authorized the next one. It is replaced with the corrected law, and
AGENTS.md carries it as a hard constraint.

WHAT LANDED

Every W3C terminal is now spelled once, in `purrdf_iri::terminals`, with its
production cited and its ranges asserted at compile time; `purrdf-iri` stays a
zero-dependency wasm-clean leaf. Corrected scanners: sparql-algebra (both
whitespace sites, the PN_CHARS and VARNAME families, three head classes, PERCENT,
PN_LOCAL_ESC), ShEx and ShapeMap (seven sites, each cited from ShapeMap's OWN
grammar rather than by association with Turtle), the three N-Triples line paths,
cdt, geo (WKT and GeoJSON), sparql-results, hextuples, trix, okf, media_type,
rdfxml, ustar, sssom, linkml, csvw, rif_xml, gts tar and openpgp, and the SHACL
text ingest.

THE WORKSPACE HELD NINE INDEPENDENT TRANSCRIPTIONS OF TABLES IT ALREADY OWNED —
five of the IRI escape set and four of the XML Name family. All are collapsed,
each proved byte-identical over all 1,114,112 scalars against the spelling it
replaced. `rdf-core`'s PN_CHARS table is deliberately kept as an independent
EGRESS oracle and is one of the gate's two reasoned exemptions, valid only while
an all-scalar equivalence test proves the two agree.

SILENT DATA LOSS FOUND ALONG THE WAY

- A lone `#xD` was not a line terminator though the file quoted
  `EOL ::= [#xD#xA]+`, so two statements separated by one arrived as a single
  line and the second was discarded.
- The line parser called `expect_dot` and returned WITHOUT checking the cursor
  was exhausted, so every token after the terminator was dropped. A missing `.`
  also parsed clean.
- The RDF/XML writer emitted predicates it could not read back: when a local
  part was not an XML Name, the fallback declared the whole IRI as a namespace
  and emitted `nsN:property`, which a conforming reader concatenates to
  `<whole_iri>property` — a DIFFERENT predicate, at exit zero. The fallback is
  deleted rather than guarded; the split now walks back to the longest lawful
  NCName suffix, so a stricter class routes ZERO predicates into a lossy path.
  A predicate nested in a quoted triple was also emitted bound to an undeclared
  prefix.
- A left-padded ustar `SIZE` field read as ZERO, truncating a member body.
- A U+00A0 inside a STRING LITERAL was scanned out as an `@prefix` declaration
  by the SHACL text ingest, flipping `purrdf validate` from a hard error to a
  reported Violation on otherwise identical inputs.

OVER-REFUSAL IS THE RISK DIRECTION, AND IT BIT THIS BRANCH TOO

Three characters look like garbage and are named by the production; each is
pinned POSITIVELY so a later cleanup cannot manufacture a refusal. U+1680 OGHAM
SPACE MARK is simultaneously Unicode `White_Space` AND inside `PN_CHARS_BASE`,
so `?s<U+1680>?p` is lawfully ONE variable — visually identical to the headline
bug and exactly its opposite. U+200C and U+200D are named by `PN_CHARS_BASE`, so
a table built by sweeping invisible format characters would tell authors their
valid queries are wrong. U+FEFF is inside `[#xFDF0-#xFFFD]`, so a byte-order mark
is a lawful NAME character. Being an offender is a property of the POSITION,
never of the character.

That claim is derived, not asserted: intersecting `char::is_whitespace` with each
of the five name classes over every scalar yields exactly {U+1680}.

This branch introduced one over-refusal and repaired it. Tightening ShapeMap's
IRIREF content class turned a pre-existing wrong-value bug into `unterminated
IRI` for `<urn:ex:aa>` — which is what PurRDF's own IRI egress emits. UCHAR
is now decoded, by extracting the ShExC lexer's existing decoder rather than
writing a second one.

TWO CORRECTIONS TO CLAIMS THIS WORK MADE

The boundary sweep is NOT a table oracle for six of its eight productions: it
reads both the probe corpus and the expected value off the predicate under test.
Measured — widening `PN_CHARS_BASE` by one scalar leaves all four sweeps green
across 35.6M probes, and only the snapshot fails. It IS an oracle for the two
content classes transcribed twice. The tables are guarded by an independent
`matches!` transcription, the macro's cardinality and sortedness const
assertions, and tracked snapshots.

And a false claim is removed from this repository's own law. The gate's docstring
and AGENTS.md both asserted that CommonMark defines its whitespace as Unicode
whitespace, licensing `char::is_whitespace` in the Markdown dialect. It does not:
a blank line is "only spaces (U+0020) or tabs (U+0009)". CommonMark does define a
Unicode whitespace character, but it governs emphasis flanking alone. The unit of
judgement is the CLAUSE a site implements, never the specification it belongs to.
Relatedly, the gate's stated blind-spot rationale was asserted and never measured,
and was false: widening the cursor test governs one further file and produces zero
new offenders — and that file was a real scanner the whole time.

A GATE, SO THERE IS NO TENTH

`scripts/check-terminal-predicates.py` runs in `make check`. It refuses a
Unicode-property test inside a file holding a character cursor, the `> 0x7F`
name catch-all the defect shipped as, and any W3C terminal predicate retyped
instead of routed. Scanners are detected structurally, not by name — names are
wrong in both directions. Files whose cursor lives in another module are named in
a `SCANNERS` ledger that rots loudly. `--census` and `--census-refusals` measure
the corpus population a tightening can move, by POSITION: 4,327 files, zero bare
tokenization-affecting occurrences, zero comment-emptied bracket pairs, zero
leading BOMs, and one head-class hit that is a negative-syntax vector whose
purpose is to be refused. The single census hit overall is `<urn:ex:\xa0>` in the
frozen W3C RDFC-1.0 fixtures — a LAWFUL IRI a careless tightening would break.

DELIBERATELY NOT FIXED, ON THE RECORD

`csvw/config.rs`'s prefix validator stays liberal: it validates JSON-LD TERM
keys, and JSON-LD 1.1 §3.1 admits most non-keyword strings, so tightening it to
NCName would be an UNSOURCED refusal. Two audits reached that independently, so
it is the gate's second reasoned exemption with the clause quoted, where it
cannot be quietly reversed.

.goals COMPLIANCE

GREENFIELD / no compromises: nine duplicate transcriptions deleted, not wrapped.
RUST-FIRST: all of it. LOW/NO OPTIONALITY, HARD FAILS: every approximation
replaced by the enumerated production; no Cargo feature added. MAXIMAL
PORTABILITY: `purrdf-iri` stays zero-dependency, `make wasm` green, and the ustar
width conversion is checked rather than cast because `usize` is 32 bits on
wasm32. RDF 1.2 ONLY: the quoted-triple predicate path is fixed, not excused.

VERIFICATION

All 18 CI checks pass on this HEAD. Locally: fmt, 16 hygiene scripts, workspace
clippy `-D warnings`, workspace check, `make rdf-core-hygiene`, `make wasm`,
`make doc` — all exit 0. 8,342 tests, 0 failed. `make conformance` GREEN across
21 rows with a row-by-row diff against the pre-remediation run that is IDENTICAL.
No corpus, fixture, frozen vector, baseline budget or xfail ledger was touched —
`git diff origin/main...HEAD -- vectors/ generated/ 'crates/**/xfail*'
scripts/conformance-baseline.json` is empty.

Bench, reported and not asserted, because this machine is not quiet and the
change pushes both ways: turtle_4k 1.20 ms / 410 MiB/s, sparql_4k 1.80 ms /
274 MiB/s, modifier_query 439 us / 225 MiB/s.

Public API changes are additive only: `purrdf_iri::terminals`,
`purrdf_core::iri_escape`, the `terminals` re-export, and
`purrdf_gts::tar::parse_octal`. The only removal was a `pub(crate)` item. The
breaking part is behavioural — input that parsed begins to fail — which is why
this lands before 2.0 rather than as a later semver violation.

COMPLETION-ADVERSARY: PASS (stage 3, zero open findings)

It spot-checked the criteria on the production surface rather than accepting the
suite: the headline query refused at byte 20, exactly the NBSP, with the ASCII
neighbour returning one correct row; `?a<U+1680>b` binding as a SINGLE variable;
`<urn:ex:a<NBSP>b>` surviving a convert round-trip as `61 c2 a0 62`. Stage 2's
pass-3 audit demonstrated all nine acceptance criteria in production, including a
33-point sweep of the whole `White_Space` property matching the production 33/33.

It adjudicated the one open review objection — that OpenPGP, tar, Markdown, SSSOM
and HexTuples were unrelated work riding along — against the diff rather than
against a slogan, and ruled every named area IN CLASS: each is a scanner or an
egress table deciding a boundary by a Unicode property, or a duplicated
transcription. It enumerated what breaks for a user of main and found NO
over-refusal; every refusal is required by a cited production and ships an
executed valid neighbour. The largest blast radius is `convert --to rdfxml`
refusing a predicate with no NCName suffix, which replaces silent corruption with
a diagnostic.

It also caught a stale claim in the PR body — a `bits.rs` fix that main had
since landed independently, leaving the file out of this diff. The claim is
withdrawn in the body rather than deleted, since a reader comparing it against
the diff would otherwise find a file that is not there.

DEFERRAL GATE: clean. Zero keyword hits across the branch diff; zero in the PR
body. The three hits in PR COMMENTS are the machine reviewer's own out-of-scope
warning and this session's adjudication KEEPING a change — neither is deferred
work. `.deficiencies` holds its notice and marker only.

NOTES FOR LATER

The gate's widening check is regex-shaped and remains evadable by a bitwise `|`,
an `if/else`, or a delegating `match` arm. Every terminal-named function the gate
currently clears was inspected for residual widening and none exists in the tree;
the check's own documentation scopes its claim precisely to "refusing a surviving
alternation" rather than to widening in general. A regex gate over Rust cannot be
complete without becoming a parser.

Two issues were filed for genuinely separate defect classes, neither reachable
from any acceptance criterion here: the XML writers' CR and C0 escape defect
(pre-existing; fixing it moves emitted bytes, the opposite risk profile from this
branch), and the `sh:pattern` XSD-regex dialect gap, which until now existed only
as a promise in a source comment. The one part of that gap which WAS this issue's
class — the `x` flag, where XPath removes exactly four code points and Rust's
`ignore_whitespace` removes 26, treats `#` as a comment and strips inside
character classes — is fixed here, not deferred.

PR #310, opened 2026-09-15
test(gts): expand public decryption boundary coverage

paudley:
Extend the restored public `cose::decrypt0` regression suite from 2 to 12 tests, covering every truncation of the frozen envelope, malformed fields before key lookup, invalid nonce lengths, short authentication tags, and tampering with every nonce/ciphertext/tag byte. A semantically equivalent protected-header re-encoding must fail authentication, proving that the original header bytes remain bound.

Positive cases cover untagged and indefinite-length envelopes, exact recipient identity including Unicode and embedded NUL, and binary payloads across block and CBOR length boundaries through 64 KiB. Existing frozen-vector and wrong-key coverage remains. Only the integration test file changes; production code, dependencies, workflows, and published versions are unchanged.

Validation: 12 public-decryption tests and 6 bounded-reader tests passed; targeted Clippy with `-D warnings` and workspace formatting passed. Full workspace and binding suites were not run locally.

Generated-case round-trips use fresh keys from the existing native test RNG, so those two tests are native-only. The frozen-vector and adversarial envelope cases remain portable. The final head again passed all 18 focused tests and targeted Clippy.


paudley:
test(gts): expand public decryption boundary coverage

Expand crates/gts/tests/cose_public_decryption.rs from 2 to 12 tests: frozen-vector interoperability, malformed and truncated envelopes, nonce and tag boundaries, authenticated header bytes, bytewise tampering, supported CBOR forms, exact recipient identities, and binary payload boundaries through 64 KiB. Generated encryption cases use fresh native test keys.

Validation at d03feb6c: all 18 public-decryption and bounded-reader tests pass; cargo fmt --all --check and targeted cargo clippy with -D warnings pass. CI workspace, documentation, MSRV, projection oracles, and CodeQL pass; other CI jobs are still running. Review threads are resolved.

Standing .goals constraints are satisfied. The emergency ledger is empty and the code deferral scan is clear; the public-text match is the automated “Out of Scope Changes” check label, not unfinished work. Base synchronization required no changes or conflict resolution.

PR #318, opened 2026-09-16
Trim third-party dependencies: first-party language tags, SCC, and hex; pinned cnSchema probe

paudley:
## Summary

Dependency-trimming pass: three third-party crates leave the workspace, each replaced by a first-party
surface, plus a gate that keeps them out and a reproducible interop probe.

- **`oxilangtag` → `purrdf_iri::langtag`** — an RFC 5646 `Language-Tag` parser derived from the §2.1
  ABNF as recursive descent, one function per production line. It carries the grammar once, behind
  four named profiles:

  | profile | what it is |
  |---|---|
  | `Rfc5646` | §2.1 as written |
  | `Rfc5646PrivateUseRelaxed` | lifts the eight-character ceiling on private-use subtags only |
  | `ConcreteSyntaxLangtagBounded` | the `LANGTAG` terminal **plus** the §2.1 eight-character ceiling, which the private-use marker lifts |
  | `ConcreteSyntaxLangtag` | the `LANGTAG` terminal as the concrete syntaxes define it, unbounded |

  The profiles are totally ordered and that ordering is pinned by a test. Beyond well-formedness the
  module provides §2.1.1 canonical-case normalization, singleton-keyed extension iteration,
  private-use subtag iteration, an owning form, and the standard trait set. The boundary is
  well-formedness: subtags are not checked against the IANA registry, and RFC 4647 language-range
  matching is outside this crate's scope.

- **`petgraph` → `purrdf_core::graph::tarjan_scc`** — one iterative Tarjan with an explicit stack, in a
  crate both consumers already reach.

  Worth stating precisely, because the first pass got it wrong: on `main` there was exactly **one**
  hand-rolled copy, in `purrdf-shex`. `purrdf-slice` called `petgraph::algo::tarjan_scc` and declared
  the dependency. Dropping `petgraph` by adding a second hand-rolled copy to `purrdf-slice` traded a
  third-party dependency for a maintained duplicate — which is the shape the standing goals forbid, and
  the ban-list entry then enshrined both copies as "the replacement". The consolidation removes the
  duplicate this change itself introduced.

- **`hex` → `core::fmt::LowerHex`**, plus a first-party `purrdf_core::hex::lower`. Banning a crate for
  a job the workspace then did many different ways in-tree would not have been coherent, so the copies
  fold. `hex::lower`'s test compares against an independently written nibble lookup, so implementation
  and oracle cannot share a bug.

  This description deliberately carries **no count**. An earlier revision said "fourteen", which was
  accurate for one commit and stale the next; correcting it to "twelve, six of them in `purrdf-core`"
  was worse, because no single counting rule makes both halves true — named `fn` definitions give
  twelve workspace-wide but two in `purrdf-core`, while counting inlined render loops gives six in
  `purrdf-core` but far more than twelve overall. A tally in prose has no gate behind it and rots on
  the next commit.

  The durable form is the rule, and it lives in `hex.rs` where a future caller will actually read it:
  anything turning a `&[u8]` into an owned lowercase-hex `String` calls `lower`, and exactly four
  categories correctly do not — hot paths inside a fixpoint loop, renderers writing into a fixed buffer
  or a caller-supplied byte sink, crates that cannot reach `purrdf-core` (`purrdf-gts`, whose only
  first-party edge is the zero-dependency events crate), and operations that are not this one at all
  (`skolem`'s selective escape, the shape-label accumulator).

- **`scripts/check-banned-deps.py`** (in `make check` + CI) reads **every git-tracked `Cargo.lock`**
  and **every git-tracked `Cargo.toml`**, not just the root and its members — the two first-party roots
  excluded from the workspace, the geo determinism harness and the GTS fuzz target, are exactly where
  a banned crate had in fact re-entered unseen. The ban is split by edge kind: the ox-family,
  `oxilangtag` and `petgraph` are banned on **any** edge, while `hex` is banned only as a **direct**
  dependency of a first-party crate — banning a transitive third-party use would fail `make check` over
  an edge that was never PurRDF's own, with no fix available here.

- **`scripts/check-corpus-frozen.py`** now guards `crates/rdf/tests/corpus/w3c` and the two vendored
  SPARQL suites. Those hold every language-direction negative vector this change's correctness rests
  on, and an edit to one would not have tripped the freeze gate. Scoped to the vendored subdirectories
  rather than `suite/` as a whole, because that directory also holds first-party suites that grow with
  feature work — freezing those would turn a loud, reviewable re-vendor into routine noise and destroy
  the signal for the corpora that need it.

- **`make cnschema-probe`** — one command reproducing the pinned cnSchema 4.0 round-trip evidence
  (source SHA-256, RDFC-1.0 canonical N-Quads SHA-256 on both legs, 40,936 quads). Fetch-by-digest,
  never vendored. It runs on a schedule, so its pinned constants cannot rot silently.

## The parity claim is now falsifiable

As originally written the headline claim — that the first-party parser accepts exactly what the crate
it replaced accepted — could not fail. Removing `oxilangtag` from `Cargo.lock` and banning its return
destroyed the only oracle that could contradict it.

`crates/iri/tests/langtag_differential_vectors.txt` freezes **3935 inputs** with accept/reject
verdicts. The inputs are generated independently from the RFC — a systematic sweep over the §2.1 ABNF,
the closed §2.2.8 grandfathered list, Appendix A with case variants, and structural noise. The verdicts
were labelled once by the replaced crate, run as an external oracle outside this repository; no
upstream source or test data was read or copied. 743 accept, 3192 reject.

The table carries a SHA-256 of its own body that the test recomputes, so editing a verdict to make a
change pass is detected. Mutation probes confirm it catches parser regressions rather than only table
edits: raising the subtag ceiling is caught by 56 vectors, loosening the variant minimum by 142,
dropping the ALPHA requirement from `region = 2ALPHA` by 102.

**The ABNF-derived parser matched all 3935 vectors on its first run**, before any test was touched.

## One owner for language tags

The first cut rewired only the two call sites that had imported the replaced crate, leaving every other
surface on its own hand-rolled grammar — the RDF/XML parser took any alphanumeric subtags with no
length cap at all, the Turtle document parser validated nothing whatsoever, and the SPARQL, ShEx, CDT
and CSVW surfaces each carried their own transcription. A tag could bind in a SPARQL query and be
refused by the N-Triples reader.

The CSVW pair was found by audit rather than by any test, after this description already claimed the
job was done: `purrdf project --profile csvw-terms` accepted `1`, `9-9` and `123-456` as language tags
and wrote them into artifacts. Every gate was green over it, because no test crosses into that
subcommand with a malformed tag — the same blind spot that hid the round-trip defect below.

Every site that decides well-formedness now calls `purrdf_iri::langtag`; none carries a character class
or a shape rule of its own. (No count here on purpose — see the hex bullet above on why a tally in
prose rots.) Every site was tightened, so every site carries both
halves: an input it wrongly accepted before that is now refused, and a neighbouring valid input that
must still pass. Error types, byte offsets and public signatures are unchanged; the grammar's reason
and code are surfaced inside the existing payloads, so a refusal names the rule that refused it on
every surface.

### PurRDF could not read its own output

Unifying the readers exposed — and this branch briefly *introduced* — a round-trip defect. The Turtle
reader took the `LANGTAG` terminal while the N-Triples and RDF/XML readers took §2.1, so `convert
--from turtle --to nquads` emitted `"a"@fr-be-fbcl` and reading that same file back failed. Those tags
are lawful `LANGTAG` terminals, they are in the vendored shexTest corpus, and `main` accepted them.
Every gate stayed green, because no test crossed codecs.

Asymmetric acceptance between a writer and its matching reader is always a round-trip bug. The four
codec paths now name one profile, `ConcreteSyntaxLangtagBounded`, which satisfies both fixed points:
the W3C negative vector `cantbethislong` is a fourteen-character primary and stays refused — now
reported as the length it is, rather than a production it never matched — while the long private-use
families downstream projects publish keep parsing. `crates/rdf/tests/langtag_codec_crossing.rs` closes
the hole: each tag is parsed from Turtle, written to N-Triples, N-Quads, RDF/XML and TriG, and parsed
back from each.

This re-widens three paths back to what `main` accepted. Genuine garbage — `1`, `-`, `9-9`, `123-456`,
`en-` — is now refused on **every** path, which `main` did not do.

### One owner is not one acceptance language

Three value spaces are in play, each with a specification behind it, and collapsing them would be a
bug rather than a simplification:

| surface | space | why |
|---|---|---|
| the codecs | `ConcreteSyntaxLangtagBounded` | the `LANGTAG` terminal plus the §2.1 ceiling, which the private-use marker lifts |
| CSVW `lang` / `@language` | `Rfc5646` + an exact-lowercase spelling rule | CSVW Metadata §5.7 requires a BCP 47 tag, and this workspace's own CSVW reader judges it there |
| an `xsd:language` cell value | `ConcreteSyntaxLangtag` + one named subtag ceiling | XSD 1.1 §3.3.3 is a flat pattern — no production structure, no private-use carve-out — so the ceiling cannot be the profile's, which lifts after `x` |

Widening the CSVW property to the codec profile would have been the obvious fix and the wrong one: it
would mint a `lang` that this workspace's own reader drops on the way back in. The visible consequence
is asserted rather than left implicit — a private-use tag whose subtag runs past eight characters is
written by the codecs and refused by both CSVW surfaces, because RFC 5646 bounds `privateuse` at
`1*8alphanum` and the XSD pattern caps every subtag. One test names all three spaces and both specs.

## Scope, stated plainly

- **Canonical-case normalization is library API, not a CLI surface.** Its production consumer is the
  CSVW metadata reader (`read_csvw`), where it replaced an ad-hoc `to_ascii_lowercase`. The CLI's
  `csvw-exact` lift profile refuses `@context.@language` outright, so the normalization is not
  reachable from a shipped executable; in-tree it is exercised by the W3C CSVW harness.
- **Three sites deliberately do not normalize** — language-map keys (which must agree with a lowercase
  fold at lookup), the JSON-LD value walker (which validates the caller's own document rather than
  rewriting it), and the embedding target (whose packed dictionary hard-errors on anything not already
  at the lowercase fixed point).
- **`purrdf-gts` keeps its own renderer, deliberately.** It does not depend on `purrdf-core` and must
  not start — its only first-party edge is the zero-dependency events crate, and adding the IR kernel
  plus `regex`, `csv` and `sha2` to every container consumer to save four lines would invert the
  layering that puts `purrdf-rdf` above both. Its `files` module stopped cloning its own sibling
  `wire::hex` instead, so that crate has one renderer for a local reason.
- **The renderers that stay are enumerated in `hex.rs`, with the reason each is not a stray copy.**
  The UUID builtin is *not* among them — it was exempted in an earlier draft of this description on the
  grounds that folding would add an allocation, which was the inverse of the truth: it built sixteen
  `String`s, a `Vec` and five `join`s for sixteen bytes, so folding removed roughly twenty-one
  allocations. It is folded, and pinned by a 512-iteration test asserting the RFC 4122 version and
  variant nibbles, the `8-4-4-4-12` boundaries, and that the hyphen-stripped string re-parses to
  exactly the bytes it rendered.
- **No breaking change**: no public item removed, no signature changed, no field added to an existing
  public struct, no version edited. The only public-surface change is added items.

## Verification

- Full conformance matrix **GREEN**, 22 suites, 4801 pass / 19 xfail against a budget of 19 / 0 fail.
  Syntax codecs (Turtle/TriG/NT/NQ/RDF-XML) 264/264 with 0 allowlisted gaps; ShEx 2.1 validation
  1105/1105 with zero trait-skips.
- CSVW's W3C harness passes with an empty `XFAIL` list — noted separately because it is **not** a
  conformance-matrix row: it runs as `crates/rdf/tests/csvw_w3c.rs` in the `test` job. The list is
  enforced in both directions, failing on an unexpected pass as well as a stale entry.
- Production surface: `convert --from turtle --to nquads` then feeding that output back both exit 0
  and every tag still parsing on the way back (case is folded at intern time — see below); `@1` in
  Turtle is refused with `langtag-terminal-primary-not-alpha`, reported with its line and column;
  `ntriples-langdir-bad-4` still refused, as `langtag-subtag-length-over-eight`, with its
  eight-character neighbour `cantbeth` accepted to prove the refusal is the length.
- `make cnschema-probe` reproduces all three pinned constants; a digest mismatch exits non-zero on the
  real entry point, not merely in a self-test.
- The banned-deps gate: restoring the pre-fix `crates/geo/determinism/Cargo.lock` makes the real gate
  fail naming that file — it was a genuine violation the root-only scan was blind to. A transitive
  `hex` still passes; a direct `hex` still fails.


paudley:
## Stage 2 — Gap Analysis & Remediation

Three independent audits (context/reviewer-feedback, gap-analysis, completion-adversary) were run
against this branch, and every load-bearing finding was then re-verified directly before being
accepted. **20 gaps** to remediate: 3 CRITICAL, 5 HIGH, 8 MEDIUM, 4 LOW.

### Context

All 18 CI checks are green, and that proves less than it looks like it does. A dependency-replacement
PR destroys its own oracle: once `oxilangtag` leaves `Cargo.lock` (and joins the ban list), the claim
*"acceptance language identical to the replaced crate"* becomes **unfalsifiable** — no test in the
repo could go red if it were false. That is the exact shape the contributor guide warns about ("a
refusal is a claim too — prove it"; over-refusal "hides well: every test passes").

The underlying engineering is **sound**: the parser's acceptance language really does match the crate
it replaces, and the Tarjan implementation really is correct. What is not sound is that the
*provenance record is false*, the *central claim is untested*, and several standing repo goals
(no duplication, subsume-don't-subset) are left violated.

### Decisions taken (maintainer)

| # | Decision | Chosen |
|---|---|---|
| D1 | `langtag.rs` derives from oxilangtag 0.1.6 (MIT) while `PROVENANCE.md` denies it | **Genuine clean-room rewrite** — re-derive from the ABNF, keep the greenfield claim |
| D2 | Byte-identical Tarjan in `purrdf-slice` and `purrdf-shex` | **Hoist to `crates/rdf-core/src/graph.rs`** — common ancestor, zero release-lane cost |

### Verified evidence

- `crates/iri/tests/langtag_corpus.rs:162` contains `ab-abc-abc-abc-abc`; so does
  `oxilangtag-0.1.6/tests/lib.rs:417`. It appears **nowhere in RFC 5646** — it is a synthetic
  four-extlang probe that cannot have been independently derived. `PROVENANCE.md:46` states *"Written
  from the RFC, not from the replaced `oxilangtag` dependency's tests."* That sentence is false.
- `GRANDFATHERED[26]` matches oxilangtag's `GRANDFATHEREDS[26]` in content and order; all seven error
  variants carry oxilangtag's exact names.
- `diff` of `slice/src/cache.rs:264-323` against `shex/src/structure.rs:429-489` → identical but for
  one line's position.
- `Cargo.toml:138` — orphaned half-sentence, now documenting `pretty_assertions`.
- `link_units` (slice's only `tarjan_scc` consumer) has one workspace caller: `cache/tests.rs:203`.
  The shex copy has three live callers.
- cnSchema probe reproduces exactly: source `92329fd9…`, both legs `8a4b7c11…`, 40,936 lines.

### Cleared — will not be re-chased

`cns-core.jsonId` is **correct upstream** (`.jsonId` → 200, `.jsonld` → 404; "fixing" it to `.jsonld`
would have broken a working probe) · acceptance-language parity has no divergence · the Tarjan is
correct (back edge uses `index[child]`, child-return uses `low[node]`, no double-push, deterministic
via the member/unit sorts) · `check-banned-deps.py` is sound, its self-test is genuine, it is wired
into both `make check` and CI, and it exits 1 on violation · `hex`/`petgraph`/`oxilangtag`/
`fixedbitset` are absent from `Cargo.lock` · RFC 4647 and registry validity are **not** a capability
regression (the replaced crate never had them) · wasm and determinism clean · zero TODO/FIXME, zero
issue references, zero banned-name hits in the diff · deficiency ledger empty.

---

## Gaps

Each gap = one fix → one commit → push → comment → wiring-verify. No batching.

### CRITICAL

**G1 — Clean-room rewrite of `crates/iri/src/langtag.rs`** (D1)
Re-derive from the RFC 5646 §2.1 ABNF with genuinely independent state factoring (not a rename of the
replaced crate's states); regenerate `GRANDFATHERED` from §2.2.8 prose; rename all seven error
variants; purge every oxilangtag-sourced vector starting with `ab-abc-abc-abc-abc`
(`langtag_corpus.rs:162-164`, `langtag.rs:551`, `:569`); rebuild the corpus from Appendix A and the
ABNF only. Correct `PROVENANCE.md:46` and the PR body to state what is actually true.

**G2 — Make the parity claim falsifiable: frozen differential vectors**
In a throwaway crate outside the repository that still has `oxilangtag`, use it as an *oracle* only —
never copying its code or its test data. Inputs come from independent sources: a systematic ABNF
boundary sweep plus RFC 5646 Appendix A. Label each `(input, accept|reject)` by running the oracle,
freeze the table under `crates/iri/tests/`, and add a `PROVENANCE.md` row naming the oracle and the
input derivation. This survives the ban and turns the headline claim into something that can go red.

**G3 — Cover the load-bearing untested branch**
`langtag.rs:352-356` decides extlang admissibility via `if subtag.len() < 4`. No corpus vector
exercises a 4-alpha or 5–8-alpha primary language *followed by anything* — flipping `<` to `<=` keeps
every test green. Add, each half asserted: accept `abcd-Latn`; accept `AaBbCcDd-x-y-any-x`; refuse
`abcd-abc`; accept the neighbour `ab-abc`.

### HIGH

**G4 — One language-tag owner; retire the three divergent validators**
`rdfxml.rs:1029`, `text_parse.rs:1135`, and `term.rs:391` each hand-roll a *looser* grammar, so `@1`
is accepted by the RDF/XML and Turtle parsers and rejected by CSVW and embedding — silent widening in
the ingestion path. Give `purrdf_iri::langtag` a named profile enum (`Rfc5646` /
`Rfc5646PrivateUseRelaxed`, the latter documenting the GTS `x-purrdf-*` relaxation as one explicit
widening rather than three copies) and replace all three sites. Each swap tightens acceptance, so each
needs a paired refused/accepted vector proving the tightening is intended.

**G5 — Discharge SUBSUME/EXTEND/ENHANCE, with consumers**
The replacement is a strict subset of what it replaced. Restore and **wire** (unwired additions would
be dark, violating the module's own stated rule): §2.1.1 canonical-case normalization → wire to
`csvw/metadata.rs:1768`, which currently fakes it with `to_ascii_lowercase()`; singleton-keyed
`extension_subtags()`; `private_use_subtags()`; an owning form; `Hash`/`Ord`/`FromStr`/`AsRef`/
`Borrow`. Also thread the typed `LanguageTagError` through both consumers so `diagnostic_code()`
reaches a user instead of collapsing to `bool` at `target.rs:1233` and `metadata.rs:2107`.

**G6 — Single Tarjan owner** (D2)
Create `crates/rdf-core/src/graph.rs` with the one `tarjan_scc`; delete both copies; repoint
`purrdf-slice` and `purrdf-shex`; update the ban-list entry that currently enshrines two
implementations as the answer (`check-banned-deps.py:33`). Give the single owner a real battery:
chained SCCs, two disjoint non-trivial SCCs, a self-loop, an isolated node, the empty graph, a deep
path (~100k) proving no stack growth, and an assertion of the claimed ordering ("units sorted by
smallest member") that the current 3-node test never makes.

**G7 — Harden `scripts/cnschema-probe.py`** (covers both open review comments)
Add `--locked` at line 117 — an unlocked `cargo run` can rewrite the very `Cargo.lock` that this PR's
new gate treats as authoritative, which puts the PR's two new artifacts in direct tension. Resolve
`--file` (111) and `--purrdf` (115) before the `cwd=REPO_ROOT` switch. Fetch to a temp file and rename
only after the digest verifies, so an error-page response cannot poison the cache. Replace the
tautological self-test (writes 14 bytes, asserts sha ≠ constant, never calls `verify_digest`) with one
that runs the real failure path in a subprocess and asserts a non-zero exit, plus a truncated-file
case exercising the quad-count guard.

**G8 — Run and report `make conformance`**
`make check` excludes it, and the CSVW suite is exactly what would falsify *"no refusal decision
changes at either consumer."* Report the CSVW row against the baseline budget.

### MEDIUM

- **G9** — delete the orphaned half-sentence at `Cargo.toml:138`.
- **G10** — one first-party hex renderer reachable by all callers (`entail/explain.rs:223`,
  `datalog/cache.rs:264`, `entail/chase.rs:528` each hand-roll their own); pin one literal 64-hex
  golden for `bundle_content_id` so "no output change" is falsifiable — `crates/slice/` currently
  contains zero pinned digests, so all 112 tests would pass on a truncated or uppercased spelling.
- **G11** — the ban gate matches 7 literal names only; `const-hex`/`faster-hex`/`rustc-hex`/
  `hex-literal` or any rename re-enters silently. Add name-variant coverage and scan
  `[patch]`/`source`/`replace`.
- **G12** — `make cnschema-probe` runs nowhere; add a scheduled or release-lane job so a canonicalizer
  regression that moves `CANONICAL_SHA256` fails visibly instead of rotting.
- **G13** — `CHANGELOG.md` entry and an explicit bump decision for the new public `langtag` module.
- **G14** — zero doctests on `langtag` (every other `purrdf-iri` module has them: 20 total, none from
  this one); also clears the failing docstring-coverage check.
- **G15** — pair the three unpaired refusals: `a1`→`aa`, `en-a-b-cc`→`en-a-bb-b-cc`, `a-DE`→`ab-DE`.
- **G16** — proptest over generated subtag sequences (the crate already has `proptest` as a
  dev-dependency and a `tests/proptest.rs`).

### LOW

- **G17** — refetch once on a cache digest mismatch instead of dying and requiring manual deletion.
- **G18** — `quads` at `cnschema-probe.py:125` counts lines; rename to match.
- **G19** — `LinkUnit::is_cycle()` (`cache.rs:180`) is `members.len() > 1`, so a slice that depends on
  itself reports `false`. Pre-existing — petgraph behaved identically — but owned now.
- **G20** — remove process/deferral prose from shipped source (`langtag.rs:11-12, 24, 26-27`,
  `lib.rs:36`).

---

### Review-signal note

CodeRabbit could not clone the repository, so its review ran on the diff alone. Its two findings are a
floor rather than a clean bill — it missed G9, in the very file this PR is about. Both of its findings
are folded into G7.

### Verification before `/stage3`

`make check` (unpiped, so the exit code is make's own) · `make conformance` with the CSVW row reported
· `make wasm` · `make cnschema-probe` still PASSing with all three pinned constants after G7 ·
`check-banned-deps.py --self-test` and the real run both exit 0 · the G2 differential vectors green
**and demonstrably able to fail** (flip one expected verdict, observe red, revert) · the
completion-adversary re-run returning PASS · CI green on the PR, not merely locally.

paudley:
### Stage 2 progress — G9, G11 closed

**G9 — orphaned manifest comment** · `1e25be75`
Removing `oxilangtag` deleted two of its three comment lines in the workspace manifest. The first
survived, truncated mid-sentence, and came to sit above `pretty_assertions`, mis-describing it as an
"allocation-free RFC 5646 language-tag parser". No gate catches a stale TOML comment — it builds and
lints clean, which is why both `make check` and CI were green over it.

**G11 — banned-dependency gate: split by edge kind** · `f8b31b5d`

The audit's original suggestion was to widen the ban list with name variants (`const-hex`,
`faster-hex`, …). That was rejected: those are *different crates*, and pre-banning them is exactly the
over-refusal the contributor guide calls a first-class bug. The real defects were narrower.

1. **The `hex` ban was an over-refusal trap.** It was enforced across the entire resolved closure, so
   the day any upstream crate used `hex` internally, `make check` would fail with no fix available in
   this repository — over an edge that was never PurRDF's own. `hex` now fails only as a *direct*
   dependency of a workspace member, read from member manifests (`[dependencies]`,
   `[dev-dependencies]`, `[build-dependencies]`, and `[target.'cfg(…)'.*]`) plus root
   `[workspace.dependencies]` — not from `Cargo.lock` at all.
2. **The ox-family, `oxilangtag`, `petgraph` keep the any-edge ban.** Each has a first-party
   replacement good for every edge kind, so a transitive reappearance is still a failure.
3. **Substitution scanning added** — `[[patch.unused]]` blocks and `replace = "name:version"` lines
   are matched against the any-edge tier, so an override cannot smuggle a banned name back.
4. **The docstring now claims only what the code enforces.** It previously implied nothing could pull
   a banned crate back in. It now states the limit plainly: a fork published under a *different*
   package name is invisible to a name-based lock scan, and closing that needs a source-URL allowlist
   or provenance check, which this gate does not perform.

Verified independently of the implementing agent, exercising both directions:

| case | expected | observed |
|---|---|---|
| `--self-test` | exit 0 | exit 0 |
| real tree | exit 0 | exit 0 |
| `petgraph` anywhere in the lock | flagged | `['petgraph']` |
| `replace = "oxilangtag:0.1.6"` | flagged | `['oxilangtag']` |
| direct `hex` in a member manifest | exit 1 | exit 1 |
| **transitive-only `hex` in the lock** | **not flagged** | **`[]`** |

The last row is the over-refusal test and the point of the change: a valid tree must keep passing.

paudley:
## Stage 2 complete — 20 gaps closed, 6 adversary findings, 2 escalations, all fixed

22 commits. `make check` **EXIT=0** (unpiped). Both `.deficiencies` ledgers empty — nothing was
descoped, logged, or filed to a follow-up.

### What the audits found that green gates did not

All 18 CI checks were green on the original branch, and that proved very little. A
dependency-replacement PR destroys its own oracle: once `oxilangtag` left `Cargo.lock` and joined the
ban list, "acceptance language identical to the replaced crate" became unfalsifiable — no test in this
repository could have gone red if it were false.

**Three defects mattered most, and none of them could be caught by any gate that existed:**

1. **The provenance record contained a false sentence.** The parser was a transliteration of the MIT
   crate it replaced — the 26-entry grandfathered array matched in content and order, all seven error
   variants carried upstream's exact names, and the vector `ab-abc-abc-abc-abc` appeared in upstream's
   test suite and nowhere in RFC 5646 — while `PROVENANCE.md` stated it was written from the RFC. The
   remedy chosen was a genuine clean-room rewrite: recursive descent from the §2.1 ABNF, one function
   per production line, error taxonomy re-derived, grandfathered list re-derived in the RFC's own
   order. **It matched all 3935 frozen vectors on its first run**, before any test was touched.

2. **PurRDF could not read its own output** — and this branch introduced it. Giving the Turtle reader
   the `LANGTAG` terminal while the N-Triples and RDF/XML readers kept §2.1 meant `convert --from
   turtle --to nquads` emitted `"a"@fr-be-fbcl` and reading that same file back exited 1. Those tags
   are lawful `LANGTAG` terminals, they are in the vendored shexTest corpus, and `main` accepted them.
   Conformance stayed GREEN because no test in the repo crosses codecs. Asymmetric acceptance between
   a writer and its matching reader is always a round-trip bug.

3. **A second committed lockfile carried the banned crate.** `crates/geo/determinism/Cargo.lock` still
   resolved `oxilangtag 0.1.6`; the gate read only the root. Restoring that lock and running the real
   gate proves it was a genuine violation the root-only scan was blind to, not a synthetic test.

### The safety net is real, and was attacked

`crates/iri/tests/langtag_differential_vectors.txt` freezes 3935 inputs (743 accept / 3192 reject)
generated independently from the RFC and labelled once by the replaced crate run as an external oracle
outside this repository. No upstream source or test data was read or copied. The table carries a
SHA-256 of its own body that the test recomputes, so editing a verdict to make a change pass is caught
even when the vector count is preserved.

Mutation probes, run by the auditor rather than the implementer, at four separate sites:

| mutation | vectors that caught it |
|---|---|
| subtag ceiling 8 → 9 | 56 |
| variant minimum length 5 → 4 | 142 |
| `region = 2ALPHA` drops the ALPHA requirement | 102 |
| `is_script` 4ALPHA → 4-or-5ALPHA | 44 |

### Every gap, closed

| # | Gap | Resolution |
|---|---|---|
| G1 | Parser derived from the replaced crate under a false provenance record | Clean-room rewrite from the ABNF |
| G2 | Parity claim unfalsifiable | 3935-vector frozen differential table, SHA-256 self-checked |
| G3 | 4-alpha / 5–8-alpha language branch untested | Named boundary quartet; the coverage hole was already closed by G2 |
| G4 | Eight hand-rolled validators with different acceptance languages | All route through `purrdf_iri::langtag`; four named profiles |
| G5 | Replacement was a strict subset of what it replaced | §2.1.1 normalization wired to its CSVW consumer, singleton-keyed extensions, private-use iteration, owning form, trait set |
| G6 | Tarjan duplicated byte-for-byte in two crates | One owner in `purrdf-core` + 13-test battery incl. a 100k-node no-stack-growth proof |
| G7/17/18 | Probe: no `--locked`, unresolved paths, self-poisoning cache, tautological self-test | All four fixed; self-test now drives the real entry point |
| G8 | Conformance unreported | GREEN, 22 suites |
| G9 | Orphaned manifest comment | Removed |
| G10 | "No output change" had no pinned golden | Literal 64-hex golden, verified against `sha256sum` |
| G11 | Ban gate over-claimed; `hex` banned on any edge | Split by edge kind; docstring claims only what it enforces |
| G12 | Probe ran nowhere | Scheduled workflow |
| G13 | CHANGELOG | Generated by git-cliff from commit subjects — verified, no file change correct |
| G14/15/16 | No doctests, unpaired refusals, no generative coverage | 1 → 63 doctests; pairs added; 6 proptests at 512 cases |
| G19 | `is_cycle` blind to self-dependency | `is_cycle_in(edges)` added **without** widening `LinkUnit` |
| G20 | Process prose in shipped source | Replaced with scope statements |

### Adversary findings (pass 2), all discharged

F1 round-trip · F2 four more validators · F3 hex fold incomplete · F4 normalization reachability ·
F5 upstream test data survived the scrub · F6 second lockfile.

Discharged by demonstration, not assertion: an 8 tags × 5 formats × 5 formats matrix (200 conversions,
`WRITE-FAIL=0 READBACK-FAIL=0`); **271 W3C positives accepted, 88 negatives refused, zero leaks**; four
production surfaces each emitting the same diagnostic code with an accepted neighbour.

### Escalations (pass 3), both fixed rather than accepted

- **N1** — the direct-edge tier reproduced, one layer down, the blind spot the any-edge tier had just
  fixed: it iterated `workspace.members` only, so the two git-tracked-but-excluded first-party roots
  were invisible. The manifest set is now the union of two independent derivations, proven against a
  synthetic third excluded root, so the class is closed rather than the instance.
- **N2** — the body justified leaving four renderers by calling it "a workspace refactor". False: those
  crates already declare `purrdf-core`, making it a call-site substitution. The renderers were folded
  (fourteen total) rather than the sentence reworded. Two survivors are documented as genuinely
  different operations.

Plus one defect found by static analysis that no test could reach: the gate's git-discovery failure
handler named a renamed exception, so a gate that could not reach `git` would have raised `NameError`
instead of reporting why. Proven fixed by making `git` unavailable.

### Honest notes

- **Three ingestion paths are re-widened** to what `main` accepted (`e`, `a-DE`, `en-US-abc`,
  `de-419-DE`). Refusing them was this branch's regression. Genuine garbage — `1`, `-`, `9-9`,
  `123-456`, `en-` — is now refused on **every** path, which `main` did not do.
- **Canonical-case normalization is library API, not a CLI surface.** Its consumer is `read_csvw`; the
  `csvw-exact` lift profile refuses `@context.@language`, so it is not reachable from a shipped
  executable. The PR body says so.
- **Language tags are not byte-preserved across a round trip.** RDF 1.2 Concepts gives them a
  lower-case value space and the IR folds case at intern time, so `@en-US-abc` returns as
  `@en-us-abc`. The module's prose previously promised otherwise; it no longer does.
- **No breaking change**: no public item removed, no signature changed, no field added to an existing
  public struct, no version edited — verified mechanically against `origin/main`.

Ready for Stage 3.

paudley:
Trim third-party dependencies: first-party language tags, SCC, and hex; pinned cnSchema probe

Three third-party crates leave the workspace, each replaced by a first-party surface that subsumes
rather than narrows it, plus a gate that keeps them out and a reproducible interop probe.

WHAT SHIPS

oxilangtag -> purrdf_iri::langtag. An RFC 5646 Language-Tag parser derived from the section 2.1 ABNF
as recursive descent, one function per production line, behind four named profiles: section 2.1 as
written; that with the eight-character ceiling lifted on private-use subtags; the concrete syntaxes'
LANGTAG terminal bounded by the section 2.1 ceiling; and that terminal unbounded. The profiles are
totally ordered and the ordering is pinned by a test. Beyond well-formedness the module provides
section 2.1.1 canonical-case normalization, singleton-keyed extension iteration, private-use subtag
iteration, an owning form and the standard trait set. The boundary is well-formedness: subtags are
not checked against the IANA registry, and RFC 4647 matching is outside the crate's scope.

petgraph -> purrdf_core::graph::tarjan_scc. One iterative Tarjan with an explicit stack, in a crate
both consumers already reach. Stated precisely because the first pass got it wrong: main had exactly
one hand-rolled copy, in purrdf-shex, while purrdf-slice called petgraph and declared the dependency.
Dropping petgraph by adding a second hand-rolled copy traded a third-party dependency for a
maintained duplicate, and the ban-list entry then enshrined both as the replacement. The
consolidation removes a duplicate this change itself introduced.

hex -> core::fmt::LowerHex plus a first-party purrdf_core::hex::lower, whose test compares against an
independently written nibble lookup so implementation and oracle cannot share a bug. The description
carries no count of the copies folded, deliberately: an earlier revision said fourteen and was stale
one commit later, and correcting it to twelve was worse because no single counting rule made both
halves true. The durable form is the rule, and it lives in hex.rs where a future caller will read it.

check-banned-deps.py reads every git-tracked Cargo.lock and every git-tracked Cargo.toml, not just
the root and its members. The two first-party roots excluded from the workspace are exactly where a
banned crate had in fact re-entered unseen. The ban is split by edge kind: the ox-family, oxilangtag
and petgraph on any edge; hex only as a direct dependency of a first-party crate, because banning a
transitive third-party use would fail make check over an edge that was never ours, with no fix
available here.

check-corpus-frozen.py now guards the W3C syntax corpus and the two vendored SPARQL suites, which
hold every language-direction negative vector this change rests on. Scoped to the vendored
subdirectories: the suite directory also holds first-party suites that grow with feature work, and
freezing those would turn a loud, reviewable re-vendor into routine noise.

make cnschema-probe reproduces the pinned cnSchema 4.0 round-trip evidence -- source digest, the
canonical N-Quads digest on both legs, and the exact quad count -- fetch-by-digest, never vendored,
and now on a schedule so the pinned constants cannot rot silently.

THE PARITY CLAIM IS FALSIFIABLE

As first written, the headline claim could not fail: removing oxilangtag from Cargo.lock and banning
its return destroyed the only oracle that could contradict it. A frozen table now holds 3935 inputs
with accept/reject verdicts -- 743 accept, 3192 reject -- generated independently from the RFC and
labelled once by the replaced crate run as an external oracle outside this repository. No upstream
source or test data was read or copied. The table carries a SHA-256 of its own body that the test
recomputes, so editing a verdict to make a change pass is detected even when the vector count is
preserved. Mutation probes at four separate sites confirm it catches parser regressions rather than
only table edits: 56 vectors, 142, 102 and 44. The ABNF-derived parser matched all 3935 on its first
run, before any test was touched.

ONE OWNER FOR LANGUAGE TAGS

The first cut rewired only the two call sites that had imported the replaced crate. Every other
surface kept its own grammar: the RDF/XML parser took any alphanumeric subtags with no length cap,
the Turtle document parser validated nothing at all, and the SPARQL, ShEx, CDT and CSVW surfaces each
carried a transcription. Every site that decides well-formedness now calls purrdf_iri::langtag, and
none carries a character class or a shape rule of its own.

One owner is not one acceptance language. Three value spaces are in play, each with a specification
behind it: the codecs take the LANGTAG terminal with the section 2.1 ceiling, lifted after the
private-use marker; a CSVW lang property takes the RFC 5646 production, because CSVW Metadata section
5.7 requires BCP 47 and this workspace's own CSVW reader judges it there; an xsd:language cell value
takes XSD 1.1 section 3.3.3's flat pattern, which has no production structure and no private-use
carve-out, so its ceiling cannot be the profile's. Widening the CSVW property to the codec profile
would have been the obvious fix and the wrong one -- it would mint a lang this workspace drops on the
way back in. The visible consequence is asserted rather than left implicit, in a test naming all
three spaces and both specifications.

PURRDF COULD NOT READ ITS OWN OUTPUT

Unifying the readers exposed, and this branch briefly introduced, a round-trip defect: the Turtle
reader took the terminal while the N-Triples and RDF/XML readers took section 2.1, so converting
turtle to nquads emitted a tag that reading the same file back refused. Asymmetric acceptance between
a writer and its matching reader is always a round-trip bug. The codec paths now name one profile,
satisfying both fixed points -- the W3C negative vector stays refused, now reported as the length it
is rather than a production it never matched, while the long private-use families downstream projects
publish keep parsing. A codec-crossing test closes the hole that hid it.

That defect then recurred twice more, each time one layer further out, and each time found by audit
rather than by any gate. The CSVW projection accepted 1, 9-9 and 123-456 and wrote them into
artifacts. Four codec readers -- JSON-LD, YAML-LD, HexTuples and TriX -- never consulted the grammar
at all, so a JSON-LD document with a space in its language tag converted at exit 0 with an empty loss
ledger, emitting bytes no RDF parser can read. Finally the construction path: nothing checked the
grammar before a literal reached the dataset, so the same corrupt output was reachable from the
Python, wasm and C ABI surfaces.

The last of those is closed with one check at validate_components, the convergence rather than a
convenience -- the intern path, the Python constructor, the C ABI decoder, embedding encode and
pack-dictionary admission all already call it. The infallible constructors are deliberately left
ungated because their fields are public and a gate there is bypassable by construction. Three
boundary validators would have rebuilt the duplicated dialect this whole change exists to end.

SCOPE, STATED PLAINLY

Three ingestion paths are re-widened to what main accepted; genuine garbage is refused on every path,
which main did not do. Canonical-case normalization is library API consumed by the CSVW reader, not a
CLI surface. Three sites deliberately do not normalize: language-map keys, which must agree with a
lowercase fold at lookup; the JSON-LD value walker, which validates the caller's document rather than
rewriting it; and the embedding target, whose packed dictionary hard-errors on anything not already
lowercase. Language tags are not byte-preserved across a round trip -- RDF 1.2 Concepts gives them a
lower-case value space and the IR folds case at intern time. purrdf-gts keeps its own hex renderer
because it does not depend on purrdf-core and must not start.

Two remnants are disclosed rather than hidden. The Python constructor keeps a redundant empty-tag
pre-check with its own wording; it agrees with the core's verdict and fires a few lines earlier.
The wasm factory calls the bare constructor by design, which is safe only because a wasm term has no
standalone serializer -- if one is ever added it must route through the IR.

GOALS

No Cargo features added. No generated file or frozen vector hand-edited. No dependency version pinned
outside the root workspace table. No nondeterminism introduced. Every release crate stays
wasm32-unknown-unknown clean. No vocabulary IRI minted. Every refusal added or tightened carries a
neighbouring acceptance, and the one over-refusal found along the way -- a CSVW check demanding a
two-character primary subtag, refusing a, x and i-enochian against XSD's own pattern -- was removed.

NO BREAKING CHANGE

The semver guarantee covers Rust callers and is intact: no published signature changed, no public
item removed, no field added to an existing public struct, no version edited. The grammar check at
the IR kernel keeps validate_components' exact signature, carrying the code and message out through a
crate internal.

COMPLETION-ADVERSARY: PASS

Audited at acc0ff6b, the merge result, main having not advanced. Every finding across three stage-2
passes and three stage-3 passes is discharged on the production surface or in the contract text:
the provenance record, the unfalsifiable parity claim, eight then ten hand-rolled validators, the
subset replacement, the duplicated Tarjan, the probe's four defects, the unpinned hex golden, the
over-claiming ban gate, the orphaned manifest comment, a round-trip regression, upstream test data
surviving a clean-room scrub, a second lockfile carrying the banned crate, a gate blind spot one tier
down, seven CodeQL alerts, fifteen silently bumped third-party pins, two CSVW dialects, four
unguarded codec readers, and the construction path.

The load-bearing measurement: these commits tightened validation at the IR kernel and across four
previously unguarded readers -- the most over-refusal-prone change possible, since every literal in
every ingress now passes the grammar. The conformance matrix is unchanged: 22 suites, 4801 pass,
19 xfail against a budget of 19, 0 fail; syntax codecs 264 of 264; ShEx 1105 of 1105 with zero
trait-skips. A tightening that rejected real data would have shown up there.

DEFERRAL GATE: PASS

Zero hits across the diff and all 27 commit messages. Two body/comment hits adjudicated as false
positives with evidence: a review bot's own "Out of Scope Changes check skipped" tool output, and a
sentence that is a negation asserting the absence of deferral. Both .deficiencies ledgers carry only
their notice and marker -- nothing was descoped, logged, or filed away.

GATES

make check exits 0, run locally to completion. CI is green on this SHA: test, conformance, wasm,
workspace, capi, pytest, doc, msrv, projection-oracles, the book build, CodeQL with zero open alerts,
and every Analyze job.

NOTE FOR WHOEVER TOUCHES THIS NEXT

The recurring defect on this branch was never a wrong answer; it was a surface that was never asked.
Four times a validator set was verified exhaustively and the enumeration itself went unexamined -- and
a search that finds rival dialects cannot find absent validation, because there is no code to match
on. What ended it was choosing a point complete by construction rather than by enumeration. If you
add a codec, a projection or a binding, the question is not whether its grammar is right; it is
whether it asks.

PR #321, opened 2026-09-17
feat(319): implement purrdf-retrieval composition layer

paudley:
Implement the retrieval composition layer over ranked property-function producers as four
composable stages, with `search` **defined as** their composition rather than implemented a second
time to behave like it.

```
plan(request, registry, statistics) -> Plan
compile(plan, env)                  -> per-stratum SPARQL units
execute(units, registry, dataset)   -> per-stratum ranked streams
fuse(streams, profile, top_k)       -> one ordered answer

search = fuse ∘ execute ∘ compile ∘ plan
```

Every stage boundary is a supported place to stop and to start, and none is selected by a flag
asking the pipeline for less.

## What it does, demonstrated

`cargo run -p purrdf-retrieval --example fused_search` builds a real `TextIndex` and a real
`EmbeddingSpace` over a sealed PURREMB artifact, registers both through `register_ranked`, and runs
one `search`:

```
1. <…/doc-alpha>  score 0.032266458495  [lexical rank 3 (+0.015873015873), neighbour rank 1 (+0.016393442622)]
2. <…/doc-beta>   score 0.032258064516  [lexical rank 2 (+0.016129032258), neighbour rank 2 (+0.016129032258)]
3. <…/doc-delta>  score 0.032018442622  [lexical rank 1 (+0.016393442622), neighbour rank 4 (+0.015625000000)]
4. <…/doc-gamma>  score 0.015873015873  [neighbour rank 3 (+0.015873015873)]
```

Neither producer's order survives intact — BM25 ranks `delta` first, kNN ranks `alpha` first, and
the cross-stratum sum reorders both. `gamma` reaches the answer on one producer's evidence alone.
The trailer names each producer's own status, and the request term nothing served is reported as
such rather than silently dropped.

## Changes

- **Producer metadata on the registry, not the seam.** A producer's stratum, accepted request-term
  shapes with their argument placements, depth placement, candidate position, ordering guarantee,
  duplicate policy and mandatory flag are a `RankedDeclaration` supplied at
  `PropertyFunctionRegistry::register_ranked`. Nothing in `purrdf-core` changes, and the
  property-function trait gains no required method — a relation that does not fuse is simply
  registered without a declaration.
- **One stratum, one producer**, refused where the declaration is committed. A rank means something
  only inside the list that assigned it, so the refusal names both exits, and the third
  configuration that shares a scoring law and still needs separate strata.
- **`Plan` as pure data** with a canonical encoding, BLAKE3 `PlanId`, layout version, statistics
  snapshot, both registry identities, and per-term evidence for any term nothing served.
- **`FusionProfile` as an identity**, with validated construction, a derived contribution ceiling,
  two decay rules, and a documented coupling between weight, scale and admissible depth in both
  directions.
- **Fagin NRA fusion** over exact fixed point: summation across strata, score-descending order, a
  declared total tie-break, a bounded stop that reports where each stratum's evidence ends rather
  than draining to force an ending, and loud refusal on overflow.
- **Admission at the `compile` waist** across twelve typed dimensions, covering freshly planned,
  caller-edited and deserialized plans on the same terms.
- **Public `RegistryId` and a durable `content_fingerprint`** — the issue's one concrete API request.
- **Reachable three ways**: the crate, `purrdf::retrieval` on the umbrella facade, and
  `purrdf.retrieval` from Python.

## Verification

- `make check` exit 0; `make conformance` GREEN; the doc gate (separate from `make check`) exit 0
  for every touched crate; `check-doc-claims` and `check-issue-refs` exit 0.
- 246 tests in `purrdf-retrieval`, 818 in the Python bindings, plus the evaluator, facade and text
  suites.
- Report-only allocation bench for the fusion frontier — peak bytes and resident-set delta in its
  own process, asserting no threshold.
- `purrdf-core` untouched; no Cargo feature added; no dependency version pinned in a member crate.

## Not built, and stated rather than implied

Unfused enumeration is **materialized per stratum** and bounded by that stratum's own result, not
by the consumer's depth. The evaluator exposes no incremental result surface, so the design record's
§7 was corrected to describe what ships; making enumeration incremental is separate, larger work.

Complete *fused* enumeration remains excluded for its own reason: complete enumeration may not
discard, so it must retain, and retention is linear wherever it lives — emitted set, frontier or
spill. Fused top-k and unfused are the rungs this surface offers.

Geo is not a ranked producer. It computes a set with no score or rank, so it composes as a
constraint rather than as a stratum.


paudley:
# Plan — issue #319: Design: a composition layer over the ranked property-function producers

## Prior Art

**First recorded instance** — no `GhpRsq-Defect-Class` trailers exist in this repo; recurrence cannot be counted (UNKNOWN, not zero).

Related closed issues built the substrate but never attempted the composition layer itself:
- #185: property-function registry (the seam this layer sits over)
- #193: `purrdf-text` deterministic BM25 producer (a concrete ranked producer)
- #286: dataset pipeline bundles (composition precedent at a different layer)
- #179: SPARQL evaluation budgets (`EvalOptions` typed controls)

No retained lessons in Hindsight or fleet doctrine apply to this problem class.

Scope recommendation from historian: root-cause design — generalize the pipeline-bundle pattern downward to the property-function layer rather than point-fixing any single producer. **Acted on:** this plan designs the composition layer itself (four stages over the seam), not a fix to any producer; no recurrence count exists to force wider scope, so the historian's recommendation is satisfied by the plan as scoped.

## Standing Constraints (inviolable, from .goals)

- **GREENFIELD-FIRST / NO COMPROMISES** — SOTA design; learning from prior projects but not accepting their limitations.
- **RUST-FIRST, PYTHON-SURFACE** — all core functionality in high-quality Rust.
- **SUBSUME, EXTEND, ENHANCE** — if existing substrate is substandard, replace it.
- **LOW/NO OPTIONALITY, HARD FAILS** — fail fast and early; no optional components, no conditional imports, no swallowed errors.
- **RDF 1.2 ONLY** — no RDF 1.1 shortcuts, no quads; RDF* and RDF 1.2 are first-class.
- **MAXIMAL UTILITY / PERFORMANCE / PORTABILITY** — wasm-able core; byte-determinism; exact-arithmetic precedent from `purrdf-text`.
- **NO semantic Cargo features** — `scripts/check-no-features.py` gates this.
- **Kernel ring-fence** — `purrdf-core` and zero-dependency foundations (`purrdf-iri`, `purrdf-xsd`, `purrdf-events`) untouched.
- **No hardcoded vocabulary** — caller-supplied configuration, `example.org` in fixtures.

## Decisions Taken (from user approval)

1. **Placement:** One crate: `purrdf-retrieval` (`crates/retrieval`) as a sibling over the seam, co-locating all four stages; a single crate minimizes dependency overhead.
2. **Fixed-point parameters for reciprocal-rank fusion** — **amended at assembly on review evidence.** The approved parameters (scale `1_000_000`, round-half-up) are superseded: at scale 10⁶ with K=60, ranks 1000 and 1001 both round to 943 micro-units (1/1060 ≈ 0.000943396, 1/1061 ≈ 0.000942507), destroying strict monotonicity at depth ≥ 1000 and stalling threshold certification. Issue requirement 27 mandates "the exact-arithmetic precedent of `purrdf-text` is followed, not re-argued," and open question 2 assigns the concrete scale to "the existing exact-arithmetic records." The amended parameters are therefore the `purrdf-text` ones:
   - Type: `purrdf_text::fixed::Fixed` (public at `crates/text/src/lib.rs`), reused directly — `i128`, scale 10¹² (`SCALE_DIGITS = 12`).
   - Rounding: truncation toward zero (no tie-break rule, no cross-platform divergence).
   - Overflow: checked arithmetic returning `Err(FusionError::Overflow)` immediately; never wrap, never saturate silently; wide-intermediate multiply-divide so an intermediate overflow is not mistaken for an answer overflow.
3. **Scope boundary:** The issue is design-first and the record says it fixes structure, not a wire signature. However, the acceptance criteria extracted from testability assertions are implementation-heavy. This plan implements all acceptance criteria to a working state: design record, `RegistryId` visibility, producer metadata, `Plan` scaffolding, `fuse`, `plan`, `compile`/`execute`, composition identity, and boundary tests. A lighter design-only PR would leave the acceptance criteria unprovable; the chosen scope satisfies them.

### Design Refinements Integrated at Assembly (from eight-reviewer panel)

1. **Admission environment is explicit.** `compile(plan)` alone cannot evaluate admission: the registry's declared invariants and current statistics are not inside a serializable `Plan`, and `RegistryId` is a process-lifetime `AtomicU64` counter (`registry_id.rs`) with no cross-process meaning. `compile` takes `compile(plan, env: &AdmissionEnvironment)` where the environment carries the current registry and statistics snapshot. `Plan` records **two** registry identities: the durable content fingerprint (declaration digest, stable across processes for identical declarations) and the in-process `RegistryId` (distinguishes two structurally identical live registries, per the issue's API request). Admission matches `RegistryId` for same-process plans and the content fingerprint for deserialized plans.
2. **Producer metadata is declarative and serializable, never a function pointer.** `PropertyFunction` is a **trait** (`property_fn.rs:334`), so "additive fields" is realized as one additive required trait method returning an owned, serializable capability declaration over the registry's existing declaration/describe route: a closed enum `NotRanked | Ranked { stratum, accepted term patterns, rank-ordering guarantee, duplicate policy }`. No `fn(&Term) -> bool` anywhere (function pointers cannot serde and are not a stable description of accepted request language). Existing volatility/arity/access-pattern declarations are **consumed** through the registry's existing machinery (`describe()`, `property_fn_plan`, `QueryOptions` binding), never duplicated into parallel fields; the duplicate-IRI panic is preserved.
3. **The request is typed, not a flat string.** The issue's motivating producers consume different modalities — `purrdf-text` (lexical), `knn` (embedding vectors), `purrdf-geo` (WKT/GeoJSON geometries) — and requirement 25 speaks of "request terms." `plan` takes `&RetrievalRequest`, a closed term enum (`Lexical`, `Vector`, `Spatial`, `EntitySeed`); producers declare which term shapes they accept as serializable data, and matching is a lookup, not inference.
4. **Fusion is a verified stream protocol running Fagin NRA.** Each input stream is boundary-validated (1-based contiguous ranks, monotone non-increasing contributions, declared duplicate policy, terminal receipt) with typed protocol-violation errors. `FusionStream` maintains per-candidate lower/upper bounds and a threshold over stream heads, pulls greedily from the highest-contribution head, emits only certified rows (with a threshold witness), and prunes dominated candidates — memory proportional to the frontier. Total tie-break declared on the profile: fused score descending → best (minimum) stratum rank ascending → canonical term byte order ascending.
5. **Producer failure is isolated per stratum.** A single monolithic SPARQL query would let one producer's `EvalError` abort every stratum, making "two answered and one could not" unreachable (requirement 18). `compile` therefore emits per-stratum query units (each independently executable SPARQL text); `execute` runs each unit independently through `purrdf-sparql-eval` and maps a unit failure to that stratum's `ProducerStatus::ExecutionFailed` while the rest stream on.
6. **The fused answer is an evidence-carrying certificate.** Every fused row carries its per-stratum contributions (stratum, rank, score) and threshold witness; the trailer carries a typed `ProducerStatus` (`Exhausted | CeilingReached | ExecutionFailed | TermsRejected`) for every contributing **and** every applicable-but-failed producer, modeled on the `fallible.rs`/`governed.rs` partial-result taxonomy; the report names `(PlanId, FusionProfileId)`, with `PlanId` flowing plan → compiled bundle → streams → answer.
7. **Identities follow the `RankingProfile` discipline.** `Plan` and `FusionProfile` have validated construction, a versioned canonical length-framed encoding, and domain-separated BLAKE3 digests as their identities (`PlanId`, `FusionProfileId`); version mismatch on decode refuses loudly. Identity is derived from canonical bytes, never from `Hash` or serde bytes.

## Completeness Contract

| Requirement | Task | Acceptance Check |
|---|---|---|
| Ship `docs/design/purrdf-retrieval-ladder.md` with §1–§8 intact and open questions | T1 | File exists, renders, contains all design-record sections verbatim from issue body. |
| Make `RegistryId` public **and** expose a durable content fingerprint on `PropertyFunctionRegistry` | T2 | `RegistryId` (or `instance_id()`) is `pub`; `content_fingerprint()` is `pub`; tests: two identical live registries distinguishable by `RegistryId`; identical declarations yield identical fingerprints across instances; differing declarations yield differing fingerprints. |
| Producer capability metadata (stratum, accepted term patterns, ranked flag, ordering guarantee, duplicate policy) as serializable registry declarations; `Plan` as pure-data inspectable/serializable value with canonical encoding, BLAKE3 `PlanId`, version tag, statistics snapshot, both registry identities | T3 | Closed `NotRanked \| Ranked{..}` declaration type; zero function pointers (grep-gate); `Plan` is `Clone+Debug+PartialEq+Eq`; serde round-trip; canonical bytes stable; digest changes on any field change; unit tests for equality and snapshot recording. |
| `PropertyFunctionRegistry` otherwise unchanged — duplicate-IRI panic preserved; existing volatility/arity/access-pattern declarations consumed, not duplicated | T3, T5, T6 | Test: registering a second producer under the same IRI still panics; audit/unit test: planner and compiler read existing declarations, and the crate contains no parallel volatility/arity/access-pattern fields. |
| `FusionProfile` as identity (decay rule, weights, K, scale/rounding, tie-break, item comparator) with validated construction (K ≥ 1, weights > 0, admitted max-score ceiling) and BLAKE3 `FusionProfileId` | T4 | Profile derives identity from canonical bytes; invalid K/weight rejected at construction; two profiles differing in any field have different ids. |
| `fuse(streams, profile)`: fixed-point stratified RRF over `purrdf_text::fixed::Fixed`, Fagin NRA with greedy head-pull, checked additions, loud overflow, summation across strata, score-descending order, declared total tie-break | T4 | Tests: (1) one item in two strata — fused score equals checked sum of both contributions; (2) order is score descending; (3) equal-score pair ordered by the profile's declared tie-break; (4) overflow returns `Err(FusionError::Overflow)`; (5) golden byte-identical order; (6) differential test against an eager oracle on small streams. |
| Ranked-stream protocol with boundary-validated invariants and typed violations | T4 | Tests: out-of-order ranks, non-contiguous ranks, duplicate items under a no-dup policy, and error-after-rows each produce a typed protocol violation, never a plausible wrong order. |
| Preserve per-producer status at fusion; no aggregate flag; completeness asserted by trailer, not stream head; prefix reader holds incomplete evidence | T4 | Fused stream test: typed `ProducerStatus` for contributing + failed producers present in trailer; mid-stream status absent; early consumer sees incomplete evidence; drop/cancellation test proves all opened streams terminate with forge-proof final receipts. |
| Typed `RetrievalRequest` term lattice (`Lexical`/`Vector`/`Spatial`/`EntitySeed`); `plan(request, registry, statistics)` pure — no store/corpus/clock; golden-testable | T5 | Planner test with mock registry + mock statistics → asserted `Plan`; golden file diff captures planner changes; planner matches terms to producers by declared capability lookup only. |
| `compile(plan, env)` → per-stratum SPARQL query units; semantic admission at the narrow waist covering planned/edited/deserialized origins; refusal names exact violated dimension | T6 | Admission tests, all through the one `compile` entry: (1) fresh `plan(...)` output admitted; (2) caller-edited plan dropping an always-applicable producer refused; (3) serialize→deserialize round-trip admitted iff invariants hold; (4) tampered deserialized plan refused with the same variant as the edited case. Distinct variants for: (a) deleted always-applicable producer, (b) always-applicable producer present but bound to less than the registry requires, (c) depth bound violation, (d) weight referring to an undeclared stratum, (e) weight invalid under §5, (f) stale statistics, (g) registry identity mismatch. |
| Compiled units independently executable; producer failure isolated per stratum | T6 | Each emitted unit runs through `purrdf-sparql-eval` with no composition layer in path; a forced failure in one stratum yields `ProducerStatus::ExecutionFailed` for that stratum while the remaining strata stream to completion. |
| `execute` returns per-stratum ranked streams (delegates to `purrdf-sparql-eval`, unchanged); pinned-plan replay reproduces candidate set and per-stratum ranks | T6 | Streams tagged by stratum, ranks 1-based; replay test: same `Plan` + same registry → identical candidate set and per-stratum 1-based ranks from `execute(compile(plan, env))`; same-fingerprint/different-implementation registry is refused at admission. |
| `search` convenience entry point defined AS composition `fuse∘execute∘compile∘plan`; testable equality | T7 | `search(...)` result byte-equal to manual `fuse(execute(compile(plan(...), env)), profile)`. |
| Every stage boundary independently enterable/stoppable as supported seam; no mode flags | T8 | Six tests: stop at plan (inspect value); stop at compile (run SPARQL directly through evaluator); stop at execute (consume unfused streams unboundedly); start at compile (hand-built `Plan`, admitted on the same terms); start at execute (evaluator-only SPARQL, no composition types in path); start at fuse (hand-built streams). No `Option<skip_fuse>` or similar flags anywhere. |
| Fused enumeration is top-k; unfused is unbounded; no complete-fused-enumeration API | T8 | Memory-bounded top-k test for `fuse` (frontier proportional, verified by bench + allocation test); unfused stream test iterates arbitrarily deep. |
| Reporting a fused order names both pinned plan and fusion profile | T4 (implements), T8 (proves) | Answer/report type carries `(PlanId, FusionProfileId)`; `PlanId` flows plan → compiled bundle → streams → answer; test asserts both present. |
| Stale-statistics replay detectable at admission with declared behavior | T6 | Admission test: plan with mismatched statistics snapshot refused with `AdmissionError::StaleStatistics`. |
| Fused rows carry per-stratum contribution provenance | T4 | Test: fused row exposes `(stratum, rank, score)` for every contributing stratum; contributions sum to the fused score. |
| No new vocabulary; fixtures use `example.org`; no default registry or shipped producers | All | Audit: zero `blackcatinformatics.ca` namespaces; zero default producers; `example.org` in all test fixtures. |
| New crate builds for `wasm32-unknown-unknown`; no semantic feature | All | `cargo check --target wasm32-unknown-unknown -p purrdf-retrieval` passes; `scripts/check-no-features.py` passes; native-vs-wasm execution test proves identical canonical plan/profile bytes and fused output. |
| Byte-determinism for planner goldens and fusion order | T4–T5 | Golden files for planner output; fusion golden for cross-target order; diff-empty on re-run. |

- **Shiplift `purrdf-datalog` for admission now** — DECLINED. Admission uses direct checks against registry-declared invariants; the issue names Datalog as the substrate only if admission ever outgrows direct enumeration, which it does not today.
- **Ship a default registry with built-in producers** — DECLINED. Violates standing constraint (caller-supplied configuration, no fabricated default).
- **Express fusion as SPARQL extension function** — DECLINED by design (§5): embeds ranking policy in query language where conformance corpora cannot follow.
- **Implement complete fused enumeration** — DECLINED by design (§7): algebraically different from top-k; full enumeration redirected to the unfused rung.
- **Split into two crates (`fuse` primitive + planner)** — DECLINED per user approval. Single crate keeps stages co-located and minimizes dependency overhead.
- **Compile to a `purrdf-sparql-algebra` AST instead of SPARQL text** (enhancement-auditor-2) — DECLINED. Issue requirements 11–13 make SPARQL *text* the deliberate narrow waist: golden-reviewable, independently executable, and "nothing may live between planning and execution." An AST path between `compile` and `execute` would create a second, non-textual artifact in exactly the gap the design closes; the re-parse cost is accepted as the price of the seam contract.
- **Rank-aware relational pattern spec / conjunctive graph-pattern pushdown** (enhancement-auditor-2) — DECLINED for this PR. Anticipatory of hybrid structured/unstructured queries the issue does not ask for; `Plan` is a versioned, extensible value, so a `filter_pattern` field is additive later without breaking the four-stage contract. Recorded here so the extension point is deliberate.
- **Multiple rank-decay rules (Borda, logarithmic) now** (enhancement-auditor-2) — DECLINED as implementations; the decay rule is a named field of `FusionProfile` with one variant (`ReciprocalRank { k }`), so profile identity already covers the rule and further monotone rules are additive variants without re-opening the engine.
- **Separate explain/diagnostic projection surface** (enhancement-auditor) — DECLINED. The `Plan` value itself (selected/rejected producers with reasons, bindings, depths, digests) plus structured admission errors carry the essential diagnostics; a second projection surface is additive later and is not required by any acceptance criterion.
- **Non-optional `stratum`/`parse_predicate` with sentinel defaults** (plan-reviewer) — DECLINED. The closed `NotRanked | Ranked{..}` capability enum makes "does not participate" an explicit, serializable variant rather than a sentinel or a bare `None`, which is the stronger form of the same discipline.
- **Explicit sub-agent delegation markers and SOLID framing in task text** (plan-reviewer, plan-reviewer-2) — DECLINED. Delegation boundaries are the orchestrator's concern; the task structure already demonstrates the principles, and process annotations add no verifiable property to the artifact.

## Risks

| Risk | Mitigation |
|---|---|
| New required method on the `PropertyFunction` trait breaks existing in-repo implementors | Loud compile-time breakage is the correct behavior (§H); all in-repo implementors are updated in T3. No defaulted no-op method, which would silently exclude a producer from fusion. |
| NRA frontier memory growth under rank-skewed streams | Greedy highest-head pull + dominated-candidate pruning; allocation/peak-frontier criterion bench (report-only) added in T4; bounded-memory proof test in T8. |
| Depending on `purrdf-text` for `Fixed` pulls index machinery into `purrdf-retrieval` | `fixed.rs` is self-contained `i128` arithmetic; dependency direction is retrieval → text (no cycle, text never depends on retrieval). Dep closure reviewed at implementation; the kernel ring-fence is untouched. |
| Ephemeral `RegistryId` relied on across a process restart | Plan records both identities; admission matches content fingerprint for deserialized plans and `RegistryId` only in-process; mismatch is a named refusal, never a silent rebind. |

## Tasks

### Task 1: Ship design record `docs/design/purrdf-retrieval-ladder.md`
Write the file from the design record embedded in the issue body. Include the SPDX `CC-BY-4.0` header. Verify all sections (§1 Four stages, §2 Plan-as-value, §3 SPARQL compile target, §4 Seams, §5 Exact arithmetic fusion, §6 Status preservation, §7 Boundedness, §8 Pure planning, Open questions) are present.

**Acceptance:** `docs/design/purrdf-retrieval-ladder.md` exists, contains all sections, renders correctly in markdown.

### Task 2: Expose `RegistryId` and a durable content fingerprint publicly
In `crates/sparql-eval`: make `RegistryId` (or `PropertyFunctionRegistry::instance_id()`) `pub`. Add `pub fn content_fingerprint(&self) -> Result<String, EvalError>` on `PropertyFunctionRegistry`: an IRI-sorted digest over registered declarations (predicates, arities, volatilities, modes, and the new retrieval capability declarations from T3), built on the existing `property_fn_plan::registry_fingerprint` machinery but **excluding** the ephemeral instance id so the digest is stable across processes for identical declarations. The existing instance-inclusive fingerprint behavior used for prepared-plan pinning is unchanged.

**Acceptance:** `cargo test -p purrdf-sparql-eval` passes with new tests: two structurally identical live registries remain distinguishable via `RegistryId`; identical declarations produce identical content fingerprints across instances; differing declarations produce differing fingerprints. Both identities are visible outside the crate.

### Task 3: Crate scaffolding + producer capability declarations + `Plan` value
Create `crates/retrieval` (`purrdf-retrieval`, publishable): register the workspace member, keep dependency versions in `[workspace.dependencies]`, run `make metadata` clean, SPDX headers on all files.

In `crates/sparql-eval`: add one additive **required** method to the `PropertyFunction` trait returning an owned, serializable capability declaration routed through the registry's existing declaration/describe machinery (contained like every other declaration): a closed enum `RetrievalCapability = NotRanked | Ranked { stratum: Iri, accepted_terms: Vec<TermPattern>, ordering: RankOrdering, duplicates: DuplicatePolicy }`. `TermPattern` is declarative data (term kind, datatype/language constraints, predicate constraints) — no function pointers. Update all in-repo implementors. Preserve the duplicate-IRI panic and all existing declarations untouched.

In `crates/retrieval`: define `RetrievalRequest { terms: Vec<RequestTerm> }` with `RequestTerm = Lexical { text, language, predicate } | Vector { embedding, metric, index_hint } | Spatial { geometry, predicate, max_distance } | EntitySeed { entity }`. Define `Plan` as pure data: version tag, request terms, per-producer bindings, per-stratum depths, stratum weight map (as planned against), statistics snapshot, selected/rejected producers with reasons, and both registry identities (content fingerprint + `RegistryId`). Derive/implement `Clone`, `Debug`, `PartialEq`, `Eq`. Versioned canonical length-framed encoding; domain-separated BLAKE3 digest over the canonical bytes is `PlanId`; version mismatch on decode refuses loudly.

**Acceptance:** Unit tests for `Plan` equality, serialization round-trip, canonical-byte stability, digest sensitivity to any field change, and statistics snapshot recording pass. Grep-gate: zero `fn(` pointers in `purrdf-retrieval` and in the new declaration types. Duplicate-IRI panic test passes. `make metadata` diff-empty; `cargo check --target wasm32-unknown-unknown -p purrdf-retrieval` passes.

### Task 4: `fuse` primitive — verified-protocol NRA fusion over exact fixed point
Reuse `purrdf_text::fixed::Fixed` (`i128`, scale 10¹², truncation toward zero, checked ops, wide-intermediate mul-div). `ReciprocalRank` contribution: `weight * recip(K + rank)` computed with wide intermediates.

`FusionProfile` is an identity over all fields: decay rule (`ReciprocalRank { k }`, single variant), sorted weight map, declared scale/rounding constants, declared total tie-break (fused score descending → best stratum rank ascending → canonical term byte order ascending), and the item identity comparator/canonical item encoding. Validated construction following the `RankingProfile` discipline: `K ≥ 1`, every weight `> 0`, admitted maximum fused score computed from `max_contributions × max_weight × scale` and checked against the overflow ceiling; canonical length-framed description; BLAKE3 fingerprint is `FusionProfileId`.

`RankedStream` input protocol: boundary-validated invariants — 1-based contiguous ranks, monotone non-increasing contributions, one contribution per `(producer, item)` unless a declared duplicate policy says otherwise, and a terminal exhaustion/failure receipt. Violations return a typed `ProtocolViolation` error, never a plausible wrong order.

`FusionStream`: Fagin no-random-access top-k — per-candidate lower bound `L(x)` (summed seen contributions) and upper bound `U(x)` (plus unseen stream heads), threshold `T` over current heads, greedy pull from the highest-contribution head, emission only when `L(x*) ≥ max U(others)` and `L(x*) ≥ T` (equality resolved by the declared tie-break against every possible unseen identity), pruning of dominated candidates. Memory proportional to the frontier.

Fused rows carry: entity, fused score, per-stratum contributions `(stratum, rank, score)`, and the threshold witness. The trailer carries typed `ProducerStatus = Exhausted { rows_emitted } | CeilingReached { bound } | ExecutionFailed { reason } | TermsRejected` for every contributing **and** every applicable-but-failed producer, modeled on the `fallible.rs`/`governed.rs` taxonomy; a prefix reader holds evidence of incompleteness. The answer/report type carries `(PlanId, FusionProfileId)`; define how `PlanId` flows from `plan` through the compiled bundle and streams into the answer.

Add a criterion bench (report-only) for frontier allocation/peak memory.

**Acceptance:** Fusion tests: (1) one item in two strata — fused score equals the checked sum of both contributions at the declared scale; (2) order is score descending; (3) equal-score pair ordered by the declared tie-break; (4) overflow returns `Err(FusionError::Overflow)`; (5) golden byte-identical order; (6) differential test against a simple eager oracle over small streams and profiles; (7) protocol-violation tests (out-of-order/non-contiguous ranks, duplicate items, error after rows, never-ending source) each typed; (8) drop/cancellation test — all opened streams terminate and final receipts cannot be forged; (9) trailer/status tests; (10) contribution-provenance test — row contributions sum to the fused score; (11) profile construction tests (invalid K, non-positive weight, ceiling overflow, digest sensitivity); (12) native-vs-wasm execution equality for canonical profile bytes and fused output.

### Task 5: Pure planner
`plan(request: &RetrievalRequest, registry: &PropertyFunctionRegistry, statistics: &impl Statistics) -> Result<Plan, PlanError>`. Pure function — no store, I/O, or clock. Matches request terms to producers by lookup over their declared capability patterns (never inference); records selected/rejected producers with reasons, per-stratum depths (from producer cost declarations and statistics), the stratum weight map planned against, the statistics snapshot, and both registry identities. Golden-testable: text-in/text-out.

**Acceptance:** Planner goldens (request + mock registry + statistics → asserted plan) pass; re-running produces identical golden diff; planner consumes existing volatility/arity/access-pattern declarations (unit test proves no parallel re-declaration).

### Task 6: Compiler + admission + execute
`compile(plan: &Plan, env: &AdmissionEnvironment) -> Result<CompiledRetrieval, AdmissionError>` where `AdmissionEnvironment` carries the current `&PropertyFunctionRegistry` and current statistics snapshot. Semantic admission checks, each a distinct error variant naming the exact dimension: (a) `MissingMandatoryProducer` — always-applicable producer absent; (b) `InsufficientBindings` — always-applicable producer present but bound to less of the request than the registry declares it must receive; (c) `DepthBoundViolation`; (d) `UndeclaredStratumWeight` — weight keyed to an undeclared stratum; (e) `InvalidWeight` — weight invalid under §5; (f) `StaleStatistics` — statistics snapshot mismatch; (g) `RegistryMismatch` — `RegistryId` mismatch in-process, or content-fingerprint mismatch after deserialization. One `compile` entry covers all three plan origins (freshly planned, caller-edited, deserialized).

`CompiledRetrieval` carries per-stratum query units `(stratum, SPARQL text)` — each independently executable through `purrdf-sparql-eval` with no composition layer in the path — plus the `PlanId` and the registry identity `execute` requires, so callers cannot run the text against a different registry and obtain a different meaning silently.

`execute(compiled, registry)` runs each stratum unit independently through `purrdf-sparql-eval` (unchanged); a unit failure is captured as that stratum's `ProducerStatus::ExecutionFailed` while remaining strata stream to completion. Streams are tagged `(stratum, PlanId)`; ranks are 1-based within each stratum.

**Acceptance:** Admission tests through the one entry: fresh plan admitted; edited plan (dropped producer) refused; serialize→deserialize round-trip admitted iff invariants hold; tampered deserialized plan refused with the same variant as the edited case; each of the seven refusal dimensions returns its distinct variant. Compiled units run independently through the evaluator. Failure-isolation test: one stratum forced to fail, others complete, statuses distinguishable. Pinned-plan replay test: same `Plan` + same registry → identical candidate set and per-stratum ranks. Same-fingerprint/different-implementation registry refused. Stale-statistics test passes.

### Task 7: Search composition identity
`search(request, registry, statistics, profile)` calls `fuse(execute(compile(plan(request, registry, statistics), env)), profile)`. Never re-implements the pipeline.

**Acceptance:** `search(...)` result byte-equal to manual composition call.

### Task 8: Stage boundary and streaming tests
Prove each seam is independently usable: stop at plan (inspect value); stop at compile (run SPARQL directly through the evaluator); stop at execute (consume unfused streams unboundedly); start at compile (hand-built `Plan`, admitted on the same terms as planner-origin); start at execute (evaluator-only SPARQL, no composition types in path); start at fuse (hand-built streams). Prove fused is top-k with bounded memory; unfused is unbounded. Prove trailer semantics (prefix reader holds incomplete evidence). Prove reporting names both `PlanId` and `FusionProfileId`.

**Acceptance:** All six seam tests pass; bounded-memory and unbounded-stream tests pass; no mode flags exist anywhere in the new crate.

### Task 9: PR creation
Post plan to issue; create PR from branch; post plan to PR body.

**Acceptance:** PR created, linked to #319, plan posted in comments.

paudley:
# Stage 2 — Issue #319 / PR #321 gap remediation

**Issue:** Design: a composition layer over the ranked property-function producers
**Branch:** `paudley/319-design-a-composition-layer-over-the` @ `9b1509aa` · **PR #321** · 61 files, +9901/−36

---

## Context

PurRDF ranks results three ways today — full-text BM25 (`purrdf-text`), vector kNN
(`purrdf-sparql-eval`), and geospatial (`purrdf-geo`). Each returns its own ranked list.
Issue #319 asked for a layer that takes **one** request, reaches whichever of those
producers apply, and merges their separate rankings into one ordered answer — expressed as
four composable stages, `search = fuse ∘ execute ∘ compile ∘ plan`, where every stage
boundary is a supported place to both stop and start.

PR #321 built that pipeline. The fusion engine (`fusion_stream.rs`, `reciprocal_rank.rs`,
`fixed.rs`) and the registry-identity work are production-grade. **But the pipeline is
inert**: it discards the request before emitting SPARQL, runs against a hardcoded empty
dataset, ignores the depth it planned, and every real producer declares itself out of
ranked retrieval. All 75 tests pass because they drive a mock whose `open()` ignores its
arguments and returns a canned list.

Three independent audits (gap-analyzer, completion-adversary, context review) converged on
this, and each load-bearing claim was verified directly:

| Claim | Verification |
|---|---|
| Real producers opt out | `text/src/relation.rs:580`, `knn/mod.rs:701`, `geo/src/relation.rs:1339` → `NotRanked`, each with "later task"/"scaffold" prose |
| Request dropped | `grep -c request_terms crates/retrieval/src/compile.rs` → **0** |
| Empty dataset | `execute.rs:198` `RdfDatasetBuilder::new().freeze()`; no dataset parameter on `execute`/`search` |
| Depth ignored | `grep LIMIT crates/retrieval/src/compile.rs` → **0**; `stratum_depths` read only as `.keys()` at `compile.rs:102` |
| Deserialized plans refused | `admission.rs:321` instance check is unconditional, contradicting `plan.rs:177-180` |
| Root cause of the drop | `TermPattern` (`property_fn.rs:333`) has no argument-position field — the compiler *cannot* place a term |

**CI is red**: the `workspace` job fails on `scripts/check-issue-refs.py`, and `test` and
`wasm` never ran.

**Outcome intended:** the headline runs end-to-end — a real request against a real dataset
reaches real ranked producers and returns one fused, bounded, provenance-carrying answer,
demonstrated on a production surface rather than a mock.

---

## Decisions taken (stated, not asked)

| # | Decision | Why |
|---|---|---|
| **D1** | Add an **argument position** to `TermPattern`, so a producer declares *where* each accepted term binds. | Root cause of the dropped request. Without it `compile` structurally cannot emit the term. |
| **D2** | Move the capability declaration off `PropertyFunction` into a separate **`RankedProducer`** trait with a registry side table. | The issue states "Nothing in `purrdf-core` **or the property-function seam** changes to admit it." Also deletes 25 `NotRanked` stubs forced onto unrelated implementors (path relations, governor, SHACL, conformance shims) and fixes the ISP violation. `lib.rs:53-59` currently contradicts itself on this in shipped rustdoc. |
| **D3** | Thread a **dataset** through `execute`/`search`. | `geof:` is dataset-dependent and is one of the three named producers. The module doc's claim that "relations are row sources, so stored data is irrelevant" is false. |
| **D4** | Make `mandatory` an **explicit declared field**; delete `always_applicable` inference. | The issue: "admission enforces whatever the registry declared and **adds nothing of its own**." Current code derives mandatory-ness from an unconstrained pattern — over-refusal with no valid-neighbour test. |
| **D5** | Register the new crate's files in `AMBIGUOUS_PLAN_PHRASES` rather than rephrasing. | `scripts/check-issue-refs.py:389` is the gate's *designed* register for files where "the plan's" means a runtime plan value. `crates/sparql-eval/src/property_fn.rs` is already in it. The register can only shrink (staleness-checked at line 1656), so it cannot rot. |
| **D6** | `RequestTerm` gains `Temporal` and `NumericRange` now, enum stays closed. | The maintainer's own stated first preference (issue comment #2); free before publication, semver-major after. |
| **D7** | Wire into the `purrdf` umbrella facade **and** the Python surface, plus a runnable example. | `.goals`: "RUST-FIRST, PYTHON-SURFACE". The crate currently has zero dependents anywhere. |

---

## Work items (each = one commit, verified before the next)

### Phase A — unblock the gates

**W1 · CI `workspace` job is red.** 11 `check-issue-refs.py` violations on `"the plan's"` in
`crates/retrieval/src` (`admission.rs:401`, `canonical.rs:4,7`, `compile.rs:53,88,89`,
`lib.rs:28,73`, `plan.rs:155,193`, `planner.rs:63`). All are `Plan` the domain type — a true
false positive. Add the six files to `AMBIGUOUS_PLAN_PHRASES` (D5). *Every later hygiene gate
was skipped by this failure* — brand casing, kernel ring-fence, release coherence, crates.io
publish order — so the new crate's release wiring has **never been gate-verified**.

### Phase B — make the pipeline actually run (the headline)

**W2 · Producers declare where their terms bind.** Add an argument-position field to
`TermPattern` and a `mandatory` field to `RetrievalCapability::Ranked` (D1, D4). Delete
`admission.rs:234-239 always_applicable` and the derived-mandatory path at `:352-358`.

**W3 · Move the declaration to `RankedProducer`** (D2). New trait + registry side table
(`register_ranked` / `ranked_declaration`); revert the 25 forced `NotRanked` stubs; fix the
self-contradicting rustdoc at `lib.rs:53-59`.

**W4 · `compile` emits the request.** `emit_unit` (`compile.rs:132-175`) binds each producer's
declared input positions from `binding.request_terms`, selecting a **declared mode** from
`descriptor.modes` consistent with those terms — otherwise `TextSearchRelation` aborts on a
free needle (`text/src/relation.rs:590`) even once terms flow. Emit `LIMIT <depth>` per stratum
from `stratum_depths`. Refuse at planning when no declared mode is satisfiable.
*Falsifying test:* two different requests must compile to **different** SPARQL, and a request
with a depth of 3 must yield exactly 3 rows.

**W5 · `execute` takes a dataset** (D3). Thread `&dyn DatasetView` through `execute` and
`search`; replace the hardcoded empty graph at `execute.rs:198`.

**W6 · Real producers declare `Ranked`.** `TextSearchRelation`, `EmbeddingKnnRelation`,
`GeoRelation` take caller-supplied stratum / accepted-terms / ordering / duplicates
configuration in their constructors and return a real declaration. Delete all three
"later task" comments. *Falsifying test:* an end-to-end fusion of BM25 + kNN + geo rows over a
non-empty graph.

**W7 · Candidate identity.** `term_candidate` (`execute.rs:277-290`) returns hex of canonical
bytes as the candidate name, and `FusedRow.entity` carries it with no inverse. Carry `TermValue`
through to `FusedRow`. Also: the `filter_map(... find_map(Option::as_ref))` silently takes the
first bound variable in a row and silently drops all-unbound rows — make both explicit.

### Phase C — correctness defects

**W8 · Deserialized plans are always refused.** `admission.rs:321` checks the process-local
`RegistryId` unconditionally, so a plan serialized and reloaded in a new process can never be
admitted — destroying the plan-as-portable-value centrepiece of §2/§8, and contradicting
`plan.rs:177-180` which documents fingerprint matching for exactly this case. Textbook
over-refusal: both existing "deserialized" tests reuse the *same live registry*, so **no
valid-neighbour case exists**. Branch on origin; add a test that rebuilds the registry and
admits the plan, **and** one proving a genuinely different registry is still refused.

**W9 · Producer silently omitted.** Admission never checks that every `producer_bindings`
stratum has a `stratum_depths` entry; `compile.rs:102` iterates depth keys only, so an edited
or deserialized plan can drop a producer and still compile to a well-formed query — the exact
silent-drop §2 exists to prevent. The `continue` comment at `compile.rs:104` asserts an
invariant admission does not enforce. Add the inverse check.

**W10 · Fusion ceiling unenforced.** `max_contributions` and `ceiling` are validated at
construction and hashed into the profile identity, but `FusionStream::pull`
(`fusion_stream.rs:500-519`) references neither — `checked_add` catches only fixed-point
overflow. Enforce both with a dedicated error variant naming the exceeded limit; test the
refusal **and** an at-the-ceiling valid case.

**W11 · Equal plans, different `PlanId`s.** `RequestTerm` derives `PartialEq` over `f32`
embeddings while `canonical.rs:58` writes `f32::to_bits()`, so `0.0 == -0.0` yet the bytes
differ; `NaN != NaN` also breaks `Eq`'s reflexivity. The doc at `request.rs:33-37` *claims*
hand-written bit-pattern equality that `:38`'s derive and `:75`'s bare `impl Eq` do not provide.
Hand-write `PartialEq` on `to_bits()`.

**W12 · `search` destroys producer status.** `search.rs:160` discards
`ExecutionResult::statuses`; `SearchResult` has no status field, so "two answered and one could
not" is indistinguishable from "two answered". §6 binds this obligation at `fuse`; the module
doc relocates it to `execute`, which a seam does not license. Merge statuses into the trailer.
(`TermsRejected` and `CeilingReached` are currently unreachable — they become reachable once
W4 and W10 land.)

**W13 · Bounded fusion.** `fuse.rs:93-97` drains to a `Vec` and its doc says "the **complete**
ordered answer", but §7 says complete fused enumeration "is not offered". Add an explicit `k`
to `fuse`/`search`. Also export the `execute → fuse` bridge (`RankedStreamAdapter`,
`search.rs:192`, currently private and absent from `lib.rs`) so the "stop at execute, fuse
yourself" seam is reachable without re-deriving the contribution arithmetic — and have the
composition-identity test use the exported bridge rather than a copy.

**W14 · Over-refusals with no valid-neighbour test.** `declared_row_bound → unwrap_or(0)`
(`admission.rs:250`) makes any positive depth fail for a producer declaring no modes;
`UnknownStratum` (`search.rs:160`, `fuse.rs:86`) fails the *entire* request when a profile omits
one stratum's weight. Per repo doctrine, each needs both an invalid case and a neighbouring
valid case that still succeeds.

### Phase D — the maintainer's two review demands (issue comment #2, both open)

**W15 · Monotone-depth coupling.** The fixed-point constraint couples weight, scale and depth:
`r ≲ sqrt(w·S) − K`. Below that depth, ordering silently degrades from rank order to the
canonical-byte tie-break. `FusionProfile` validates only `weight > 0`; admission variant (c)
checks registry bounds, not profile arithmetic. Document it on `FusionProfile`, add admission
**variant (h)** refusing a depth beyond the stratum's monotone range, and add the
two-adjacent-ranks-collide fixture. Note: `reciprocal_rank.rs:52-56` truncates *before*
weighting, so the real bound differs from the reviewer's model — derive and document the actual
one.

**W16 · `RequestTerm` lattice.** Add `Temporal` and `NumericRange` (D6), and record the
closed-lattice commitment in the design record's open questions.

### Phase E — artifacts, surfaces, and honest tests

**W17 · Design record.** Both "open questions" still read as open though this PR answers both
(single crate; scale 10¹², truncation, overflow bound). Record the resolutions.

**W18 · README.** `crates/retrieval/README.md:12,16` tells crates.io readers that
compile/execute/fuse arrive "once the later stages land" and that the crate "currently ships the
plan value itself". Rewrite for all four stages plus `search`.

**W19 · Wire the crate outward** (D7): `purrdf` umbrella facade (which today exports the
capability *type* but not the layer that consumes it), Python bindings, and a runnable
`examples/` end-to-end search. Update `AGENTS.md:53` (crate map), `AGENTS.md:93` and `:232`
("23 crates" → 24), `README.md:685` (crate table), `README.md:715` (design-doc index).

**W20 · Replace the tests that prove nothing.**
- `native_wasm_execution_equality` (`boundary.rs:857`) runs the native path twice and asserts it
  equals itself; no wasm target is involved. Commit `9b1509aa` claims otherwise.
- `no_mode_flags_in_api` (`boundary.rs:772`) is an 8-literal `include_str!` blocklist.
- `fused_top_k_bounded_memory` (`boundary.rs:608`) uses a single stratum with
  `max_contributions = 1` — the degenerate case with no cross-stratum frontier.
- The `fn(` function-pointer grep-gate named in the acceptance criteria was never added.
- No frontier-allocation criterion bench, though the plan, the risks table and the contract each
  name one. Report-only, asserting no speedup (repo rule).

**W21 · Docstring coverage** 50.10% vs the 80% threshold across 477 functions. Correct the PR
body's "57 total tests" (actual: 75 + 7).

### Phase F — smaller defects

`#![deny(clippy::float_arithmetic)]` absent despite the exact-arithmetic mandate while carrying
`Vec<f32>` and `f64` selectivity · `purrdf-retrieval` depends on all of `purrdf-text` solely for
`Fixed` · `Statistics::selectivity` recorded but never influences planning · `.expect()` panic on
a library hot path (`search.rs:220`) · `RECIP_K = 60` magic default re-exported from the crate
root · dead `ExecutionError::EvalError` variant · `with_tie_break` speculative API on a
single-variant enum.

---

## Verification

Per work item, before its commit: `cargo test -p <crate>` foreground/serial (this machine kills
background shells under memory pressure).

Exit gate for the stage — all must hold:

1. **`make check` green** (fmt + clippy `-D warnings` + build + tests + hygiene), run unpiped —
   a piped `make` reports `tail`'s exit code, not the gate's.
2. **`make wasm` green** — `cargo check --target wasm32-unknown-unknown -p purrdf-retrieval`.
3. **CI green on PR #321**, including the `test` and `wasm` jobs that have never run, and the
   release-coherence and ring-fence gates that W1 unblocks. Local green ≠ CI green.
4. **The headline demonstrated on a production surface, not a mock:** the runnable example
   performs one search over a non-empty dataset, reaches real BM25 + kNN + geo producers, and
   returns one fused ranking carrying per-stratum provenance and every producer's status.
5. **Two different requests produce different SPARQL** — the single check that falsifies the
   whole inert-pipeline failure.
6. **Completion-adversary re-run returns PASS** with an executed production-surface
   demonstration per criterion. Its findings are binding; the stage does not close while one is
   open.

`.deficiencies` is currently clean and is expected to stay clean — no gap here is a log entry.

paudley:
### Gap W1 — CI `workspace` job red (CRITICAL) — FIXED in `3d13688e`

**Gap.** `scripts/check-issue-refs.py` failed with 11 violations, all on the phrase
`"the plan's"` in `crates/retrieval/src` (`admission.rs:401`, `canonical.rs:4,7`,
`compile.rs:53,88,89`, `lib.rs:28,73`, `plan.rs:155,193`, `planner.rs:63`).

Every hit is `Plan` the **domain type** — "the plan's canonical identity", "the plan's
fields", "the plan's binding order", "the plan's own request" — not a planning document.
A true false positive of the gate's heuristic.

**Why this mattered beyond the red X.** The failure aborted the `workspace` job at step 16,
so every *subsequent* hygiene gate was skipped: brand casing, the kernel ring-fence
(no oxigraph/PyO3 in `purrdf-core`), scraped-conformance-row checking, cross-registry
release coherence, and the crates.io publish-order check. The new 24th publishable crate's
release wiring had therefore **never actually been gate-verified**, only asserted.

**Fix.** Registered the six source files in the gate's own `AMBIGUOUS_PLAN_PHRASES` frozenset
(`scripts/check-issue-refs.py:389`) and widened that register's doc comment to name the
retrieval plan value alongside the query, compaction and execution ones.

This is the mechanism's designed use, not a suppression: the register exists precisely because
English cannot distinguish "the plan's soundness certificate" (a design document) from "the
plan's pre-order" (a runtime value), so the ban stays absolute and the code-sense sites are
named. `crates/sparql-eval/src/property_fn.rs` was already registered for the same reason.
The register is self-policing — the script fails on any entry whose file no longer contains
the phrase, so it can only shrink.

**Verification.** `python3 scripts/check-issue-refs.py` → exit 0:
`no #NNN issue-reference tokens and no new process references (… 14 runtime-plan "the plan's" site(s) registered)` — up from 8.

paudley:
### Gap W10 — the fusion profile's declared bounds were never enforced (HIGH) — FIXED in `b174a782`

Raised by CodeRabbit as Major/Data Integrity; unaddressed until now.

**Gap.** `FusionProfile` validates `max_contributions`, computes a score `ceiling` at
construction (`fusion_profile.rs:327 compute_ceiling`), folds both into the profile's
identity, and exposes both publicly (`:182`, `:188`). But `FusionStream::pull` consulted
**neither** — `grep -n "ceiling\|max_contributions" crates/retrieval/src/fusion_stream.rs`
returned zero hits. It ran a `checked_add` and nothing else.

Consequences: a candidate appearing in enough streams silently exceeded the bound the profile
declared, and `FusionProfile::new`'s refusal of `max_contributions == 0` was protecting a limit
no code checked. `checked_add` catches only fixed-point overflow, a far larger and different bound.

**Fix.** Both bounds are now enforced in `pull` through the profile's own accessors, with typed
variants that **name the exceeded limit** rather than reusing `Overflow` (as the review asked):
`FusionError::MaxContributionsExceeded { item, count, max }` and
`FusionError::CeilingExceeded { item, score, ceiling }`.

**On the ceiling's reachability — stated rather than implied.** Under the sole
`DecayRule::ReciprocalRank`, `FusionProfile::new` requires `K >= 1` and `contribution()`
requires `rank >= 1`, so every contribution is `weight * recip(K + rank)` with `K + rank >= 2`,
i.e. at most `weight/2`. Summed over at most `max_contributions` contributions, an admitted
candidate tops out at `ceiling/2` — a factor of two below the bound, always. So
`CeilingExceeded` is **provably unreachable today**. It is retained as a defensive invariant
against a future decay rule with a looser reciprocal bound, and the tests record that reasoning
rather than implying a live refusal.

**Over-refusal discipline.** Per the repository rule that a refusal is a claim requiring proof,
each new refusal ships with its neighbouring valid case:

| Test | Asserts |
|---|---|
| `contribution_count_exceeding_max_is_refused` | 3 contributions under `max_contributions: 2` → `MaxContributionsExceeded { count: 3, max: 2 }` |
| `contribution_count_at_max_succeeds` | **valid neighbour** — exactly 2 under `max_contributions: 2` still fuses, and the fused score still equals the checked sum |
| `maximal_legitimate_score_stays_below_the_ceiling` | the tightest legitimate profile reaches exactly `ceiling()/2`, by exact arithmetic |
| `ceiling_boundary_is_gated_by_the_contribution_count_check` | the one configuration whose raw arithmetic lands on `ceiling()` is caught by the count check first |

**One pre-existing test was passing only because of this bug.** `checked_addition_overflow_is_refused`
reached genuine `i128` overflow by letting a single candidate silently exceed `max_contributions` —
the exact defect being fixed. Closing the hole removed that path, so the test was rewritten to use
three *distinct* candidates, taking its overflow from `compute_threshold`'s cross-stream sum instead.

**Verification.** `cargo clippy -p purrdf-retrieval --all-targets -- -D warnings` → exit 0.
`cargo test -p purrdf-retrieval` → 79 passed, 0 failed (fusion suite 14 → 18).

paudley:
### Gap W11 — equal request terms could produce different `PlanId`s (CRITICAL) — FIXED in `ac82c38f`

Raised by CodeRabbit as Minor/Data Integrity. It is not minor: it breaks the plan-identity
contract the whole crate rests on.

**Gap.** `RequestTerm`'s doc comment claimed:

> `Eq` is implemented by hand because the vector arm carries `f32`: two embeddings are equal iff
> their bit patterns are equal, which is the identity a plan's canonical encoding uses.

The code did no such thing. It was `#[derive(..., PartialEq, ...)]` plus a bare
`impl Eq for RequestTerm {}`. The doc described an intention; the implementation was the default.

Two failures, in opposite directions:

1. **Equal terms, different identities.** Derived `PartialEq` on `Vec<f32>` uses IEEE equality, so
   `0.0f32 == -0.0f32`. But `canonical.rs:58` encodes via `f32::to_bits()`, where those differ. Two
   `RequestTerm`s comparing **equal** therefore produced **different canonical bytes and different
   `PlanId`s** — the exact inverse of the contract ("digest changes on any field change" requires
   that equal plans carry equal digests).
2. **A reflexivity that did not hold.** `NaN != NaN` under IEEE, so the bare `impl Eq` asserted an
   equivalence relation the type did not satisfy. The doc waved this off with "callers must not
   place a NaN in an embedding" — but `Plan` is public, `Deserialize`, and explicitly treated as
   untrusted input by `admission.rs`, so a NaN is reachable, not merely discouraged.

**Fix.** Hand-written `impl PartialEq for RequestTerm`, comparing embeddings element-wise by
`f32::to_bits()`. Equal bits now mean equal terms, so `-0.0` is distinct from `0.0` (matching the
encoder) and a `NaN` equals itself (making `Eq` genuine for every value, including deserialized
ones). Comparing bits rather than floats also keeps the arm clear of `clippy::float_cmp` with no
`#[allow]`. The other three arms compare exactly as the derive did.

The doc comment is rewritten to describe the implementation instead of an intention, and the
"callers must not place a NaN" caveat is gone — bit-pattern comparison handles NaN correctly
rather than tolerating it.

No `Hash` impl needed reconciling: nothing in the crate derives or implements `Hash` over
`RequestTerm`, `RetrievalRequest` or `Plan`.

**Tests** — both directions plus the ordinary case, per the repository's over-refusal rule that a
tightened equality must be shown not to have broken the valid neighbour:

| Test | Asserts |
|---|---|
| `signed_zero_embeddings_are_distinct` | `0.0` vs `-0.0` embeddings are unequal **and** their plans have different `PlanId`s |
| `nan_embeddings_are_reflexive` | a `NaN` embedding equals its clone, and the plans share a `PlanId` |
| `identical_embeddings_are_equal` | **valid neighbour** — independently built identical embeddings are equal and yield the *same* `PlanId` |
| `equal_lexical_plans_have_equal_canonical_bytes` | the non-vector arms were not disturbed |

**Verification.** `cargo clippy -p purrdf-retrieval --all-targets -- -D warnings` → exit 0.
`cargo test -p purrdf-retrieval` → 83 passed, 0 failed (plan suite 9 → 13).

paudley:
### Phase B progress — the seam refactor (steps 1–3 of 11)

Context for the commits landing on this branch. The audit found the pipeline **inert**: `compile`
discards the request, `execute` runs on a hardcoded empty graph, the planned per-stratum depth is
never applied, and all three named producers declare `NotRanked`. Those four defects mutually mask
each other, which is why 75 tests passed over them. Phase B rebuilds the connective tissue.

A design decision was taken first, and it is load-bearing for everything that follows.

#### The declaration moves off the seam

`PropertyFunction::retrieval_capability()` was a **required** trait method, so every property
function in the workspace had to name the retrieval vocabulary in order to say it wanted nothing
to do with it — path relations, governor charge points, SHACL engines, the exists-admission gate,
the parallel gates, the conformance shims. That is an interface-segregation failure, and it
contradicts the issue's own statement that *"Nothing in `purrdf-core` or the property-function seam
changes to admit it."* The crate doc asserted both halves of the contradiction in consecutive
sentences.

It is now a **registry side table**: `register_ranked(iri, relation, declaration)` /
`ranked_declaration(iri)`. This is also what "producers are caller-supplied configuration" means
operationally — the declaration arrives where the caller commits it, not baked into the relation.
A welcome consequence: `TextSearchRelation`, `EmbeddingKnnRelation` and `GeoRelation` need **no
constructor change**.

Verified safe to remove rather than deprecate: `git show rust-v2.0.2:...` shows zero references to
`retrieval_capability` or `RetrievalCapability`, and `purrdf-retrieval` does not exist at that tag.
Both symbols are unreleased. This is a pre-release correction, not a breaking change.

#### Two corrections the implementation forced

**A single argument position per accepted term is not enough.** A lexical search with a language
filter must bind the needle at flattened position 1 *and* the language at position 4
(`crates/text/src/relation.rs:74`, applied at `:632-636`). Leaving position 4 free is not a
refusal — the relation then answers over **every** language for a request that named one. So the
declaration maps *facets* of a term (value / language / predicate / max-distance) to positions.

**Depth is sometimes an argument, not a `LIMIT`.** `EmbeddingKnnRelation` declares exactly one
mode, `"fbbf"` (`knn/mod.rs:123`) — position 2 is `k`, which *is* the per-stratum depth. Left free,
the invocation is not merely unbounded, it is **refused at prepare time** by `admitted_row_bound`.
Hence a separate `depth_placement`, and both mechanisms ship.

#### Commits so far

| Commit | What |
|---|---|
| `dd62ad02` | Side table added, purely additive. Positions validated where the declaration is committed: out-of-arity candidate/placement/depth, a depth colliding with a term placement, or a candidate colliding with either, all panic naming the offending value. 17 tests. |
| `77333e86` | Every reader switched: fingerprint, planner, admission. Both planner goldens regenerated. |

`dd62ad02` gave the trait method a `NotRanked` default and a deprecation note — scaffolding, so the
~29 overrides could be retired in reviewable batches rather than one commit spanning every
implementor. An empirical finding made that viable: **implementing a `#[deprecated]` trait method
does not trip the lint; only calling one does.** So all existing overrides stayed warning-free under
`-D warnings` and exactly one internal call site needed a scoped allow.

#### On the fingerprint move in `77333e86`

The registry fingerprint changed, deliberately — a declaration now carries a candidate position, a
mandatory flag, a depth discriminant and per-pattern placements. That value is part of the
plan-cache key and of governed receipts, so caches go cold and receipts spanning the change are not
comparable. Correct rather than incidental: the declared shape genuinely moved.

Both planner goldens were regenerated, and **only `registry_content_fingerprint` moved in either
file**. `producer_bindings`, `producer_decisions`, `stratum_depths`, `stratum_weights`,
`request_terms`, `statistics_snapshot` and `version` are byte-identical — that is the evidence that
switching readers changed no planning outcome. A relation registered without a declaration still
contributes exactly one byte (`n`), verified by decoding the golden, so it cannot perturb a ranked
relation's digest.

#### A test that would have gone quietly vacuous

`capability_participates_in_the_content_fingerprint` varied the capability via the *trait method*
and asserted two fingerprints differ. Once the fingerprint read the side table instead, both arms
would have contributed the same byte — comparing two identical digests. It now varies the
declaration (the thing the digest notices) while holding the trait-level capability **constant
across both arms**, so the old path demonstrably cannot account for the difference.

Worth stating as a general hazard: when you change what a value is computed *from*, every test
asserting on that value silently changes meaning, and the dangerous ones do not go red — they go
green for a new and wrong reason.

Steps 4–11 follow: retire the ~29 stubs and the enum, add placement/rendering, make `compile` emit
the request with per-stratum bounds, thread a real dataset, make `mandatory` declared rather than
inferred, and wire the real producers with an end-to-end fusion test.

paudley:
### Phase B complete — the pipeline is no longer inert

The audit found four mutually-masking defects: `compile` discarded the request, `execute` ran
against a hardcoded empty graph, the planned per-stratum depth was never applied, and all three
named producers declared `NotRanked`. Each removed the observation that would have caught the next,
which is how 75 tests passed over them.

All four are closed, and the layer now **fuses two real producers over real data**:

| row | candidate | provenance (stratum, rank) |
|---|---|---|
| 1 | `<https://example.org/a>` | (lexical, 1), (neighbour, 1) |
| 2 | `<https://example.org/b>` | (lexical, 2), (neighbour, 2) |
| 3 | `<https://example.org/d>` | (neighbour, 3) |
| 4 | `<https://example.org/c>` | (neighbour, 4) |

Rows 3 and 4 are the load-bearing part: `c` and `d` are near `a` in the embedding space but hold
none of the needle's terms, so kNN reaches them and BM25 does not. The producers genuinely
disagree and fusion resolves it — a fixture where both strata returned the same rows would pass
just as well with one producer silently dropped.

The emitted SPARQL, which is where "the request reaches the query" becomes visible:

```
{ SELECT (?c0 AS ?candidate) WHERE { ( ?c0 ) <…/pf/search>     ( "alpha beta" ?c2 ?c3 ?c4 ?c5 ) } LIMIT 4 }
{ SELECT (?c0 AS ?candidate) WHERE { ( ?c0 ) <…/pf/neighbours> ( <…/a> "4"^^<…#integer> ?c3 ) } }
```

Text carries the stratum's `LIMIT`; kNN carries none, because the stratum depth **is** its `k`,
bound as argument 2. Leaving that argument free would not merely leave the invocation unbounded —
it would be refused at prepare time. That asymmetry was unobservable while everything was a mock.

#### Commits

| Commit | What |
|---|---|
| `dd62ad02` | Ranked declarations move to a registry side table, purely additive |
| `77333e86` | Every reader switched; registry fingerprint moves deliberately |
| `d5795c77` | Seam method and enum removed — 25 stubs deleted, **+70/−303** |
| `571204c5` | `compile` emits the request, with per-stratum bounds |
| `06af5280` | Quoted-triple seeds classified as triples, not IRIs |
| `689349a1` | `execute` runs against the caller's dataset; candidates become legible |
| `749239f5` | Real text + kNN producers registered and fused end to end |

#### Decisions worth recording

**The declaration left the seam.** `retrieval_capability()` was a *required* trait method, so every
property function in the workspace named the retrieval vocabulary in order to decline it — path
relations, governor charge points, SHACL engines, conformance shims. That contradicted the issue's
own "nothing in the property-function seam changes to admit it", and the crate doc asserted both
halves of the contradiction in consecutive sentences. Verified removable rather than deprecable:
the published `rust-v2.0.2` tag contains zero references to either symbol.

**Mandatory is declared, not inferred.** The old `always_applicable` derived mandatory-ness from a
pattern being unconstrained, which made a permissive producer silently un-droppable — a refusal the
layer invented, against a contract that says admission "adds nothing of its own". A catch-all
producer declared non-mandatory can now be bound to a subset and admitted; declared mandatory, the
same edit is still refused. Both halves pinned.

**RDF 1.2 renders rather than being declined.** Quoted triples emit in their value form and
directional literals carry their base direction, both pinned by round-tripping the emitted query
back through the evaluator. The parser accepted both all along (`parser.rs:3346`, `:5438`), so
refusing them would have been over-refusal against an existing fixture.

**Geo is described accurately.** `GeoRelation` computes a set — `pairs.sort(); pairs.dedup();` —
carries neither score nor rank, and both its slots take features rather than geometry literals. It
is not made ranked and no stratum is invented for it; it composes as a constraint. The issue's
opening counted it as a third ranked producer; that was colour, not a requirement.

#### Bugs found and fixed while proving the above

- **`seed_kind` classified quoted triples as IRIs** — it tested `starts_with('<')` first, and
  `<<( s p o )>>` also opens with `<`. So `TermKind::Triple` matched *nothing* (the `pf/quoted`
  producer in both goldens was unreachable) while an IRI-accepting producer would be handed a term
  it cannot read. Rendering was correct throughout because the decoder discriminated correctly —
  only *selection* was wrong. The two are now pinned against each other.
- **A blank-node candidate failed its entire stratum.** Refusing a blank node in an *argument*
  position is right (it is a non-distinguished variable — free, not ground). Reusing that rule to
  *name a result* discarded every valid row beside it, and the decoder already read `_:label` back.
  Candidates are named; the refusal stays where it bites, at placement, where a blank cannot seed a
  later request.
- **A fingerprint test was about to go vacuous** — it varied capability via the trait method, so
  once the digest read the side table both arms would have contributed the same byte. It now varies
  the declaration while holding the trait-level value constant, so the old path cannot account for
  the difference.

136 tests in `purrdf-retrieval` (was 75), 169 in `purrdf-text`, 1367 in `purrdf-sparql-eval`;
clippy `-D warnings` clean across all three; both planner goldens byte-identical except the one
deliberate fingerprint move.

paudley:
### Remediation record — Phase C onward

Continuing from the Phase B summary. `make check` is green; CI `workspace`, `capi`, `doc`, `msrv`
and `projection-oracles` pass. 220 tests in `purrdf-retrieval` (was 75), 818 in the Python bindings.

#### Correctness defects closed

**A deserialized plan could never be admitted.** The live-instance check compared a process-local
counter unconditionally, so a plan serialized and reloaded anywhere else was refused however
faithfully it matched — destroying the portability the plan-as-a-value design rests on. `plan.rs`
documented the correct behaviour; it was never implemented. Neither existing round-trip test could
see it, because both reused the registry that planned them.

Origin is now recorded at the decode paths rather than inferred. Inferring it from the identity was
the tempting reading and is unsound: a foreign counter and a stale one are the same number, so it
would have deleted the mismatch refusal in exactly the case that justifies it. Origin stays out of
the canonical bytes and out of equality, so a plan still equals its own round trip and no identity
moves. The fingerprint is still checked for every plan; origin selects only whether the live
instance must *also* match, and claiming same-process is the stricter direction.

**A producer bound to a stratum with no depth entry was silently omitted** — the compiler groups by
stratum but iterates the depth keys, so the query came out well-formed and answered less than the
registry promised. The compiler carried a comment asserting admission had already prevented it.

**Two over-refusals with no valid neighbour.** A producer declaring no access mode declared no row
count, and the absent count was read as a bound of zero, so any positive depth was refused. And a
fusion profile weighting *some* of a plan's strata failed the entire request — discarding every
weighted stratum to report one unweighted one. The profile is deliberately not a planning input, so
the two lists can legitimately disagree; saying nothing about a stratum is not saying the request is
malformed.

**Exactly-tied candidates stalled the fusion.** A candidate was held while any rival's upper bound
reached its score, so two candidates with exactly equal final scores each held the other back
forever. Reciprocal-rank fusion over symmetric strata produces that routinely — ranks (1,2) and
(2,1) sum identically. A measured case pulled 20,000 rows of 20,000 where 6 suffice. Every other
test in the suite passed against that mutation, including the bounded-frontier test, whose fixture
avoids ties by construction.

**Finality was arithmetic where it must be structural.** `U == L` was meant to say "no stream that
could still name this candidate is open", but a live head contributing zero adds nothing to the sum
and is indistinguishable from an exhausted one. Reachable, not theoretical: the smallest
representable weight contributes exactly zero at every rank. Reverting the fix reproduced the double
emission — `["a","z","a"]`, the second copy carrying score 0.

**A multi-producer stratum ranked its entire corpus.** A union reports no per-child row ceiling,
correctly; the descent treated that absence as final and carried it to the leaves, so each branch's
own `LIMIT` was never consulted. Live at any scale, invisible to single-producer fixtures.

#### The maintainer's two review demands

**The monotone-depth coupling** was derived from the code rather than adopted from the review — and
the reviewed formula is wrong above weight 1. The reciprocal is truncated *before* weighting, so for
any weight ≥ 1 the weight drops out of the comparison entirely and every such profile shares one
bound just past √scale. The closed form predicts ~10⁹ at w=10⁶ where the measured bound is 1,001,000.
Below 1 the closed form is sound but not tight, so shipping it would have refused legitimate depths
by 10–30%. The bound is computed, not approximated.

**`Temporal` and `NumericRange`** added to the request lattice, kept closed, with exact endpoints —
never binary floats, because a request term is part of a plan's identity.

#### The crate had no callers

It was absent from `[workspace.dependencies]` entirely, so no member *could* depend on it. Now wired
into the umbrella facade, a Python surface driving the real async entry point under a parking
executor (no shadow sync API), and a runnable example fusing BM25 against kNN where the two
genuinely disagree.

#### Tests that asserted less than they claimed

The cross-target test ran the native path twice and compared it to itself; it now runs under the
repo's wasm execution lane with hand-computed expectations. The no-mode-flags check grepped six
files for eight literals; it now pins each stage's signature so a selector fails to compile. The
bounded-memory test used one stratum — no cross-stratum frontier to bound — and counted pulls rather
than bytes.

Measured through `fuse` after the bounded stop: three strata of 1,000 rows and three strata of
1,000,000 cost the **same peak to the byte**, at 99 pulls rather than 3,000,000.

#### Allocation discipline

Two gratuitous per-row allocations removed (a stratum name cloned per row to serve a borrow; a
candidate key cloned on every pull including the majority that hit an existing candidate), and term
lexicals now build into a sized buffer rather than growing from empty one character at a time. What
stays is documented as inherent rather than left to look accidental.

#### Depth

A fused depth beyond ~10⁶ was refused by arithmetic before a row was read, for every weight.
Dividing the weight by the rank directly — one rounding instead of two — is strictly more accurate
and makes precision scale with the weight, so the monotone range grows as √w. Shipped as a second
decay rule so no existing profile's identity or numbers move; the identity test spells the expected
encoding by hand and carries a digest computed outside the code under test.

paudley:
feat(retrieval): compose ranked property-function producers into one answer

A four-stage composition over ranked producers — `plan → compile → execute → fuse` — with
`search` **defined as** their composition rather than implemented a second time to behave like
it. Every stage boundary is a supported place to stop and to start, and none is selected by a
flag asking the pipeline for less.

## What it does

`cargo run -p purrdf-retrieval --example fused_search` builds a real `TextIndex` and a real
`EmbeddingSpace` over a sealed PURREMB artifact, registers both through `register_ranked`, and
runs one `search`:

    1. <…/doc-alpha>  0.032266458495  [lexical rank 3 (+0.015873015873), neighbour rank 1 (+0.016393442622)]
    2. <…/doc-beta>   0.032258064516  [lexical rank 2 (+0.016129032258), neighbour rank 2 (+0.016129032258)]
    3. <…/doc-delta>  0.032018442622  [lexical rank 1 (+0.016393442622), neighbour rank 4 (+0.015625000000)]
    4. <…/doc-gamma>  0.015873015873  [neighbour rank 3 (+0.015873015873)]

Neither producer's order survives: BM25 ranks `delta` first, nearest-neighbour ranks `alpha`
first, and the cross-stratum sum reorders both. `gamma` reaches the answer on one producer's
evidence alone. The trailer names each producer's own status, and a request term nothing served
is reported as such rather than dropped.

## Surface

New crate `purrdf-retrieval` (22 source files, 15 test files, 249 tests). Reachable three ways:
the crate directly, `purrdf::retrieval` on the umbrella facade, and `purrdf.retrieval` from
Python.

- **Producer metadata lives on the registry, not the seam.** A producer's stratum, accepted
  request-term shapes with their argument placements, depth placement, candidate position,
  ordering guarantee, duplicate policy and mandatory flag form a `RankedDeclaration` supplied at
  `PropertyFunctionRegistry::register_ranked`. `purrdf-core` is untouched and the
  property-function trait gains no required method — a relation that does not fuse is simply
  registered without a declaration.
- **One stratum, one producer**, refused where the declaration is committed. A rank means
  something only inside the list that assigned it, so the refusal names both exits — merge inside
  one producer that owns the comparability, or give each its own stratum — and the third
  configuration that shares a scoring law and still needs separate strata, because a differential
  weight over a bounded score can act only in rank space.
- **`Plan` as pure data**: canonical encoding, BLAKE3 `PlanId`, layout version, statistics
  snapshot, both registry identities, per-term evidence for anything nothing served.
- **`FusionProfile` as an identity**: validated construction, derived contribution ceiling, two
  decay rules, and the weight/scale/depth coupling documented in both directions.
- **Fagin NRA fusion** over exact fixed point: summation across strata, score-descending order, a
  declared total tie-break, a bounded stop that reports where each stratum's evidence ends rather
  than draining to force an ending, loud refusal on overflow.
- **Admission at the `compile` narrow waist** across thirteen typed dimensions, covering freshly
  planned, caller-edited and deserialized plans on the same terms.
- **Public `RegistryId` and a durable `content_fingerprint`** — the issue's one concrete API
  request.

## Requirements satisfied

The design record ships at `docs/design/purrdf-retrieval-ladder.md`, §1–§8 plus its open
questions, with the questions the implementation answers now recorded as answered.

Every acceptance criterion was demonstrated on the production surface — by running the example,
the real-producer suite, and the emitted SPARQL through `purrdf-sparql-eval` directly — not by
unit test or code reading.

## Standing constraints respected

No Cargo feature added anywhere. No dependency version pinned in a member crate. `purrdf-core`
untouched; no oxigraph or PyO3 near it. Every release crate stays `wasm32-unknown-unknown`-clean
(CI `wasm` green). No vocabulary minted: no default registry, no built-in producer, no fabricated
datatype — an untyped embedding is refused rather than given one — and `example.org` throughout.
Serializers stay deterministic; exact fixed-point arithmetic only, with `float_arithmetic` denied
at the crate root.

## Completion-adversary: PASS

Six audit passes ran. The final pass returned PASS against the merge-into-main result with every
prior finding discharged by execution:

- **Hollow binding at the admission waist** — a plan could bind a term to a producer that places
  none of it, compile to a well-formed query carrying no trace of that term, and report the term
  served. Now refused as `HollowBinding`, by the planner's own carriage function rather than a
  restatement of it, at the one chokepoint every plan passes. One definition, three callers, so
  the two sides cannot drift. The neighbouring valid case is demonstrated by differential: one
  index separates admission from refusal on the same plan.
- **Contradictory enumeration claim** — the design record advertised stopping at the per-producer
  rung "to enumerate a million rows" while another section of the same document said a consumer
  that reads one row has already paid for all of them. The whole claim family is swept; the rung
  is now described as what it is.
- **Gates observed on the merged head**, not a predecessor: all 18 checks pass on `a3d33e09`,
  including `test`, `wasm` and `conformance`.

Earlier passes closed: two red CI gates shipped past (a private intra-doc link; a stale
conformance count in both a generated table and its restated prose), declarations read by nothing
(`ordering`/`duplicates`, now honoured — `Allowed` de-duplicates where it had been refused), a
request modality that could not be served with content, and a plan field that decided nothing
while contributing to plan identity.

## Deferral gate: passed

Nine scan hits in the diff, every one adjudicated a false positive in context — the banned word
appears inside a negation ("rather than a placeholder", "no stub, no sentinel", "not implemented
a second time", "added now rather than deferred") or as domain vocabulary ("the seed of a
follow-up request", meaning a fused row re-seeding a subsequent search, which is an executed
test). PR body: zero hits. `.deficiencies`: clean in both locations, notice and marker only.

## Out of scope, by the operator's ruling

A search that **returns** 14 million hits is not in scope for this work. Asked directly whether
"handle 14M triples in a search / where the search returns 14M hits" meant admissibility or an
executed 14M-hit answer, the operator ruled verbatim: **"It's not in scope."**

What this branch does deliver is the arithmetic that had made deep strata impossible. The decay
rule truncated the reciprocal before applying the weight, so adjacent ranks stopped differing past
the square root of the scale and every profile shared one monotone range near a million ranks
whatever it weighed — a weight of a thousand bought exactly as much depth as a weight of one. A
second decay rule divides the weight by the rank in one exactly-rounded division, which is
strictly more accurate and makes the range grow as the square root of the weight: a weight of two
hundred is strictly ordered past fourteen million ranks. A 14M-deep stratum is therefore
admissible and correctly ordered, and the old rule refuses it by name.

Enumeration itself remains materialized per stratum, and the design record says so plainly rather
than implying otherwise.

## Implementation notes

- **A rank is meaningful only inside the list that assigned it.** That single fact decides the
  one-stratum-one-producer invariant, the refusal's wording, and why a differential weight over a
  bounded metric must act in rank space.
- **Declarations are load-bearing or they are decoration.** Several fields were declared,
  encoded, folded into the registry fingerprint and read by nothing; flipping them changed no
  behaviour. Each now has a reader, and honouring the uniqueness declaration is what removes the
  per-stream identity set — the one structure that grew with rows pulled rather than rows kept.
- **The waist judges untrusted input.** Any rule the planner applies must be re-derived there, by
  the same function, or it holds only for plans the planner wrote.
- **A test that compares emitted text cannot see a hollow binding**, because a hollow binding
  emits identical text. The seam test now compares bindings and per-term evidence too.
- Test binaries and benches carry no debug info: a failing assertion prints its own file and line,
  and the workspace links well over a hundred test binaries.

## Conflicts

None. Merging 39 commits of main was clean, and every branch-owned file was verified line-for-line
unchanged by the merge.

PR #324, opened 2026-09-17
Ask the grammar in STRLANG and the SPARQL results readers

paudley:
## Summary

A query string was enough to make PurRDF emit bytes no parser reads.

`STRLANG("x", "en us")` bound a literal that serialized as `"x"@en us` in TSV and
`"xml:lang":"en us"` in JSON and XML — from the CLI, the C ABI, wasm and the Rust API alike.
`STRLANGDIR` had the same hole.

## Why the existing gates did not catch it

The recent dependency-trimming work gave language tags one owner and gated the IR kernel at
`RdfLiteral::validate_components`, so nothing enters an `RdfDataset` ungrammatical. It then claimed
every construction path reaching a serializer was covered. **That claim is false for SPARQL results,
because that path has no dataset in it.** `STRLANG`'s literal goes into a scratch arena and straight
out to the results writer.

The query **parser** does validate a tag written in query text, on this exact profile. `STRLANG` is a
*runtime* constructor, so it walked around a gate that exists and works rather than past a missing one.

The diagnostic that exposed it: **`CONSTRUCT` refused the same tag `SELECT` printed** — same engine,
same tag, opposite verdicts — because a `CONSTRUCT` template routes through the dataset builder.

A second path of the same class: the results **readers** (`from_json`, `from_xml`) read `xml:lang`
from an external document and validated nothing, which is how federated `SERVICE` results arrive. A
remote endpoint's tag propagated into locally written results unexamined.

## The fix

Both constructors and both readers now ask `purrdf_iri::langtag` on
`Profile::ConcreteSyntaxLangtagBounded` — the profile the parser, the codecs and the IR kernel already
name. No dependency was added; both crates already declared the edge.

**The failure modes differ deliberately, and each follows its own precedent.** `STRLANG` leaves the
expression **unbound**: SPARQL 1.1 §17.4.2.5 makes an ill-formed argument an error, §17.2 turns an
expression error into an unbound result rather than an aborted query, and it is what these functions
already did for an empty tag. The readers **refuse hard**, because they are parsers rather than
evaluators — an unreadable `xml:lang` is a parse error for the same reason the unreadable base
direction handled beside it already was.

The gate runs **before** the lowercase fold so a refusal can quote what was typed. That is safe rather
than lucky, and a test proves the verdict is case-blind across the whole corpus rather than leaving it
to argument.

## Anti-regression

`str_lang_agrees_with_the_profile_the_query_parser_already_enforces` asserts, over every corpus tag,
that `STRLANG` binds **iff** the query parser accepts the same tag written as a literal. If either
profile is ever changed without the other, that test reddens instead of the asymmetry silently
reopening.

## The escapes of this class that are closed

Five earlier revisions of this section made a claim about what remained, and each was falsified — by
miscounting a list of open paths, by asserting the class was closed when it was not, by naming the
wrong type and the wrong caller for an open surface, and by crediting this change with creating a door
that already existed. **This revision does not claim completeness.** Each escape below was found and
closed; the list is what has been found, not what exists. Most were found by audit *after* this
description already said the job was done, and the last two are fallout from fixing the others:

1. `STRLANG` / `STRLANGDIR` — runtime constructors that walked around the parser's gate.
2. The SPARQL results readers (`from_json`, `from_xml`) — the federated `SERVICE` path.
3. `STRDT` with `rdf:langString` — minted a literal with no tag at all, which every other door refused.
4. **GTS decode.** The tag came straight off the wire. The gate is in the term-decode loop, not at the
   event stream: that receives a `Term` that already exists, and the crate decodes a language tag in
   exactly one place — so one gate covers every `GtsEventSink` and the snapshot path that re-enters the
   same loop. It does **not** cover a graph nobody read off the wire; see escape 8.
5. **Composite-datatype lifting.** `cdt`/`list` functions lift members out of a literal's *lexical
   form*, so `"[\"x\"@en us]"^^cdt:List` carried a tag the kernel never admitted. Nobody had named
   this one; it surfaced because the gate went at the convergence rather than at the known seams.
6. **The Python fold view built from parts**, which bypasses `reader::read` entirely and so never met
   the decode gate.
7. **A SPARQL pre-binding.** `SparqlRequest` substitutions and SHACL pre-binding build a literal from a
   caller-supplied tag string straight into the scratch arena. The algebra validator does refuse a
   hand-built `Query` carrying an ungrammatical tag, at admission — but substitution rewrites a plan
   that has *already been admitted*, so that gate cannot see these. Gated at the ingress rather than
   mid-pipeline, because a refusal deeper in would have deleted a constraint rather than raised one.
8. **The Rust fold view.** `GtsFoldView::with_config` accepted a caller-assembled `Graph` without asking
   the grammar. `Graph.terms` and `Term::lang` are both `pub`, so building one field by field reached
   every renderer without meeting the reader's decode gate — the exact twin of escape 6, gated in
   Python and open in Rust. Left standing, the language a fold view renders would have depended on
   which binding assembled the graph. The constructor already returned `Result` and already refused a
   caller-assembled graph it could not walk, so this is one more arm on an existing refusal and no
   caller's signature moves.

The four SPARQL extension seams — custom functions, a remote `ServiceResolver`, property functions, and
`CustomAggregate` (which has no type constraint at all) — converge on the scratch interner, which owns
the crate's only term push. Gating that convergence is what exposed escape 5.

Escapes 7 and 8 are both worth naming for the same reason: neither was found by looking for ungated
seams. Escape 7 surfaced from repairing a regression this change itself introduced at that seam, and
escape 8 from an auditor asking why one binding's door was gated and its twin's was not. A search for
rival dialects cannot find absent validation — only a walk of where terms are *constructed* can.

**Two silent drops around the base direction, one found next door and one created here.** A base
direction outside `ltr`/`rtl` was dropped without a diagnostic, so the term kept its lexical form and
language and an `rdf:dirLangString` quietly became an `rdf:langString` — the graph then asserting a
literal the container does not contain. That is the documented mirror of over-refusal. The writer
filters direction before emitting, so such a value can only arrive from a foreign container.

The second was this change's own, and it is the same bug reached from the other side: the new tag
refusal cleared the language and left a **valid** direction standing, producing a term with a direction
and no language — outside RDF 1.2's value space. It was not inert. The importer downstream refused
such a term with `gts-direction-without-language`, naming a defect the container never had and pointing
a reader away from the tag that actually failed; the pre-change reader at least named the real cause.
The direction now goes with the tag, under its own diagnostic rather than riding silently on the tag's.
A `"dir"` with no `"l"` at all is deliberately left alone, because there the importer's complaint is
the correct name for a real defect. Keeping those two cases apart needs the refusal to be distinguished
from an absence, which a bare `None` cannot do.

**Surfaces that remain open.** This change created none of them — an earlier revision claimed it
created one, and that was wrong. Most are caller-deliberate: an infallible public constructor a caller
must hand-build an invalid literal with and then pass to a serializer. **The `crates/shapes` surface at
the end of this list is not**, and is called out as a byte path rather than filed with the rest.

- `ScratchInterner::intern`. An earlier revision called this door "created by this change". **That was
  wrong**: `intern` was already `pub` and already ungated on the merge base, and this change added the
  checked sibling `intern_checked` beside it rather than creating the door. What is true is narrower —
  restoring `intern`'s published signature is why the gate went on a sibling instead of on `intern`
  itself, so the door stayed open rather than closing. Its own documentation says so outright.
- the kernel's public emitters (`emit_quad`, `emit_term`, `Display`) on a hand-built `RdfLiteral`, and
  `TermValue::lang_literal`;
- any third-party `impl DatasetView` handed to `serialize_dataset_to_format` — the trait is public and
  unsealed.

- `purrdf_shapes::Literal::new_language_tagged_literal_unchecked` and
  `new_directional_language_tagged_literal_unchecked` (`crates/shapes/src/term.rs:131,145`). An earlier
  revision removed this surface on the claim that it "is gated with the rest"; **that claim was false**,
  and the revision after it named the wrong type (`Term`, not `Literal`) and the wrong caller. The
  accurate statement: `git diff --stat -- crates/shapes` against the merge base is empty and
  `crates/shapes` contains zero references to `purrdf_iri::langtag`. It is **not** caller-deliberate —
  its non-test caller is `import_term_object` (`schema_import.rs:1387`), which lifts `@language` out of
  an **external document** and is reachable from six public entry points: `import_json_schema`,
  `import_compiled_schema`, and the GraphQL, Pydantic, TypeScript and LinkML importers. Running any of
  them stores `en us`, `abcdefghi` and `!!!` verbatim on a literal typed `rdf:langString`. It is a byte
  path and it is open.

  It is not the only ungated surface in that crate, and this description will not put a number on them
  again — an earlier revision said "two further" and undercounted. Beyond the `_unchecked`
  constructors: `term_ref_to_native` and `term_value_to_native` (`term.rs:613,652`) populate `language`
  by **struct literal**, bypassing those constructors entirely; `sh:languageIn` tags are lifted from a
  literal's lexical form; and `schema_import` and the LinkML importer both recover tags by regex over a
  document-supplied `pattern`. The accurate and checkable statement is the crate-level one:
  **`crates/shapes` never asks the grammar at all.** That is what makes it a single coherent piece of
  work rather than a handful of patches, and it is out of scope for a change about the SPARQL runtime
  and the codecs.

## Verification

Observed on the built CLI, before and after:

| | before | after |
|---|---|---|
| `STRLANG("x","en us")` tsv | `"x"@en us` | unbound |
| `STRLANG("x","en us")` json | `"xml:lang":"en us"` | no binding |
| `STRLANGDIR("x","en us","ltr")` tsv | `"x"@en us--ltr` | unbound |
| `STRDT("x", rdf:langString)` tsv | `"x"^^...#langString` (tagless) | unbound |
| `STRLANG("x","en-US")` tsv | `"x"@en-us` | `"x"@en-us` |
| `STRDT("1", xsd:integer)` | `"1"^^xsd:integer` | `"1"^^xsd:integer` |
| `SELECT` vs `CONSTRUCT` | disagreed on both | agree on both |

Injection is structurally dead rather than patched: the profile is a whitelist
(`[a-zA-Z]+('-'[a-zA-Z0-9]+)*`), so no quote, angle bracket, ampersand, tab, newline or non-ASCII byte
can pass. Variants built indirectly — `CONCAT("en"," ","us")`, `REPLACE("en-us","-"," ")` — are refused
too.

Refused: `en us`, `1`, `9-9`, `123-456`, `en-`, `-`, `!!!`, `abcdefghi`.
Accepted: `en`, `en-US`, `zh-Hans-CN`, `de-CH-x-phonebk`, `i-enochian`, `x-purrdf-afrikaans`,
`x-gmeow-english`, `en-fr-jura`, `fr-be-fbcl`, `abcdefgh`, `en-x-cantbethislong`.

Both accept lists are the load-bearing half, and the fold-view gate (escape 8) is the one that most
needed them: it sits on a constructor every fold view in the workspace goes through, so an over-refusal
there would be felt by every caller. It is driven with that accept list, substituting
`x-gmeow-norwegiannynorsk` for `x-gmeow-english` — a strictly harder case, a 16-character private-use
subtag — including the tags
that are well-formed only under the concrete-syntax profile and not under RFC 5646 §2.1 — and the full
`purrdf-rdf` suite is green behind it. The base-direction rule is likewise driven from all three sides:
a refused tag drops the direction and says so, an accepted tag keeps it untouched, and a direction that
never had a tag beside it is left alone for the importer to report accurately.

`purrdf-sparql-eval` 1313 → **1339**. `purrdf-sparql-results` 168 → 172. `purrdf-gts` 195 → 198.
Every count here was re-measured against the current head rather than carried forward; an earlier
revision quoted 1334, which was this branch's own total from before the last two commits added to it.
**W3C SPARQL conformance: delta zero**, measured by reverting the change, re-running the identical
command and comparing — no absolute count is quoted because an earlier revision published one that
reconciled with nothing. Clippy, the brand-casing and issue-reference hygiene gates, and the
documentation gate are clean, and no vector, W3C test manifest or expectation was edited. Three
`Cargo.toml` files do change: comments, plus one `{ workspace = true }` line adding the `purrdf-gts` →
`purrdf-iri` edge. No dependency version is pinned or moved in any of them. `Cargo.lock` gains exactly one line — a workspace-internal `purrdf-gts` → `purrdf-iri` edge, a
zero-dependency leaf — and **no third-party version moved**. CI's result on this head is whatever the
checks say; this sentence no longer asserts it in advance, because an earlier revision did and was
wrong.


paudley:
fix(sparql,gts,rdf): ask the language-tag grammar wherever a term is constructed

A query string was enough to make PurRDF emit bytes no parser reads.
`STRLANG("x", "en us")` bound a literal that serialized as `"x"@en us` in TSV and
`"xml:lang":"en us"` in JSON and XML — from the CLI, the C ABI, wasm and the Rust
API alike. `STRLANGDIR` had the same hole.

## Why the existing gates did not catch it

The preceding dependency-trimming work gave language tags one owner and gated the
IR kernel at `RdfLiteral::validate_components`, so nothing enters an `RdfDataset`
ungrammatical. It then claimed every construction path reaching a serializer was
covered. That claim was false for SPARQL results, because that path has no dataset
in it: `STRLANG`'s literal goes into a scratch arena and straight out to the
results writer.

The query parser does validate a tag written in query text, on this exact profile.
`STRLANG` is a runtime constructor, so it walked around a gate that exists and
works rather than past a missing one. The diagnostic that exposed it: CONSTRUCT
refused the same tag SELECT printed — same engine, same tag, opposite verdicts,
because a CONSTRUCT template routes through the dataset builder.

## What changed

Every surface below now asks `purrdf_iri::langtag` on
`Profile::ConcreteSyntaxLangtagBounded` — the profile the parser, the codecs and
the IR kernel already name.

Failure modes differ deliberately, each following its own precedent. STRLANG
leaves the expression unbound (SPARQL 1.1 §17.4.2.5 makes an ill-formed argument
an error; §17.2 turns an expression error into an unbound result rather than an
aborted query; it is what these functions already did for an empty tag). Parsers
refuse hard, because an unreadable tag is a parse error for the same reason the
unreadable base direction beside it already was.

Eight escapes of this class were found and closed. The list is what was found,
not a completeness claim:

 1. STRLANG / STRLANGDIR — runtime constructors past the parser's gate.
 2. The SPARQL results readers (from_json, from_xml) — the federated SERVICE path.
 3. STRDT with rdf:langString — minted a literal with no tag at all.
 4. GTS decode — the tag came straight off the wire.
 5. Composite-datatype lifting — cdt/list functions lift members out of a
    literal's lexical form, so `"[\"x\"@en us]"^^cdt:List` carried a tag the
    kernel never admitted. Nobody had named this one; it surfaced because the gate
    went at the convergence rather than at the known seams.
 6. The Python fold view built from parts, which bypasses reader::read entirely.
 7. A SPARQL pre-binding. SparqlRequest substitutions and SHACL pre-binding build
    a literal from a caller-supplied tag straight into the scratch arena. The
    algebra validator does refuse a hand-built Query at admission — but
    substitution rewrites a plan already admitted, so that gate cannot see these.
    Gated at the ingress, because a refusal deeper in would have deleted a
    constraint rather than raised one.
 8. The Rust fold view. GtsFoldView::with_config accepted a caller-assembled Graph
    without asking; Graph.terms and Term::lang are both pub. The exact twin of 6,
    gated in Python and open in Rust — left standing, the language a fold view
    renders would have depended on which binding assembled the graph.

The gates sit at convergences (validate_components, the scratch interner, the GTS
term-decode loop, JSON-LD lower_literal, the ShEx ingress) rather than at
enumerated seams, so completeness does not depend on having listed every caller.
That choice is what surfaced escape 5.

## Two silent drops around the base direction

A base direction outside ltr/rtl was dropped without a diagnostic, so an
rdf:dirLangString quietly became an rdf:langString and the graph asserted a
literal the container does not contain. That is the documented mirror of
over-refusal.

The second was this change's own, found by audit: the new tag refusal cleared the
language and left a valid direction standing, producing a term with a direction
and no language — outside RDF 1.2's value space. Not inert. The importer
downstream then refused it with gts-direction-without-language, naming a defect
the container never had and pointing away from the tag that actually failed; the
pre-change reader at least named the real cause, so this was a regression in
diagnostic quality. The direction now goes with the tag under its own diagnostic.
A "dir" with no "l" at all is deliberately left alone, because there the
importer's complaint is the correct name for a real container defect — keeping
those cases apart needs a refusal distinguished from an absence, which a bare
None cannot do.

## Over-refusal — executed, not argued

Over-refusal is a first-class bug and the mirror of silent drop, so every gate was
driven with a neighbouring valid case as well as the invalid one. Accepted
throughout: en, en-US, zh-Hans-CN, de-CH-x-phonebk, i-enochian,
x-purrdf-afrikaans, x-gmeow-english, x-gmeow-norwegiannynorsk, en-fr-jura,
fr-be-fbcl, abcdefgh, en-x-cantbethislong. Several of these are well-formed only
under the concrete-syntax profile and not under RFC 5646 §2.1, which is the whole
reason the profile exists.

No corpus vector is refused by any gate. Verified by extracting every language tag
in vectors/shexTest, the W3C SPARQL suite and the RDF corpora and checking each
against the profile. W3C SPARQL conformance delta is zero, measured by reverting
the change and re-running the identical command.

The escape-8 gate was the highest risk, sitting on the constructor every fold view
in the workspace goes through; the full purrdf-rdf suite is green behind it.

## Semver

No published Rust signature changed and no version was edited. Two signatures were
broken during development and restored: ScratchInterner::intern (the gate moved to
a new checked sibling, intern_checked) and purrdf_wasm::DataFactory::literal
(restored verbatim in a non-#[wasm_bindgen] block, with the bindgen method renamed
checked_literal under #[wasm_bindgen(js_name = literal)] so JS keeps the refusal;
a compile-time test fails the build if it regresses). GtsFoldView::new and
with_config already returned Result, so the new gts-unreadable-language-tag is one
more arm on an existing refusal.

Cargo.lock gains exactly one line: a workspace-internal purrdf-gts -> purrdf-iri
edge, a zero-dependency wasm-clean leaf. No third-party version moved. Three
Cargo.toml files change: comments plus one `{ workspace = true }` line.

## Surfaces that remain open, and disclosed

Most are caller-deliberate infallible constructors: the kernel's public emitters
on a hand-built RdfLiteral, TermValue::lang_literal, any third-party impl
DatasetView, and ScratchInterner::intern (already pub and ungated on the merge
base — this change added intern_checked beside it rather than creating the door).

One is NOT caller-deliberate and is named as a byte path:
purrdf_shapes::Literal::new_language_tagged_literal_unchecked and its directional
sibling. Its non-test caller is import_term_object, which lifts @language out of an
external document, reachable from six public entry points (import_json_schema,
import_compiled_schema, and the GraphQL, Pydantic, TypeScript and LinkML
importers). crates/shapes never asks the grammar at all — a single coherent piece
of work rather than a handful of patches, and out of scope for a change about the
SPARQL runtime and the codecs. It is disclosed in the PR description rather than
softened or omitted.

## Completion-adversary

Seven passes ran. Passes 2-6 each found a real defect; pass 7 returned PASS.

Pass 6 findings, all discharged by execution in commit c686733a:
 - F7: the CI workspace job was red on a word this branch added. The fix to
   check-issue-refs.py had been verified against that checker rather than against
   the job that runs it; check-brand-casing.py beside it was failing.
 - N1: the orphaned base direction above.
 - N2: the Rust fold view. The adversary advised disclosing; it was gated instead,
   since the constructor already returned Result and the Python twin was gated.
 - N3: five false statements in the PR body, each verified against source before
   correction — the test count (measured 1339, body said 1334), import_term_object
   not import_json_schema, Literal:: not Term::, intern not created by this change,
   and an undercount replaced by a checkable crate-level claim.

Pass 7 per-criterion demonstrations, on HEAD:
 - F7: check-brand-casing.py and check-issue-refs.py both EXIT=0; all 25 steps of
   the CI workspace job enumerated and every runnable one green; the workspace job
   itself SUCCESS on CI.
 - N1: probe through the public writer -> reader::read -> fold view -> RDF import.
   Refused tag + valid ltr now yields lang:None direction:None, two DamagedFrame
   diagnostics naming both drops, and a clean import. dir-without-l still yields
   gts-direction-without-language, which is the accurate name there. Confirmed the
   crate has a single language-tag decode point, so no path skips the new arm.
 - N2: purrdf-rdf all 52 test binaries green; purrdf-gts 198 green; the
   constructor reached from exactly three non-test places, no corpus vector
   refused; published signature unchanged; the refusal message leaks no path,
   machine detail or process reference.
 - N3: every number re-measured — sparql-eval 1339, sparql-results 172, gts 198,
   Cargo.lock exact. Every code reference re-traced, including a hostile trace of
   the LinkML path to import_term_object.
 - .deficiencies: both ledgers at their normal state, notice and marker, zero
   entries.

## Deferral gate

Mechanical scan clean over the full diff and all seven commit messages. One
regex hit on commit text was a false positive: the pattern "this PR" matched
"this project" under case-insensitive search. check-issue-refs.py EXIT=0 — no
issue or PR numbers and no process references in repo docs or comments.

## Test counts

purrdf-sparql-eval 1313 -> 1339. purrdf-sparql-results 168 -> 172.
purrdf-gts 195 -> 198. purrdf-shex 196. purrdf-rdf +3.
W3C SPARQL conformance delta zero. fmt, clippy (workspace, -D warnings) and the
rustdoc gate all clean.

## Note for whoever reads this next

The recurring lesson across seven passes: the defect was never a validator that
answered wrong, it was a surface that never asked. Searching for rival dialects
cannot find absent validation — only a walk of where terms are constructed can.
Two of the eight escapes were found only by asking why one binding's door was
gated and its twin's was not.

PR #329, opened 2026-09-17
Add the micro-hardware envelope capture tool

paudley:
## Summary

`purrdf-envelope-probe` (never published): the capture side of the small-deployment validation envelope. One binary, five named profiles (`smoke`, `wasm-browser`, `sbc-32-small`, `sbc-64`, `edge-gateway`), five workloads over the public APIs:

- **roundtrip** — N-Quads/TriG/JSON-LD dataset round-trips with declared-drop reconciliation (row counts must balance against `SerializeOutcome`'s drop counters).
- **governed_query** — a corpus-scaling join under `QueryGovernors::METERED` (measure, don't bound), plus a 16-cell truncation case *and its completing selective neighbor* — the over-refusal discipline applied to budgets.
- **shacl** — governed validation over the keystone corpus; the corpus's scoped blanks violate the shape once per group, so the violation-report path is exercised at scale.
- **gts** — write/read round-trip through the events importer.
- **pack_paged** — pack build/open/query, then the paged tier: bounded scan evidence, and a refusal pair proven at the *measured* page boundary (budget−1 refuses, exact budget completes).

Output is a JSON report per run: allocator peak/retained per workload (counting `#[global_allocator]`), `VmHWM` cross-check, durations (recorded evidence, never asserted), and per-workload domain metrics including governor consumption vectors.

## Findings from the first captures (recorded, not fixed)

- At `wasm-browser` scale (~89k rows), the round-trip workload peaks ~304 MB allocated — above the profile's 256 MiB ratchet start. The single-string serializers with no incremental sink are the known cause, already on the bounded-memory audit list.
- `single_graph_pages_touched` equals the page count: a graph-selective query over the paged tier materializes every page (no per-page graph index in sealed metadata). The probe records this datum on every run.

Numbers are dev-profile, single-host, report-only.

## Verification

- `make check` green (make's own exit code, unpiped), including the five-member never-published claims updated in `AGENTS.md` and the releases page.
- `cargo build -p purrdf-envelope-probe --lib --target wasm32-unknown-unknown` clean.
- Smoke tests run every workload; full `wasm-browser`-scale run completes end-to-end.

## Summary by CodeRabbit

* **New Features**
  * Added an envelope probe tool for deterministic dataset, query, validation, serialization, and paged-query workloads.
  * Added named profiles, configurable group counts, and machine-readable JSON results with timing, memory, RSS, and workload metrics.
  * Added checks for data preservation, query limits, complete validation, and bounded scans, with detailed failure diagnostics.
[...]

PR #322, opened 2026-09-17
feat(320): deterministic HNSW index for PurRDF

paudley:
Closes #320

## Summary

`purrdf-hnsw` — a deterministic HNSW approximate nearest-neighbour index over a PURREMB
embedding matrix, registered on the evaluator's property-function seam under a
caller-supplied predicate IRI, standing beside the exact kNN relation and never replacing it.

The graph is a pure function of the data. Levels are `splitmix64(row)` rather than an RNG
draw, the build is round-structured against a frozen snapshot, candidates order by the exact
path's own `(distance, row)` comparator, and the canonical byte image is identical across
1/2/4/8 rayon workers and across `wasm32-unknown-unknown`.

## What changed after review

The determinism story was sound from the start. The graph construction was not, and a
stage-2 audit found defects that every gate had passed over.

**The build stranded rows.** Round 1 proposed against an empty snapshot and committed no
edges, so `isqrt(n)` rows ended the build with no inbound edge — unreachable from the entry
point, never returnable, not even as their own nearest neighbour at distance zero. A second
mechanism, degree-bound truncation dropping a node's last inbound edge, cost a further ~3% of
rows at `M0=8` and ~99% at `M0=2`. No determinism check could see either: a deterministically
wrong graph is still deterministic. The entry point is now computed before any linking, the
schedule doubles from one row under a cap, and a final serial pass repairs the truncation
case. **Measured stranding is zero across every parameter set from 17 to 5,000 rows**, and
the invariant that asserts it is now in the suite.

**Neighbour selection made the graph unnavigable.** Nearest-`M` truncation spends every edge
in the same direction, so greedy descent stalls and only a corpus-sized beam recovers.
Selection is now the relative-neighbourhood condition (Malkov & Yashunin, Algorithm 4). On
the crate's own 2,048-row fixture, recall@10 at `ef=64` went **0.7875 → 0.9953**, and the work
needed for 0.96 recall fell to 23% of the corpus.

**The admission gate asserted nothing.** Its only recall assertion was
`assert!((0.0..=1.0).contains(&recall))` — true of every possible run, passing at recall
0.000. Recall is deterministic, so it is now pinned as an **exact equality** across all 96
`(fixture, regime, k)` regimes, with rows-offered beside every hit count so no entry can be
read without it.

**The corpus could not exhibit the property being measured.** Fixtures were uniform i.i.d.
vectors — the one input class whose distances concentrate so hard that no index scores well,
[...]

Issue #294, opened 2026-09-11
fix(rdf): the XML writers emit literals they cannot read back


paudley:
## The defect

`escape_xml` is transcribed twice, byte-identically, in the two XML-shaped writers:

- `crates/rdf/src/native_codecs/rdfxml.rs:1190`
- `crates/rdf/src/native_codecs/trix.rs:550`

Neither escapes `#xD`, and neither escapes the C0 controls. Both are wrong, in two different ways, and both are silent.

### 1. A carriage return in a literal round-trips as a line feed

XML 1.0 §2.11 requires a parser to normalize a literal `#xD` in character data to `#xA`. A CR that must survive has to be written `&#xD;`. We write it raw:

```
$ purrdf convert --from ntriples --to rdfxml cr.nt cr.rdf     # literal "a<CR>b"
    <ns0:p>a<CR>b</ns0:p>                                     # raw CR in character data
$ purrdf convert --from rdfxml --to ntriples cr.rdf -
<urn:ex:s> <urn:ex:p> "a\nb" .                                # a DIFFERENT literal
```

Exit zero both ways. The same loss occurs through TriX.

### 2. A C0 control produces non-well-formed XML

```
$ purrdf convert --from ntriples --to rdfxml nul.nt nul.rdf   # literal "a<U+0000>b"
exit=0                                                        # raw U+0000 in character data
$ purrdf convert --from rdfxml --to ntriples nul.rdf -
purrdf: error native-codec-parse: RDF/XML: a non-XML character found at 4:13   exit=1
$ python3 -c "import xml.etree.ElementTree as ET; ET.parse('nul.rdf')"
xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 4, column 12
```

We emit, at exit zero, a document our own reader refuses and no conforming XML parser will accept. U+0000 is not in XML's `Char` production at all, so no escape rescues it — the serializer has to refuse.

## Why this is filed rather than fixed in place

It was found while auditing the scanner-terminal work, which collapsed five other escape transcriptions onto one law. These two are a sixth and seventh, and the defect is the same shape as that work's headline.

But fixing it **changes emitted bytes** for any document whose literals contain a CR, and converts today's exit-zero-with-corrupt-output into a hard error. Either may legitimately move a frozen vector or a conformance row. That belongs in a change whose diff is about byte movement and is reviewed as such, rather than riding along in one about character classes.

## What closing this looks like

[...]

Issue #280, opened 2026-09-09
Verify and benchmark under LargeRDFBench


sanmai-NL:
### Problem

I want to have additional assurance that purrdf functions correctly, after some major defects surfaced, but gladly were quickly fixed.

Moreover, since performance efficiency seems like a major concern from how this project is positioned and tested, using a large benchmark seems interesting from that angle.

### Proposed solution

- Develop a benchmark setup, perhaps out-of-tree given the space and time consumption of running it.
- I can arrange compute and storage if needed.
- Data is not yet available, for one, https://github.com/dice-group/LargeRDFBench/pull/4 must be completed first.

### Language surface(s)

- [ ] Rust crate
- [ ] Python
- [ ] JavaScript/wasm
- [ ] C ABI

paudley:
I'm more than happy to integrate a custom test suite!  Or add additional suites if you have some in mind.

PR #332, opened 2026-09-18
Add the deterministic shardable scale-corpus generator

paudley:
## Summary

Three never-published additions to `purrdf-bench`, the workspace's sixth unpublished member: the
deterministic scale-corpus generator, and the two comparison-workload lanes that consume
externally specified suites. All three are **report-only** — no number here is a gate.

### `bench-corpus` — the scale corpus

Every IRI is minted as a pure function of `(seed, index)`, so generation is **shardable**: shard
`k` of `n` emits exactly its slice and concatenating all shards is byte-identical to a whole run.
That is a property of index-pure minting, which holds at any scale; it is not a claim that a
full-scale run has been performed. At the measured density a 10¹⁰-row corpus is **~1.83 TB**, so
full scale is an operator activity, streamed into a consumer rather than stored.

Two pinned tables are the contract, and a `const` assertion proves each sums to 1000:

- **Entity-IRI classes** (shares of the entity index space): plain 400, numeric-long 200,
  chinese 200, irregular 150, very-long 50. Numerics are exactly 36 digits, beyond both `u64` and
  2⁵³, with leading zeros preserved. The irregular class escapes two RFC 3986 **reserved** octets
  in mixed case — escaping *unreserved* octets is normalization-equivalent, so a conformant
  normalizer would decode them away and the class would stress nothing. Very-long is ~628 bytes.
- **Row kinds**: entity-edge 500, plain-literal 150, zh-literal 100, typed-literal 100,
  long-text-literal 50, reified 60, blank-node 40.

The corpus carries **RDF 1.2**: triple terms and reifiers in the shape the project's own N-Quads
reader admits, typed literals across the XSD value space, and blank nodes in two families — so the
statement layer, the datatype dictionary and blank-node handling are exercised at scale rather
than only in a micro-hardware fixture.

Every decision draws from its own tagged `splitmix64` stream, so row kind, literal type, graph
membership and graph id are independent. All derivation is integer arithmetic — no RNG syscalls,
no floats, no iteration-order dependence. The profile `purrdf-scale-mixed-v1` is pinned by a
byte-level golden digest; from here any derivation change is a new profile id.

`make scale-corpus` drives the shards, streaming by default and retaining nothing. Writing shard
files is opt-in, and their names carry the profile, the parameters and a zero-padded index so
sorted concatenation reproduces a whole run byte for byte.

### The comparison workloads — acquired by digest, never vendored

The licensing review is complete and its conclusions are encoded in `scripts/benchmark-acquire.py`,
which fetches six artifacts by SHA-256 and vendors none of them. The LUBM generator is
GPL-2.0-or-later, so it is **run, never copied** into this MIT/Apache tree. `univ-bench.owl` and
the query file carry no licence text at all, so neither is redistributed. WatDiv is citation-ware:
every result derived from it must cite Aluç, Hartig, Özsu and Daudjee, ISWC 2014, pp. 197–212. A
cached file whose digest stops matching is quarantined, never re-fetched over.

**`make lubm`** generates with the stock generator unpatched, converts RDF/XML to N-Quads through
the `purrdf` CLI, normalises the published queries and runs all 14. The queries could not be run as
published: they predate final SPARQL and bind the ontology prefix to a 2004 draft namespace no
dataset uses, so unnormalised they parse, execute and answer **zero**. Five numbered rules fix that
and every application is recorded before and after. Each query carries its entailment regime from
the canonical paper, because comparing an engine that infers against one that does not is how
benchmark numbers lie. **Q1, Q2 and Q14 need no inference and return 4, 0 and 5916 — the published
answers for one university at seed zero.**

**`make watdiv`** consumes the published ten-million-triple corpus by digest. WatDiv's generator
seeds from the clock in three places and has no seed flag, so the generation *process* cannot be
pinned — only a frozen output can. Its 20 templates carry placeholders that upstream fills with
that same clock-seeded RNG; because the corpus is frozen, this lane draws from the fixed candidate
set deterministically instead, recording every choice. One seed gives one query set byte for byte.
All 20 execute, returning 434,748 rows. Five match nothing, and each was diagnosed rather than
assumed — three drew a term whose instances genuinely lack the property asked for (WatDiv's
deliberate skew), one drew a user who likes nothing, and one has no mapping at all. A run where all
20 matched nothing fails rather than reporting success.

## Verification

- `make check` → exit 0, unpiped. `make doc` → exit 0, zero rustdoc errors.
- **96 tests**, from 4 at the branch's first commit. The binary target ran **0** tests then; 20
  integration tests now drive the shipped binary rather than the library, because every silent
  `--out` divergence this branch fixed was invisible to a library test.
- `make lubm` → exit 0, 14/14 queries executed. The published answers are reproduced for the
  three inference-free queries (Q1, Q2, Q14) over the full corpus; the remaining eleven are
  answered over reduced rungs, because the entailment fact ceilings are fixed and no flag raises
  them. The lane says so in its own output: a row count on a rung below `full` is not the
  published LUBM answer.
- `make watdiv` → exit 0, 20/20 queries, row counts identical across independent runs.
- Shard concatenation byte-identical to a whole run at uneven shard counts and where shards
  outnumber quads; cross-process determinism byte-compared.
- Ten billion distinct IRIs generate in a debug build (`overflow-checks` on) reaching index
  9,999,920,083.

## What the entailment ceilings mean, stated precisely

Eleven of LUBM's fourteen queries need entailment, and the reasoning engine holds fixed ceilings:
`MAX_STORED_FACTS` is 131,072 facts, against a LUBM(1) corpus of 102,737 rows. The lane therefore
reproduces the published answers for Q1, Q2 and Q14 — the three that need no inference — over the
full corpus, and answers the other eleven over reduced rungs, saying in its own output that a row
count on a rung below `full` is not the published LUBM answer.

Two separate things are true here and it is worth not blurring them.

The **ceilings being constants** is deliberate and permanent. `MAX_JOIN_STEPS` is a `const`
because "a caller-supplied budget would mean two callers running the same program over the same
input get different answers — the same semantic optionality the project's no-Cargo-features rule
exists to prevent, merely arriving through an argument", and `MAX_STORED_FACTS` is sized so a
saturated store stays inside a `wasm32` linear memory. Making either a knob would trade away
no-optionality or portability.

**Materializing a closure whole in one store is a separate choice**, and that rationale says
nothing about it. A partitioned or incremental closure would answer these queries over the full
corpus without raising any ceiling and without introducing optionality. That work is unstarted and
lives in `purrdf-datalog`, which this branch does not touch. Claiming the current reach is "the
shape the engine is meant to have" would assert a design decision the cited rationale does not
make, so this states the reach as measured and leaves the strategy question where it belongs.


paudley:
# Stage 2 — Gap analysis and remediation: PR #332 (deterministic shardable scale-corpus generator)

## Context

**#332 is a pull request, not an issue.** GitHub shares one number space, so `gh issue view 332`
returns the PR. There is no tracking issue and no linked issue, so the acceptance contract is the
PR title + body + commit message + `crates/bench/README.md` + the in-source doc comments — *all
implementer-authored*, and therefore treated as claims to falsify rather than requirements met.

The branch (`paudley/bench-corpus`, one commit `baa52aa`, 9 files, +510/−7) adds an unpublished
`purrdf-bench` workspace member whose `bench-corpus` binary mints every IRI as a pure function of
`(seed, index)` across five adversarial classes, making generation shardable and byte-deterministic
under profile id `purrdf-scale-mixed-v1`, pinned by a golden digest.

**The architecture is right and the wiring is complete.** The defects are all in execution: the
generator does not emit the corpus its own documentation describes, and **it cannot generate at the
headline scale at all**. Because a profile id plus a golden digest *certify* a corpus, each
mismatch means the pin currently certifies the **wrong** corpus — and the digest actively *protects*
the mismatches from discovery.

Three independent audits ran: a static gap-analyzer, a completion-adversary that executed the CLI,
and my own hand-trace. Both agents independently reimplemented the generator's arithmetic and
reproduced the pinned FNV digest `0xB6E887238FD8AB29` byte-for-byte, so every measurement below is
authoritative rather than estimated. All five CodeRabbit findings are unaddressed and all five are
real; two were materially **under**-rated.

---

## Decisions taken (recorded — do not re-litigate)

| Decision | Resolution |
|---|---|
| Profile id across byte-changing fixes | **Keep `purrdf-scale-mixed-v1`, re-pin the golden digest once.** The profile is *born* in this PR: never on `main`, never published (`publish = false`), zero consumers. Correcting its definition pre-publication is not a silent regeneration. (Overrides the adversary's recommended `-v2` bump.) |
| `CorpusSpec` invariants | **Private fields + `new(..) -> Result`**, `shard_range()` infallible by construction, `parse_args` a thin caller. |
| Correlated draws | **Separate stream tags** per decision. |
| RDF 1.2 / term-kind coverage | **Full: reifier/triple-term rows, typed literals across the XSD value space, blank nodes in two scopes**, at pinned per-mille shares. |
| 10B corpus on CI | **Not materializable — CI runners lack the disk** (~1.9 TB at measured 190.7 B/quad). Drives the H6 design below: streaming/pipe-first, CI smoke at small scale only, and the 10B claim restated honestly. |

---

## Verified correct — do NOT "fix" these

Measured or executed, and cleared. Recorded so remediation does not churn them:

- **`clippy` is clean for `purrdf-bench`.** I ran `cargo clippy -p purrdf-bench --all-targets
  --locked`: zero warnings. The static audit flagged `lib.rs:227`'s positional `{:064x}` as a
  probable `uninlined_format_args` failure against `-D warnings`; it is a **false positive** —
  `splitmix64(o)` cannot be inlined, which exempts the whole call. The PR's "make check green"
  claim stands on the clippy axis.
- **Class mix is exact**: measured 399.2–403.0 / 200.6–201.8 / 197.5–200.2 / 149.2–151.4 /
  47.5–49.6 per mille against the declared 400/200/200/150/50. Modulo bias ≈ 5e−17.
- **`shard_range` partitioning is correct** (`lib.rs:53`): no gap, no overlap, for `shards=1`,
  uneven division (`quads=50000, shards=7` → last shard `[42852, 50000]`), the final shard, and the
  degenerate `shards > quads`. `shard*base < quads`, so no overflow.
- **36-digit numerics**: measured digit-width set is exactly `[36]`; 0 of 40,358 sampled values fell
  below `u64::MAX`.
- **Named-graph fraction**: measured 0.16674 — exactly the documented one sixth.
- **The golden digest genuinely pins the production surface**: the CLI's emitted bytes FNV to
  `0xB6E887238FD8AB29`, identical to the test pin. The golden is not library-only.
- **Manifest JSON is valid** — `push_str("}}")` is not a format macro, so it correctly closes both
  objects; all fields numeric or fixed, so no escaping needed.
- **`crates/bench/Cargo.toml` obeys every hard constraint**: no `[features]`, `publish = false`,
  `[lints] workspace = true`, `purrdf-rdf = { workspace = true }` (no member-local pin, dev-only, so
  the generator has zero runtime deps), SPDX on all four new files.
- **`make wasm` cannot break** — it builds an explicit `-p` list with `--lib` and does not include
  `purrdf-bench`.
- **Workspace wiring complete; the count claim is TRUE.** Both agents verified by enumeration:
  exactly six members carry `publish = false` (bench, capi, cli, envelope-probe,
  sparql-conformance, python). `crates/gts/fuzz`, `crates/geo/determinism`,
  `crates/hnsw/determinism` are workspace-`exclude`d, are not members, and are correctly not
  counted — `scripts/check-doc-claims.py:393-410` does not reach them either.
- **N-Quads is not a `.goals` violation.** Evidence, not intent: the project ships a strict
  `application/n-quads` reader (`lib.rs:299`); `docs/RDF12-CANON-PROFILE.md:145` treats canonical
  N-Quads as a deliberately supported projection; `gts_fixtures.rs:83-93` uses
  `push_quad(s,p,o,Some(graph))` — a graph *name*, which RDF 1.2 defines. The `.goals` ban bites at
  the model layer (4-tuples in place of triple terms), which this does not do.
- **No over-refusal anywhere.** Every refusal was paired with its neighbouring valid input and all
  pass: `--quads 0`/`--iris 0`/`--shards 0`/`--shard>=--shards`/non-numeric/no-args/`--bogus` all
  exit 2 with a diagnostic, while `--quads 1 --iris 1`, `--shards 1`, `--shard 6 --shards 7`,
  `--iris 1`, `--iris 4294967295`, `--quads u64::MAX` and `--quads 3 --shards 8` all succeed. The
  failure mode on this branch is the **opposite** — under-refusal (C1, H1, M7).
- **Throughput is sound**: 2M quads in 0.477 s, ~190.7 B/quad. The 10B claim fails on C1, not speed.
- **`.deficiencies` is clean** (notice + marker, zero entries) in both the worktree and the main
  checkout. No laundering there, and no issues were filed by the implementing session.
- **Not a DUPLICATE**: `purrdf_rdf::gts_fixtures::keystone_*` is a small fixed micro-hardware
  fixture; this is a shardable 10B-scale profile. Different jobs. (Cross-reference them — M12.)

---

## Gaps to fix

### CRITICAL

#### C1 — `--iris` is structurally capped at 2³², so the generator cannot produce B2's ten-billion-IRI corpus, and the manifest certifies a corpus that was never generated

`crates/bench/src/lib.rs:172-178`. **Two defects, and the second is the one CodeRabbit missed.**

(a) `(biased * iris) >> 32` overflows u64 for `iris > 4_294_967_298`. The root `[profile.dev]` sets
`overflow-checks = true` (inherited by `[profile.test]`), so dev and test **panic**; release wraps
silently. Executed:

```
debug   --quads 100000 --iris 10000000000   exit=101  panicked at lib.rs:177: attempt to multiply with overflow
release --quads 100000 --iris 10000000000   exit=0    max entity index emitted = 4294869752
release --quads 200000 --iris 2^40          exit=0    max entity index emitted = 4294885632
release --quads 200000 --iris u64::MAX      exit=0    distinct subjects = 2 / 200000 rows, no diagnostic
```

(b) **Widening to `u128` stops the panic but does not make ten billion reachable.** `hi = r >> 32`
discards the low 32 bits, so the entity map has at most 2³² distinct outputs *by construction*.
Meanwhile `manifest()` faithfully records `"iris": 10000000000`. The pin therefore certifies a
corpus that does not exist — exactly the failure that makes every downstream capacity claim inherit
the error.

**Fix:** derive the skew from the **full 64-bit draw** (128-bit square in fixed point, `u128`
throughout), then narrow with the bound `result < iris` encoded in the type rather than a bare cast.
Changes bytes at *every* `--iris`, not only the broken range.

**Tests:** at `--iris 10000000000` assert every emitted entity index is `< iris` **and** that the
observed maximum approaches `iris` (the ceiling test (b) would otherwise pass silently); run under
debug **and** release.

#### C2 — the emitted entity:literal ratio is 4/7 : 3/7, not the documented 2/3 : 1/3

`crates/bench/src/lib.rs:207`, contradicting `lib.rs:196` and the PR body. The uncommented
`!o.is_multiple_of(7)` guard diverts 1/7 of the entity arm to literals: P = 2/3 × 6/7 = 4/7.
Measured independently by both agents over 200k–300k rows: **0.5717 / 0.4283** and
**0.5722 / 0.4278**.

**Fix:** remove the guard, restoring exactly 2/3 : 1/3. Verified all three literal branches stay
reachable: with `o%3==2`, `o%15 ∈ {2,5,8,11,14}` so `o%5` takes every value.

#### C3 — the "mixed-case percent-encoded tail" is false three ways, and the class stresses nothing

`crates/bench/src/lib.rs:146-154`. Format `…/x/%{:02X}{:02X}/…` emits `%46` then bare literal `72`.
Measured over 300k rows: rows with a second percent-escape = **0**; tails containing a lowercase hex
digit = **0**. Every tail is `%` + four uppercase hex chars (`%5A69`, `%4672`, …).

**The deeper defect neither review raised:** the single encoded byte is `0x41 + (h>>16)%26` =
`A`–`Z`, an RFC 3986 **unreserved** character. Percent-encoding an unreserved character is
normalization-equivalent (RFC 3986 §6.2.2.2), so any conformant normalizer decodes it away and the
class's stated purpose — "defeats suffix tricks" — evaporates entirely.

**Fix:** emit `%{:02X}%{:02x}` (two escapes, genuinely mixed case) over bytes that are **not**
unreserved (reserved/sub-delims or non-ASCII octets), so the stress survives normalization.

#### C4 — the corpus behind an RDF 1.2 capacity claim contains no RDF 1.2, and no typed literals

Measured over 300k generated rows: triple terms `<<(` = **0**, `<<` = 0, reifiers = **0**,
base-direction literals = **0**, typed literals `^^` = **0**, blank nodes = **0**. Only plain
literals and `@zh`. `.goals`: "RDF* and RDF1.2 are FIRST CLASS concerns and the only way we work."

`lib.rs:24-26` parks this — "the statement layer and blob surfaces are out of scope for the scale
corpus (they are exercised by the envelope probe's keystone corpus)". The pointer is *factually*
correct (`crates/rdf/src/gts_fixtures.rs:95-100` really does emit `push_reifier_in_graph` +
`push_annotation_in_graph` every 8th group) but **innocence-test leg 2 fails**: that is a small
fixed micro-hardware fixture, so **no capacity claim at scale covers the statement layer, blank-node
scoping, or the XSD value space** — and it cannot stress the annotation-layer indexing `purrdf-text`
advertises. The absence of typed literals is not even an RDF 1.2 question; it leaves the
datatype-IRI dictionary entirely unexercised.

**Fix (decided — full):** extend the existing class-mix machinery with pinned per-mille shares for
reifier/triple-term rows, typed literals spanning the XSD value space (integer/decimal/date), and
blank nodes in two scopes. Rewrite `lib.rs:22-26` to state the profile's coverage as a property,
with no "out of scope" deferral.

### HIGH

#### H1 — `--out` silently does the wrong thing, three ways

`crates/bench/src/main.rs:46` — `out = arguments.next()` neither requires an operand nor rejects a
flag-shaped one, and manifest mode ignores `out_path` entirely. Executed, unpiped:

```
--quads 5 --iris 10 --out                     exit=0  741 bytes of rows to STDOUT, stderr empty
--quads 5 --iris 10 --out --manifest          exit=0  created a file literally named "--manifest" holding ROWS
--quads 5 --iris 10 --out /tmp/x.nq --manifest exit=0  manifest to stdout; --out SILENTLY IGNORED, file never created
```

A shard-parallel driver would produce zero files and a green exit. Direct `.goals` "HARD FAILS"
violation. Every *numeric* flag is already safe — `take()` errors correctly on absent and
flag-shaped operands; `--out` is the sole unguarded operand.

**Fix:** put `--out` on the same mandatory-operand discipline, and honour `--out` in manifest mode
(a capture wants the manifest on disk beside its shard). Do **not** reject all leading-`-` paths —
that would be the over-refusal `CLAUDE.md` warns about.

**Over-refusal counter-check (mandatory):** `--out ./corpus.nq`, `--out /tmp/x.nq`, and
`--out /tmp/-weird.nq` must all still succeed.

#### H2 — one operation, two error disciplines: the manifest write is unchecked

`crates/bench/src/main.rs:69`. Executed:

```
--manifest > /dev/full          exit=101  panicked: failed printing to stdout: No space left on device
--quads 200000 --iris 100 > /dev/full  exit=1   bench-corpus: write failed: No space left on device   [correct]
```

The row path handles the identical failure cleanly 20 lines below. ETHOS §19 in the literal sense.

**Fix:** route the manifest through the same checked `write_all` + `flush` sink, returning
`ExitCode::FAILURE`.

#### H3 — `CorpusSpec` admits panicking and garbage states; validation lives only in the CLI

`crates/bench/src/lib.rs:34-63`. All fields `pub`, no constructor, so the library and CLI surfaces
have different contracts:

- `shards: 0` → `shard_range()` divides by zero and **panics**. Clippy cannot catch this; integer
  division is not a `missing_panics_doc` trigger.
- `shard >= shards` → `shard * base` overflows (dev panic / release wrap → `start > end` → an empty
  range emitted **silently**).
- `iris: 0` → `skewed_entity` returns 0 for every slot; the entire corpus is one subject, no error.

**Fix (decided):** `CorpusSpec::new(seed, quads, iris, shard, shards) -> Result<Self, SpecError>`
with private fields; `shard_range()` becomes infallible by construction; `parse_args` becomes a thin
caller so the invariant has exactly one home.

#### H4 — no test drives the production surface

`cargo test -p purrdf-bench` reports **"Running unittests src/main.rs … running 0 tests"**.
`parse_args` — the only new logic with branching — has zero coverage, which is precisely why H1
shipped past authoring, review, and `make check`. The four lib tests use a private `corpus()` helper
that re-implements the emit loop rather than invoking the binary.

This contradicts the repository's own stated principle, `crates/cli/tests/shex_cli.rs:2`:

> "…driving the built binary (`env!("CARGO_BIN_EXE_…")`) — **never the library** — so every
> assertion pins the shipped [surface]."

`purrdf-envelope-probe` also ships untested, but an existing omission does not authorize repeating
it under `.goals` GREENFIELD / NO COMPROMISES.

**Fix:** add `crates/bench/tests/corpus_cli.rs` on the established idiom (`const BENCH: &str =
env!("CARGO_BIN_EXE_bench-corpus")` + a `run()` helper returning `(exit code, stdout, stderr)`, per
`crates/cli/tests/smoke_cli.rs:20-32`), covering: cross-process determinism; shard stitching
(`--shards 1`, uneven division, final shard, `shards > quads`); manifest parsed and field-asserted
including `shard_rows == shard_range()`; the full hard-fail matrix from H1/M7 asserting exit code
**and** stderr, never merely "non-zero"; and the over-refusal counter-cases.

#### H5 — the tests that should have caught C1–C3 assert nothing falsifiable, and the golden protects the bugs

This is the root cause. Fixing C1–C3 without this leaves the next edit free to break them again.

- `every_row_parses_as_strict_nquads` (`lib.rs:297`) asserts only `rdf_row_count() > 0`. A generator
  emitting 2,000 copies of one trivial row passes. No round-trip equality (minted IRI == parsed
  IRI), no assertion that graph names, `@zh` tags, or raw-Han IRIs survived. This is the exact
  silent-drop shape `CLAUDE.md` names.
- `class_mix_is_exercised_and_indexed_minting_is_stable` (`lib.rs:305`) asserts each class appears
  **at least once**. A mix of 990/5/2/2/1 stays green, so `CLASS_MIX_PER_MILLE` — "the
  anti-compressibility contract" — is decorative rather than load-bearing.
- The determinism half of `generation_is_deterministic…` (`lib.rs:282`) compares a pure function to
  itself, same process, same inputs. With no RNG, no interior mutability, and no iteration order it
  **cannot fail**. (Its *stitch* half is genuinely strong — keep it.)
- `golden_digest_pins_the_profile` (`lib.rs:319`) is a real byte pin, but it pins the
  **undocumented** behaviour — the 4/7 ratio, the single-`%` tail, the 628-byte class — and
  therefore actively shields C2/C3 from discovery. A digest is only as good as the property tests
  standing beside it.

**Fix:** one property test per documented claim, measured from generated output: per-mille shares to
a stated tolerance; the 2/3 : 1/3 ratio; per-class byte-length bands; percent-encoding validity and
case; the 1/6 graph fraction *and its independence from object kind*; 36-digit width;
`CLASS_MIX_PER_MILLE` sums to 1000; `shard_range` unit cases; `skewed_entity` at and above 2³².

#### H6 — producer with no consumer, and two claims that cannot be verified where the disk does not exist

Exhaustive search for `bench-corpus|purrdf-bench|scale-corpus|purrdf-scale-mixed` across
`*.rs *.md *.sh *.py *.yaml *.toml` returns only the crate itself plus three bookkeeping lines.
Specifically: **`docs/BENCHMARKS.md` — the document that governs every performance claim — is
untouched and does not know the generator exists**; `make bench` omits `-p purrdf-bench`;
`.github/workflows/benchmarks.yaml` does not run it; and the commit message's "behind every
**B-series** capacity claim" names a B-series that appears **nowhere in the tree** (B2 lives only in
untracked `ROADMAP-xuan.md`, so no reviewer can verify it). The PR also claims "10B-scale corpora
generate in parallel with no coordination" while shipping no driver and no shard-file naming
convention.

**Constraint (from the user): CI runners do not have the disk for a 10B corpus** (~1.9 TB at the
measured 190.7 B/quad). This shapes the fix rather than blocking it — and note the generator already
streams through a `BufWriter` to stdout, so a piped consumer never needs the corpus to land at all.

**Fix — disk-bounded, streaming-first:**

1. Ship the **shard-parallel driver** (a `make` target) that is *pipe-first*: it streams
   `bench-corpus … --shard k --shards n` to a consumer by default, and only writes files under an
   explicit shard-file naming convention when an operator asks. A 10B run therefore never
   materializes 1.9 TB.
2. Add a **`docs/BENCHMARKS.md` scale-lane section** defining the parameters, the manifest, the
   output digest, what is claimed — and **the disk/stream requirement**, so the 10B operating point
   is documented as operator-driven and off-CI.
3. **CI runs a small parameterized smoke only** (order 10⁶ quads, streamed, nothing retained) —
   enough to catch regressions, sized for the runners that exist.
4. **Restate the two unverifiable claims honestly.** "10B-scale corpora generate in parallel with no
   coordination" becomes a stated property of the *algorithm* (index-pure minting ⇒ shards need no
   coordination) rather than a demonstrated capability, with the storage requirement named. Delete
   the "B-series" reference — the roadmap owns sequencing, and per `.baseline` it does not belong in
   the repo at all.

### MEDIUM

- **M1 — Dual sources of truth for the mix and the predicates.** `class_of`'s trailing
  `CLASS_MIX_PER_MILLE.len() - 1` (`lib.rs:106`) **silently absorbs** any mix that does not sum to
  1000: a mix of 900 dumps 10% of entities into "very-long" with every test green. `write_entity_iri`
  (`lib.rs:115-167`) hardcodes arms `0..3` + `_`, so a sixth entry silently routes to "very-long".
  `PREDICATES: [&str; 8]` is indexed by a hardcoded `% 8` (`lib.rs:203`). *Fix:* a const assertion
  that shares sum to 1000; a dispatch that cannot silently fall through; `PREDICATES.len()`.
- **M2 — One draw drives four correlated decisions** (decided: separate tags). `o` selects object
  kind (`o%3`, `o%7`), literal subtype (`o%5`), graph membership (`o%6`) and graph id (`o>>16`).
  Because `o≡0 mod 6` forces `o≡0 mod 3`, only **14.3%** of named-graph rows carry a literal object
  versus 42.8% corpus-wide (literal-in-graph occurs at ~1/42 instead of ~1/18). The joint
  distribution is an artifact of residue arithmetic, not of the stated design. `draw(seed, tag,
  index)` already takes a tag — use four.
- **M3 — "~512-byte very-long" is 627–630 bytes** (measured min 627, max 630, mean 629.4 = 24 +
  30×20 + index); repeated in the PR body and commit message. *Fix:* state the true figure. Prefer
  the doc fix — 628 bytes serves "length itself is the stress" equally well, at no byte cost.
- **M4 — False doc claim `lib.rs:33-34`: "the manifest digests exactly these fields".** `manifest()`
  computes **no** digest and the emitted JSON contains none. *Fix:* delete the claim or emit a digest.
- **M5 — The manifest misreports what it describes.** `class_mix_per_mille` omits that the shares are
  of the **entity space** (`class_of` is a function of the entity index; under the skew, row-level
  shares differ), and nothing reports the **achieved** distinct-entity count — measured 77,872 from
  `--iris 100000 --quads 200000`, so a dictionary-size claim is unverifiable from the artifacts the
  tool produces. *Fix:* name the basis in the key, and emit the achieved distinct count (derivable
  exactly from the shard range, no set needed).
- **M6 — Process-flow prose in shipped artifacts** (`.baseline` violation). `crates/bench/README.md:22-23`
  ("will join this crate once their upstream artifacts clear the acquisition/licensing review") and
  `crates/bench/Cargo.toml:6` (`description` claims the crate is "the home of the comparison-workload
  drivers" — shipped metadata describing what does not exist). Note the two audits **disagree** on
  whether a pre-existing issue covers the drivers (one cites #280 "Verify and benchmark under
  LargeRDFBench"; the other searched #294–#333 and found none) — immaterial, because the promise is
  deleted either way and the roadmap owns sequencing. Also unverifiable: the PR body's "that review
  is complete" references artifacts absent from this branch.
- **M7 — No `--help`, no usage on error, duplicate flags silently last-win.** `bench-corpus --help`
  → `unknown argument "--help"`, exit 2; the usage block exists only in a `//!` comment users never
  see (`main.rs:7-13`). `--quads 10 --quads 1000000` silently takes the last. *Fix:* `--help`/`-h`
  printing usage and exiting 0; echo usage on parse error; hard-fail repeated flags.
- **M8 — Dead silent fallbacks that would substitute corpus bytes.** `unwrap_or(0)` /
  `unwrap_or('\u{4E00}')` at `lib.rs:136,137,203,220,221` are unreachable by construction (range
  `0x4E00..=0x9FA4`, all valid scalar values), but each would **silently emit a different corpus
  byte** if a derivation changed. *Fix:* provably-infallible expressions, or `expect` with a reason.
- **M9 — `manifest()` has zero test coverage** (`lib.rs:239-258`), despite being "the JSON parameter
  record a capture stores". *Fix:* parse it with the project's own JSON reader and assert every field.
- **M10 — Literal subtype shares undocumented and lopsided.** `lib.rs:196` says "plain ASCII, `@zh`
  Han, or long text" with no shares; measured 19.95% / 19.81% / **60.2%** (the `_` arm covers
  `o%5 ∈ {2,3,4}`). *Fix:* pin the shares in a const table beside `CLASS_MIX_PER_MILLE`.
- **M11 — Portability claim ungated.** `lib.rs:20` "The library half is portable" rests on a one-off
  manual build recorded in the PR body; `make wasm`'s allowlist omits the crate (policy-consistent
  for `publish = false`), so nothing catches a regression. *Fix:* add the lib to the allowlist, or
  soften the claim to what is gated.
- **M12 — Two deterministic corpus generators, no cross-reference.** `purrdf_rdf::gts_fixtures` is
  described as "the shared deterministic keystone generator" (`crates/envelope-probe/src/lib.rs:19`);
  this adds a second with an incompatible derivation and no pointer in either direction. *Fix:* state
  in both crates' docs why two profiles exist and which regime each owns.

### LOW

- **L1 — `--quads N` counts *lines*, not distinct quads** (measured 2 duplicate rows in 300k; the
  golden test's own message admits it). The manifest should carry the emitted-line count and any
  claim must name which number it is.
- **L2 — Root `README.md` crate table omits `purrdf-bench`** — but it already omits
  `purrdf-envelope-probe`, so this follows precedent. Decide whether the table is "all crates" or
  "the library surface" and make both tooling crates obey it.
- **L3 — No `CHANGELOG.md` entry.** #322 (HNSW) added one; #329 (envelope-probe — the closest
  precedent, an unpublished tooling crate) added none. Precedent-consistent; confirm intent.

---

## Execution order

Each gap: implement → commit → push → post summary to the PR → wiring-verify. **No batching.**

1. **H3** (`CorpusSpec::new`) — lands first so the CLI work can delegate to it.
2. **H1**, **H2**, **M7** (CLI hard-fails, manifest sink, `--help`/duplicates) — each with its
   over-refusal counter-check.
3. **H4** (binary-driving tests) — lands before the byte-changing work so steps 4–5 are pinned as
   they change.
4. **C1** (full-64-bit skew), **C2** (ratio guard), **C3** (percent-encoding), **C4** (RDF 1.2 +
   typed literals + blank nodes), **M2** (separate tags), **M1** (single source of truth) — all
   byte-changing, so they land together and the golden digest is re-pinned **once**, keeping profile
   id `…-v1` per the recorded decision.
5. **H5** (property tests per claim) — after the bytes settle, so every documented claim is measured
   against the final output.
6. **H6** (driver + `docs/BENCHMARKS.md` scale lane + CI smoke + honest claim wording).
7. **M3**–**M6**, **M8**–**M12**, **L1**–**L3** (documentation, manifest accuracy, dead fallbacks).

---

## Verification

From the worktree `/home/paudley/Active/purrdf/.worktrees/bench-corpus`.

**The gate** — unpiped, so the exit code is `make`'s own and never `tail`'s:

```sh
make check
```

**Production-surface demonstrations** (the CLI, not the library):

```sh
# C1 — the headline: ten billion IRIs must not panic in debug, and must actually be reachable.
cargo run -p purrdf-bench -- --quads 100000 --iris 10000000000 --out /tmp/b2.nq
cargo run -p purrdf-bench --release -- --quads 100000 --iris 10000000000 --out /tmp/b2r.nq
# assert: max emitted entity index approaches --iris (NOT capped near 4.29e9) and every index < iris

# Cross-process determinism.
cargo run -p purrdf-bench --release -- --quads 200000 --iris 50000 --out /tmp/a.nq
cargo run -p purrdf-bench --release -- --quads 200000 --iris 50000 --out /tmp/b.nq
cmp /tmp/a.nq /tmp/b.nq

# Shard stitching, shards NOT dividing quads evenly.
for k in 0 1 2 3 4 5 6; do
  cargo run -p purrdf-bench --release -- --quads 200000 --iris 50000 --shard $k --shards 7
done > /tmp/stitched.nq
cmp /tmp/a.nq /tmp/stitched.nq

# H1/H2/M7 — hard-fail matrix: assert exit code AND stderr.
cargo run -p purrdf-bench -- --quads 10 --iris 10 --out                      ; echo "exit=$?"  # 2
cargo run -p purrdf-bench -- --quads 10 --iris 10 --out --manifest           ; echo "exit=$?"  # 2
cargo run -p purrdf-bench -- --quads 10 --iris 10 --out /tmp/m.json --manifest            # must honour --out
cargo run -p purrdf-bench -- --quads 10 --iris 10 --quads 20                 ; echo "exit=$?"  # 2
cargo run -p purrdf-bench -- --help                                          ; echo "exit=$?"  # 0
cargo run -p purrdf-bench -- --quads 5 --iris 10 --manifest > /dev/full      ; echo "exit=$?"  # 1, not 101

# Over-refusal counter-checks — these MUST still succeed.
cargo run -p purrdf-bench -- --quads 10 --iris 10 --out /tmp/ok.nq
cargo run -p purrdf-bench -- --quads 10 --iris 10 --out /tmp/-weird.nq
cargo run -p purrdf-bench -- --quads 1 --iris 1
cargo run -p purrdf-bench -- --quads 3 --shards 8 --shard 7 --iris 10
```

**Measure every documented claim against generated bytes** — measured, never restated: per-mille
class shares (entity space); entity:literal ratio (must be 2/3 : 1/3); named-graph fraction (1/6)
**and its independence from object kind** (literal-in-graph must approach ~1/18, not 1/42); the
`%XX%xx` shape, case, and that the escaped bytes are not RFC 3986 unreserved; 36-digit width; the
very-long byte length against M3's corrected figure; and non-zero measured counts for triple
terms/reifiers, typed literals, and blank nodes (C4).

**H6 smoke, sized for the runners that exist** — streamed, nothing retained:

```sh
cargo run -p purrdf-bench --release -- --quads 1000000 --iris 500000 | wc -c
```

**Exit gate:** re-run the completion-adversary (STAGE=2) against the final HEAD. The branch is not
ready for `/stage3` until every criterion returns `DEMONSTRATED-IN-PRODUCTION`.

---

## Process notes

- **No deferrals.** Every gap above is fixed in this stage. Nothing is filed to a follow-up and
  nothing is recorded in `.deficiencies`, which holds only its notice and marker — as it should.
- Where the two audits overlapped, the adversary's executed measurement wins over static inference;
  where they disagreed (M6's tracking issue) the disagreement is immaterial to the fix. One static
  finding (clippy at `lib.rs:227`) was **disproved by running clippy** and is recorded as cleared so
  no one "fixes" working code.

paudley:
## Stage 2 remediation — gap H3 closed (`961a78b6`)

**Gap (HIGH):** `CorpusSpec` carried five public fields and no constructor, so every validation
lived in `parse_args` and the library surface had a different contract from the CLI. Three bad
states were reachable directly from library code:

| State | Observed behaviour |
| --- | --- |
| `shards: 0` | divide-by-zero **panic** in `shard_range()` |
| `shard >= shards` | `shard * base` overflows (dev/test carry `overflow-checks = true`) or yields `start > end` → an inverted or out-of-range slot range, **silently** |
| `iris: 0` | every row collapses onto entity index 0, no error |

**Fix:** the invariant now has exactly one home. Fields are private behind
`CorpusSpec::new(seed, quads, iris, shard, shards) -> Result<Self, SpecError>`; `shard_range()`
is infallible by construction and documents why; `parse_args` accumulates raw values and
delegates, so the duplicated checks are deleted. `SpecError` carries a `core::fmt::Display` impl
and deliberately **no** `std::error::Error`, keeping the library half free of `std` for wasm32.

**Verification:**

- `cargo test -p purrdf-bench --locked` → **11 passed, 0 failed**, including
  `tests::golden_digest_pins_the_profile ... ok` — **unchanged**.
- `cargo clippy -p purrdf-bench --all-targets --locked` → clean (workspace pedantic + nursery).
- `cargo fmt --all --check` → clean.
- **Zero corpus-byte change confirmed by diff inspection**, not by assertion: no `+`/`-` line
  touches `CORPIS_PROFILE_ID`, the FNV constants, `write_row`, `write_entity_iri`, `class_of`,
  `draw`, `skewed_entity`, `splitmix64`, `PREDICATES`, or `CLASS_MIX_PER_MILLE`. The only line
  mentioning the profile id is the test module's import list gaining `SpecError`.

**Tests added (8):** the four refusals asserted by *specific* `SpecError` variant rather than
`is_err()`, each paired with a neighbouring **valid** input that must still succeed — the
over-refusal counter-check `CLAUDE.md` mandates — plus `shard_range` at `shards == 1`, at uneven
division (`quads: 50_000, shards: 7` → last shard `[42_852, 50_000)`), and where shards outnumber
quads (`quads: 3, shards: 8`, stitching exactly with no gap and no overlap).

---

### Correction to the posted analysis (C1, defect (b))

Re-deriving in exact arithmetic, one claim in the analysis above was imprecise and it changes the
test C1 needs. The analysis said widening to `u128` "does not make ten billion reachable". The
**range** is in fact not capped: with the overflow removed the current formula's maximum does
reach ~`iris` (9,999,803,244 at `iris = 1e10`). The measured max of 4,294,869,752 is the release
**wrap** — defect (a) — not a range ceiling.

Defect (b) is real but narrower: `hi = r >> 32` leaves `biased` at most 2^32 possible values, so
the map has at most 2^32 **distinct** outputs and ten billion distinct IRIs stay unmintable. That
is invisible to a max-value assertion. The discriminator is a **collision** test: at
`iris = u64::MAX` over 200,000 slots the fixed derivation yields 200,000 distinct entities while
the current one yields 199,972 — colliding precisely because it funnels through 32 bits. C1 will
carry one test per defect.

paudley:
## Stage 2 remediation — gaps H1, H2, M7 closed

**H1 (HIGH) — `--out` diverged silently, three ways.** Observed before the fix:

| Invocation | Before |
| --- | --- |
| `--quads 5 --iris 10 --out` | exit **0**, 741 bytes of rows to **stdout**, stderr empty |
| `--quads 5 --iris 10 --out --manifest` | exit **0**, created a file literally named `--manifest` holding rows |
| `--quads 5 --iris 10 --out /tmp/x.nq --manifest` | exit **0**, manifest to stdout, `--out` **silently discarded**, file never created |

A shard-parallel driver would have produced zero files and a green exit.

**H2 (HIGH) — one operation, two error disciplines.** The manifest went through `print!`, which
*panics* on write failure, while the row path twenty lines below reported the identical failure
cleanly and returned `ExitCode::FAILURE`.

**M7 (MEDIUM)** — `--help` returned `unknown argument "--help"` (exit 2); usage existed only in a
`//!` comment no user sees; duplicate flags silently took the last value.

### Fix

The sink is now built **once, before** the manifest-or-rows branch, and both paths write through a
single checked `write_all` + `flush` helper. That one restructuring closes H2 *and* H1's third case
together and leaves exactly one error discipline for one operation. `--out` requires an operand and
rejects only the nine recognized flag tokens. Usage lives in one `const USAGE`, printed on
`--help`/`-h` and echoed after every parse error. Repeated flags are an error.

### Verification — executed against the built binary, not `cargo run`

```
--quads 10 --iris 10 --out            exit=2  bench-corpus: --out requires a path
--quads 10 --iris 10 --out --manifest exit=2  --out requires a path, not the recognized flag "--manifest"
--quads 10 --iris 10 --quads 20       exit=2  --quads may not be given more than once
--bogus                               exit=2  unknown argument "--bogus"
                                              (each also echoes USAGE to stderr)
--help / -h                           exit=0  893 bytes of usage on stdout, stderr empty
--quads 5 --iris 10 --manifest >/dev/full
                                      exit=1  write failed: No space left on device  (was exit 101, panic)
--quads 10 --iris 10 --out /tmp/m.json --manifest
                                      exit=0  241-byte manifest written to the file; stdout empty
```

No stray `--manifest` file is created anywhere.

**Over-refusal counter-checks (mandatory per `CLAUDE.md` — a refusal is a claim too).** Rejecting
every operand beginning with `-` would have traded a silent-divergence bug for its mirror image, so
each must still succeed:

```
--out /tmp/-weird.nq   exit=0  file written   <- leading dash in a PATH is legal
--out /tmp/rel.nq      exit=0  file written
--quads 1 --iris 1     exit=0  1 row
--quads 3 --iris 10 --shard 7 --shards 8   exit=0  3 rows (last shard absorbs the remainder)
```

### Gate state

- `cargo test -p purrdf-bench --locked` → **21 passed, 0 failed** (11 lib + 10 new against
  `parse_args_from`; the binary target previously ran **0** tests). `golden_digest_pins_the_profile
  ... ok`, **unchanged**.
- `cargo clippy -p purrdf-bench --all-targets --locked` → no findings in `crates/bench`.
- `cargo fmt --all --check` → clean.
- `crates/bench/src/lib.rs` is **untouched** in this commit — `git diff --stat` shows `main.rs`
  only, so `CORPUS_PROFILE_ID` and the byte-level digest cannot have moved.

Exit codes, stream contents and write-failure behaviour are pinned next by the integration tests
that drive the built binary (gap H4).

paudley:
## Stage 2 remediation — gap H4 closed

**Gap (HIGH):** `crates/bench/` had no `tests/` directory. `cargo test -p purrdf-bench` reported
`Running unittests src/main.rs … running 0 tests`. Nothing executed the shipped binary, so its
exit codes, stream contents and file side effects were unpinned — which is precisely how three
silent-success `--out` paths cleared authoring, review and `make check`. The unit tests call the
library, and the library was never wrong.

The repository already states the rule, in `crates/cli/tests/shex_cli.rs:2`:

> "…driving the built binary (`env!("CARGO_BIN_EXE_…")`) — **never the library** — so every
> assertion pins the shipped [surface]."

`purrdf-envelope-probe` also ships untested, but an existing omission does not authorize
repeating it under `.goals` GREENFIELD / NO COMPROMISES.

### Fix — `crates/bench/tests/corpus_cli.rs`, 16 tests over `env!("CARGO_BIN_EXE_bench-corpus")`

- **Determinism across a real process boundary** — two separate invocations, byte-compared. The
  existing in-process test compares a pure function to itself and, with no RNG, no interior
  mutability and no iteration order, *cannot fail*; only a process boundary tests the claim.
- **Shard stitching through the CLI** — `--shards 1`, uneven division (`--quads 5000 --shards 7`),
  the final shard checked independently against `shard_range()`, and `shards > quads`
  (`--quads 3 --shards 8`). No gap, no overlap.
- **Manifest** — every field asserted, including `shard_rows` against the library's
  `shard_range()`, parsed with a real JSON parser rather than string matching; plus
  `--manifest --out` writing to the file with stdout empty.
- **A 13-row hard-fail matrix** asserting the exact exit code **and** the exact diagnostic —
  never merely "non-zero".
- **Six over-refusal counter-checks**, so a later tightening cannot silently convert these
  refusals into over-refusals: a leading-dash filename, a relative path, `--quads 1 --iris 1`,
  `--shards 1`, the last legal shard, and `--iris 4294967295`.

### The round-trip assertion pins an equality, not an inequality

The first draft of this suite ended with `dataset.rdf_row_count() <= line_count`. That was
rejected before commit: it is satisfied by losing 3,999 of 4,000 rows — the same silent-drop shape
`CLAUDE.md` names, and barely stronger than the `> 0` it replaced.

The exact expected value is computable, because the generator is deterministic:

```
--quads 4000 --iris 50 --seed 31337
  emitted lines           4000
  distinct emitted lines  3791
  dataset.rdf_row_count() 3791   <- exact equality, corroborated independently with `sort -u`
```

The dataset holds exactly one row per **distinct** emitted line: identical `(s, p, o, g)` rows
minted at different slots collapse under set semantics, and **nothing else may be lost**. A second
assertion guards the distinct fraction against wholesale collapse, with the bound taken from the
measured 94.8% rather than picked arbitrarily.

### Gate state

- `cargo test -p purrdf-bench --locked` → **37 passed, 0 failed** across three targets
  (11 lib + 10 bin + 16 integration; the integration target did not exist before).
  `golden_digest_pins_the_profile ... ok`, **unchanged**.
- `cargo clippy -p purrdf-bench --all-targets --locked` → clean (`--all-targets` covers the new
  tests under the workspace pedantic + nursery table).
- `cargo fmt --all --check` → clean.
- `git diff --stat -- crates/bench/src/` → **empty**. This is a test-only commit; the profile id
  and byte-level digest cannot have moved.

`serde_json` joins as a **dev**-dependency only, matching the precedent at
`crates/cli/Cargo.toml:92-93`, so the generator itself keeps zero runtime dependencies.

paudley:
## Stage 2 remediation — gaps C1, C2, C3, C4, M1, M2, M10 closed in one commit

These landed together because each moves corpus bytes; the golden digest is re-pinned **exactly
once**. Profile id stays `purrdf-scale-mixed-v1` per the recorded decision.

### C1 (CRITICAL) — the `--iris` knob was broken at the number it exists for

Two defects. The scaling multiply overflowed `u64` above 2^32 (dev/test carry
`overflow-checks = true`, so they **panicked**; release wrapped in silence); and squaring only the
high half of the draw capped the map at 2^32 **distinct** outputs regardless of row count.

Squaring the full draw in 128-bit fixed point fixes both. Verified on the production surface, in a
**debug** build — the invocation that previously aborted at exit 101:

```
bench-corpus --quads 200000 --iris 10000000000 --seed 31337
  exit=0   200000 lines   max entity index emitted = 9,999,920,083
```

Ten billion distinct IRIs now generate. The skew shape is unchanged (sqrt-CDF to four places).
The discriminator test for the distinct-count ceiling: at `iris = u64::MAX` over 200,000 slots the
fixed derivation yields **200,000/200,000** distinct, the old one **199,968**.

### C2 + C4 + M1 + M10 — one table replaces four defects

A guard on a seventh made the documented 2/3 entity edges actually **4/7**; literal subtypes split
20/20/60 with no share ever written down; and a roll past the end of a mix table was **silently
absorbed** into the last class, so a table summing to 900 would have dumped a tenth of the entity
space into `very-long` with every test green.

`ROW_MIX_PER_MILLE` now pins every row kind, a `const` assertion proves both tables sum to 1000,
and a roll past the end is `unreachable!()` rather than absorbed. Measured over 200,000 rows:

| row kind | pinned | measured |
| --- | --- | --- |
| entity-edge | 500 | 497.83 |
| plain-literal | 150 | 150.93 |
| zh-literal | 100 | 100.41 |
| typed-literal | 100 | 100.46 |
| long-text-literal | 50 | 49.90 |
| reified | 60 | 60.35 |
| blank-node | 40 | 40.13 |

### C4 (CRITICAL) — the RDF 1.2 corpus now contains RDF 1.2

Measured on generated output, before → after:

| | before | after |
| --- | --- | --- |
| triple terms / reifier rows | 0 / 0 | **12,070 / 12,070** |
| typed literals (`^^`) | 0 | **20,091** (all four XSD datatypes) |
| blank nodes | 0 | **8,026** (4,095 subject + 3,931 object family) |

Reifier rows use the exact shape the project's own N-Quads reader already admits
(`crates/rdf/src/native_codecs/parse.rs:818`). `rdf:reifies` and the `xsd:` datatypes are W3C
vocabulary — required by the specs, not minted; all corpus data stays under `example.org`.

### C3 (HIGH) — the percent-encoded tail now stresses something

It advertised a *mixed-case percent-encoded* tail and emitted one escape, uppercase, over `A`–`Z`.
Percent-encoding an **unreserved** octet is normalization-equivalent (RFC 3986 §6.2.2.2), so a
conformant normalizer decoded it away and the class stressed nothing at all. Now two escapes over
reserved octets, one upper and one lower — verified on output:

```
sample: %3F%3b      12 distinct escaped octets, ALL reserved
        unreserved octets escaped: 0        lowercase hex present: yes
```

### M2 (MEDIUM) — four decisions no longer ride on one draw

Because `o ≡ 0 mod 6` forced `o ≡ 0 mod 3`, a row in a named graph could scarcely hold a literal.
Measured: **14.3%** against 42.8% corpus-wide. Each decision now draws from its own tagged stream:

```
literal share corpus-wide  0.40170
literal share in graphs    0.40355     <- 1.9 per mille apart
named-graph fraction       0.16639     (target 0.16667)
```

### The reader round trip stayed an equality

Adding reifier rows breaks the equality H4 pinned, because **reifier rows are not base quads**. It
was **re-derived, not relaxed** — each term pinned separately so a drop cannot hide by moving rows
between tables:

```
quad_count() + reifiers().count() + annotations().count() == distinct emitted lines
  at --quads 4000 --iris 50 --seed 31337:  3588 + 230 + 0 == 3818 distinct of 4000 lines
```

`annotations().count() == 0` holds by construction: reifier IRIs live under a path segment no
entity-IRI shape can produce, so no row can annotate one.

### Gate state

- `cargo test -p purrdf-bench --locked` → **53 passed, 0 failed** (25 lib + 10 bin + 18
  integration). The branch started with 4 tests.
- `cargo clippy -p purrdf-bench --all-targets --locked` → no `crates/bench` findings.
- `cargo fmt --all --check` → clean.
- Golden digest `0xB6E8_8723_8FD8_AB29` → `0xEA2E_E654_BA6F_44D3`, re-pinned **once**; line count
  still exactly 2,000 (one slot emits one line for every row kind, so shard stitching stays
  byte-exact).

paudley:
## Stage 2 remediation — gap H5 (remainder) and gap M3 closed

### M3 (MEDIUM) — the "~512 byte" claim was never measured

The very-long class is documented at ~512 bytes and emits **627–630**. Measured per-class subject
IRI lengths over 200,000 rows:

| class | measured bytes |
| --- | --- |
| plain | 23–27 |
| numeric-long | 58 (always exactly) |
| chinese | 41–45 |
| irregular | 54–58 |
| **very-long** | **627–630** |
| reifier | 38–42 |

The **documentation** is corrected, not the code. Changing the segment count to reach 512 would
move corpus bytes and force a second digest re-pin for a round number that buys nothing — 628
bytes stresses arena growth, bucket boundaries and wire framing exactly as well. (The README turned
out never to state a byte figure, so it needed no change.)

### H5 (HIGH, remainder) — four claims were carried by prose alone

Earlier commits pinned the row mix, class mix, RDF 1.2 non-vacuity, percent-encoding shape,
named-graph independence, skew shape and typed-literal lexical forms. These were what remained:

1. **Per-class byte-length bands**, taken from measurement with headroom — explicitly *the test
   that would have caught M3*.
2. **The 36-digit numeric width and magnitude** — exactly 36 digits, leading zeros preserved
   verbatim (e.g. `065978635128795439000000000000002174`), and the value compared **as a digit
   string** against `u64::MAX` and 2^53. Parsing it into a `u64` to prove it exceeds `u64::MAX`
   would overflow by construction.
3. **The named-graph fraction** — **no new test**: `named_graph_membership_is_independent_of_row_kind`
   already asserts the one-sixth share within tolerance. Recorded rather than duplicated.
4. **Term-level round-trip fidelity.** The existing round trip pinned a *count* identity, which a
   corpus of 2,000 identical rows would also satisfy. It now also pins that the terms survive
   parsing: a raw-Han IRI arrives un-escaped, a `@zh` literal keeps both its tag and its Han
   lexical form, a named-graph name resolves as a graph term, and a reserved-octet escape is
   still an escape.

### The C3 fix is confirmed to do what it was for

Item 4 was the one most likely to surprise, so it was reported as observed rather than assumed:
the strict reader **preserves** `%3F%3b` verbatim rather than decoding it —

```
https://s60b8.example.org/x/%3F%3b/K3993~ee0ad06aae2d60b8
  parsed TermRef::Iri still contains  %3F%3b   (not the decoded "?;")
```

That is correct spec behaviour — percent-escapes are part of an IRIREF's lexical identity — and it
is precisely why the class had to stop escaping *unreserved* octets, which a normalizer is entitled
to decode away.

### Gate state

- `cargo test -p purrdf-bench --locked` → **56 passed, 0 failed** (28 lib + 10 bin + 18
  integration).
- `golden_digest_pins_the_profile` → **unchanged** at `0xEA2E_E654_BA6F_44D3`. This commit is
  documentation and tests only; `git diff` confirms no derivation function, constant or mix table
  was touched.
- `cargo clippy -p purrdf-bench --all-targets --locked` → clean. `cargo fmt --all --check` → clean.

paudley:
## Stage 2 remediation — gaps H6 and M11 closed

### H6 (HIGH) — a producer with no consumer, and two claims nothing could verify

A repo-wide search for `bench-corpus|purrdf-bench|scale-corpus|purrdf-scale-mixed` outside
`crates/bench/` returned only bookkeeping lines. In particular **`docs/BENCHMARKS.md` — the
document that governs every performance claim — did not know the generator existed**; `make bench`
omitted it and `benchmarks.yaml` never ran it. A corpus nobody can generate on purpose is not
evidence.

**Design constraint (from the maintainer): CI runners do not have the disk for a full-scale
corpus.** That shaped the fix rather than blocking it. Measured density: **177.4 bytes/row**, so
10¹⁰ rows ≈ **1.77 TB** — at full scale the corpus must move *through* a consumer, not land on one.

**1. A pipe-first shard driver.** `scripts/scale-corpus.sh` + `make scale-corpus`, streaming by
default and retaining nothing; writing shard files is opt-in. Verified running for real
(`make` exit 0, listed in `make help`):

```
scale-corpus: profile=purrdf-scale-mixed-v1 mode=stream shards=4 (retaining nothing)
shard 1/4 rows=50000 bytes=8734207 bytes_per_row=174.7 sha256=a877bf7c...
total rows=200000 bytes=34982935 bytes_per_row=174.9
```

Shard names carry the profile, the parameters and a zero-padded index, so sorted concatenation
reproduces a whole run **byte for byte** — verified independently at 500k rows over 7 shards:

```
whole unsharded run                      sha256 = d22b00ae028fabb7078f487ba041d56c...
files mode, cat of sorted shard files    sha256 = d22b00ae028fabb7078f487ba041d56c...
  purrdf-scale-mixed-v1.seed31337.quads500000.iris50000.shard-00000-of-00007.nq
```

Any failing shard fails the run (proven with a read-only output dir → `exit=1`).

**2. `docs/BENCHMARKS.md` gains the lane** as a third layer beside the criterion suites and the
Python harness — parameters, the manifest-plus-digest a capture records, the density arithmetic
behind the terabyte figure, and what the profile does and does not cover.

**3. CI smokes the lane** at 10⁶ rows, streamed through a digest, **retaining not one byte**, placed
*before* the two-hour bench steps so a broken generator reports in a minute rather than after two
hours.

**4. The two unverifiable claims are now honest.** "B-series" appears in **zero tracked files**
(`git grep` confirms) — it existed only in a commit message, and the roadmap owns sequencing.
The parallel-generation claim is restated as a property of the *algorithm* — index-pure minting
lets shard *k* proceed without coordinating with shard *j*, at any scale — stated separately from
what has actually been run, with the storage requirement named.

### M11 (MEDIUM) — the portability claim was ungated

`lib.rs` says "The library half is portable" while `purrdf-bench` sat outside the wasm32 allowlist,
so the claim rested on a one-off manual build. It now joins that list — **and the fix went one file
further than specified, correctly**: the gating CI job carries its own copy of the crate list rather
than calling `make wasm`, so a Makefile-only change would have left the claim ungated exactly where
it counts.

Verified rather than asserted:

```
cargo build -p purrdf-bench --lib --target wasm32-unknown-unknown   exit=0
```

The `bench-corpus` binary is std-only and correctly stays out; the allowlist builds `--lib` only.

### Gate state

- `make scale-corpus` → exit 0, unpiped. `bash -n scripts/scale-corpus.sh` → parses, executable.
- `check-issue-refs.py`, `check-doc-claims.py` (114 claims), `check-brand-casing.py` → all pass.
- `cargo fmt --all --check` → clean. `cargo test -p purrdf-bench --locked` → 56 passed.
- `git diff --stat -- crates/bench/src crates/bench/tests` → **empty**; the golden digest does not
  move in this commit.

Two bugs were found and fixed *within* this work and are worth recording: `> /dev/stdout`
truncates when stdout is a regular file, which silently clobbered earlier output in a redirected
run; and the refactor that fixed it left a stale variable that broke `files` mode. The second was
caught only because all three modes were re-run rather than the first green result being trusted —
the earlier run had discarded output and ignored the exit status.

paudley:
## Stage 2 remediation — gaps M4, M5, M6, M8, M12, L1, L2, L3 closed

**M4** — `lib.rs` claimed "the manifest digests exactly these fields". It digests **nothing**. The
sentence now says what is true: the manifest *records* the fields; digesting the emitted bytes is
the capture's job, and a capture is the manifest **plus** that digest. A new test pins that no
manifest key is a hash.

**M5 + L1** — the manifest misdescribed its own corpus:

- `class_mix_per_mille` → **`entity_class_mix_per_mille`**. The shares are of the entity *index*
  space (`class_of` is a function of the index); `row_mix_per_mille` is a different axis entirely,
  and a reader had no way to tell which described what it measured.
- **`iris` is a target, not an achievement** — now documented as such in six places. It is the size
  of the index space slots draw from.
- **`emitted_lines` added** — arithmetic on the shard range, free at any scale. No achieved
  distinct count is computed: counting one means enumerating the corpus, which is exactly what
  streaming at full scale exists to avoid.

```json
{"profile": "purrdf-scale-mixed-v1", "seed": 1592642302, "quads": 1000000, "iris": 100000,
 "shard": 3, "shards": 16, "shard_rows": [187500, 250000], "emitted_lines": 62500,
 "entity_class_mix_per_mille": {...}, "row_mix_per_mille": {...}}
```

> **A figure from the earlier analysis did not survive checking.** The gap analysis cited "77,872
> distinct subjects at `--iris 100000 --quads 200000`". Re-measured two independent ways (a Python
> re-implementation of the derivation, and counting the emitted bytes — which agree exactly), the
> corpus names **89,569 distinct entity IRIs**, of which **75,548** ever appear in subject
> position. Derivation-side variants land at 77,943 and 77,352 — near the cited figure but not
> equal, so it was not used. `docs/BENCHMARKS.md` carries the two verified numbers and states that
> "distinct entity count" is itself ambiguous and a claim must disambiguate it.

**M6** (a `.baseline` violation) — shipped package metadata described the crate as "the home of the
comparison-workload drivers" it does not contain, and the README promised them "once their upstream
artifacts clear the acquisition/licensing review". Both deleted; the roadmap owns sequencing. A
`// GAP M3:` process marker introduced earlier in this very remediation was also removed.

**M8** — five fallbacks in the derivation (`unwrap_or(0)`, `unwrap_or('\u{4E00}')`) would have
**silently substituted a different corpus byte** rather than failing, in a generator whose entire
value is that its bytes are pinned. Each is unreachable by construction, so each now says why and
panics if it ever is not. Byte-neutral — the digest did not move.

**M12** — the workspace now has two deterministic corpus generators and neither acknowledged the
other. Each now says why the other exists: the keystone fixture is small, fixed and
statement-layer-focused for the micro-hardware envelope; the scale profile is shardable and
anti-compressible for capacity.

**L2** — the root crate table gains `purrdf-bench` *and* `purrdf-envelope-probe`, which it had also
been missing. **L3** — `CHANGELOG.md` `[Unreleased] / Added` gains entries for both, matching the
existing voice, no issue or PR numbers.

### Gate state

`cargo test -p purrdf-bench --locked` → **59 passed, 0 failed** (30 lib + 10 bin + 19 integration).
`golden_digest_pins_the_profile` unchanged at `0xEA2E_E654_BA6F_44D3`. Clippy clean, fmt clean, and
`check-issue-refs` / `check-doc-claims` / `check-brand-casing` / `check-versions` all pass.

---

## Two further gaps found — both mine, both being fixed now

Checking CI and new review feedback rather than declaring done surfaced two real defects introduced
by this remediation:

- **`--help` panics on a closed stdout.** `bench-corpus --help | head -n 0` → **exit 101**.
  The earlier commit unified the sink for the manifest and row paths and *missed the help branch* —
  the same defect, one branch over. Reported by review, reproduced on the binary.
- **The `doc` CI check is red.** `error: public documentation for write_row links to private item
  NAMED_GRAPHS`. `make check` does **not** run rustdoc — it is a separate gate — so this branch was
  green locally while CI failed. (First diagnosis attempt also misread `head`'s SIGPIPE as
  rustdoc's exit code; re-running unpiped gave the real error.)

Neither is a deferral. Both are fixed in the next commit.

paudley:
## Gate state — both gates green, run unpiped

Two more defects were found by **running the gates rather than assuming them**, and both are fixed.

### The `doc` gate was red, and `make check` cannot see it

`make check` does **not** run rustdoc — it is a separate gate
(`RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps …`). So this branch was green locally
while CI failed on:

```
error: public documentation for `write_row` links to private item `NAMED_GRAPHS`
error: could not document `purrdf-bench`
```

Fixed by **removing the link, not exporting the constant**. `CLASS_MIX_PER_MILLE` and
`ROW_MIX_PER_MILLE` are public because the manifest echoes them and a capture reads them back; the
named-graph spread reaches no contract, so widening the API to satisfy a doc link would promise
something the profile does not make good on.

> A first diagnosis attempt misread this: `cargo doc … | grep …` reported **grep's** exit status,
> not rustdoc's. Piping a gate hides its result. Re-running redirected gave the real error.

### `--help` still panicked

CodeRabbit flagged, and the binary confirmed, that `--help` remained on `print!` after the manifest
and row paths were unified onto a checked write — the same defect, one branch over:

```
before:  bench-corpus --help > /dev/full   ->  exit 101  (panic)
after:   bench-corpus --help > /dev/full   ->  exit 1    bench-corpus: write failed: No space left on device
counter-check: bench-corpus --help        ->  exit 0, 1574 bytes usage on stdout, stderr empty
```

It survived because the earlier work verified the two paths it had just changed rather than every
path that writes. A Linux-gated regression test now pins it.

### One hygiene violation, which was my own process failure

`make check` then failed at `check-issue-refs.py`: the new test's comment said "this branch",
which the gate reads as a development-process reference. **It slipped because I wrote that fix's
command ceiling without the hygiene scripts in it** — the agent had no way to run them. Reworded to
state the invariant directly, per the gate's own advice.

### Final gate results — real exit codes, not a pipe stage's

```
make check   REAL EXIT = 0
    cargo fmt --all --check
    cargo clippy --workspace --all-targets --locked -- -D warnings
    cargo check --workspace --lib --tests --locked
    every hygiene script (issue-refs, doc-claims, brand-casing, versions, publish-order, ...)
    cargo test --workspace --locked
    make wasm

make doc     REAL EXIT = 0     rustdoc errors: 0
```

`cargo test -p purrdf-bench --locked` → **60 passed, 0 failed** (30 lib + 10 bin + 20 integration).
The branch began with 4 tests and a binary target that ran **zero**.
`golden_digest_pins_the_profile` unchanged at `0xEA2E_E654_BA6F_44D3`.

paudley:
feat(bench): deterministic scale corpus and the two comparison lanes

Three never-published additions to purrdf-bench, the workspace's sixth
unpublished member: the deterministic shardable scale-corpus generator, and
the LUBM and WatDiv comparison-workload lanes that consume externally
specified suites. All three are report-only; no number here gates anything.

THE SCALE CORPUS

Every IRI is minted as a pure function of (seed, index), so shard k of n emits
exactly its slice and concatenating the shards is byte-identical to a whole
run. That is a property of index-pure minting and holds at any scale; it is
not a claim that a full-scale run was performed. At the measured density a
ten-billion-row corpus is about 1.83 TB, so full scale is an operator activity
streamed into a consumer rather than stored.

Two pinned tables are the contract and a const assertion proves each sums to
1000. Entity-IRI classes, as shares of the entity index space: plain 400,
numeric-long 200, chinese 200, irregular 150, very-long 50. Row kinds:
entity-edge 500, plain-literal 150, zh-literal 100, typed-literal 100,
long-text-literal 50, reified 60, blank-node 40.

The corpus carries RDF 1.2 -- triple terms and reifiers in the shape the
project's own N-Quads reader admits, typed literals across the XSD value
space, and blank nodes in two families -- so the statement layer, the datatype
dictionary and blank-node handling are exercised at scale rather than only in
a micro-hardware fixture. Every decision draws from its own tagged splitmix64
stream, so row kind, literal type, graph membership and graph id are
independent. All derivation is integer arithmetic: no RNG syscalls, no floats,
no iteration-order dependence. The profile purrdf-scale-mixed-v1 is pinned by
a byte-level golden digest.

THE COMPARISON WORKLOADS

Nothing is vendored. scripts/benchmark-acquire.py fetches six artifacts by
SHA-256 at use time. The LUBM generator is GPL-2.0-or-later, so it is run and
never copied into this MIT/Apache tree; univ-bench.owl and the query file
carry no licence text at all, so neither is redistributed; WatDiv is
citation-ware, so every result derived from it must cite Aluc, Hartig, Ozsu
and Daudjee, ISWC 2014, pages 197-212. A cached file whose digest stops
matching is quarantined rather than re-fetched over.

make lubm generates with the stock generator unpatched, converts RDF/XML to
N-Quads through the purrdf CLI, normalises the published queries and runs all
fourteen. The queries could not be run as published: they predate final SPARQL
and bind the ontology prefix to a 2004 draft namespace no generated dataset
uses, so unnormalised they parse, execute and answer zero. Five numbered rules
normalise them and every application is recorded before and after. Each query
carries its entailment regime from the canonical paper, because comparing an
engine that infers against one that does not is how benchmark numbers lie.

make watdiv consumes the published ten-million-triple corpus by digest.
WatDiv's generator seeds from the clock in three places and has no seed flag,
so the generation process cannot be pinned -- only a frozen output can. Its
twenty templates carry placeholders that upstream fills with that same
clock-seeded RNG; because the corpus is frozen, this lane draws from the fixed
candidate set deterministically instead and records every choice, so one seed
gives one query set byte for byte.

WHAT THE ENTAILMENT CEILINGS MEAN

Eleven of LUBM's fourteen queries need entailment, and the reasoning engine
holds fixed ceilings: MAX_STORED_FACTS is 131,072 facts against a LUBM(1)
corpus of 102,737 rows. The lane therefore reproduces the published answers
for Q1, Q2 and Q14 -- the three needing no inference -- over the full corpus,
and answers the other eleven over reduced rungs, stating in its own output
that a row count on a rung below full is not the published LUBM answer.

Two separate things are true and are not blurred here. The ceilings being
constants is deliberate and permanent: a caller-supplied budget would mean two
callers running the same program over the same input get different answers,
and the fact ceiling is sized so a saturated store stays inside a wasm32
linear memory. Materialising a closure whole in one store is a separate
choice, and that rationale says nothing about it -- a partitioned or
incremental closure would answer these queries over the full corpus without
raising any ceiling and without introducing optionality. That work is
unstarted and lives in purrdf-datalog, which this branch does not touch.

FILES

Added: crates/bench/{Cargo.toml,README.md,src/lib.rs,src/main.rs},
crates/bench/tests/{corpus_cli.rs,make_scale_corpus.rs,make_bench_lanes.rs},
scripts/{benchmark-acquire.py,lane-common.sh,scale-corpus.sh,lubm-lane.sh,
lubm-queries.py,watdiv-lane.sh,watdiv-queries.py}.

Modified: Cargo.toml and Cargo.lock (workspace member), Makefile (four lane
targets), AGENTS.md, README.md, CHANGELOG.md, docs/BENCHMARKS.md and
docs/book/src/project/releases.md (documentation and the six-unpublished-member
count), scripts/release-crates.sh (publish exclusion),
.github/workflows/{ci.yaml,benchmarks.yaml} (wasm32 allowlist, CI smoke),
.gitignore (see below), scripts/check-issue-refs.py (see below),
crates/envelope-probe/src/lib.rs (cross-reference to the second generator).

TWO FIXES BEYOND THE HEADLINE

.gitignore carried /target/ and **/target/, both directory-only patterns. A
shared build tree here is a symlink to an out-of-tree cache, which those
patterns do not match, so git ignored none of it; a clone whose owner had no
personal ignore file would have offered the GPL generator and two no-grant
files as untracked content waiting to be committed. The trailing slashes are
gone, which also covers the pre-existing probe cache.

scripts/check-issue-refs.py missed short letter-and-digit labels written with
a dash rather than a colon. Fixing
that made it fire on any comment line merely starting with a letter and
digits, which in a library for RDF means N3, F32 and F64, and on continuation
lines, so re-wrapping a paragraph could redden the gate. A label is now the
letters, the digits and the punctuation that closes them; markdown heading
levels are spared when the line says heading; and the self-test carries both
arms of every shape.

GOALS COMPLIANCE

No Cargo features. No version pinned in a member crate. The library half
builds for wasm32-unknown-unknown and is on both the Makefile and CI
allowlists; the binary is std-only and correctly stays off. Determinism is
byte-level and pinned. All corpus data IRIs are under example.org; rdf:reifies
and the xsd: datatypes are W3C vocabulary required by the specs, not minted.
Every refusal added carries a neighbouring-valid counter-check, per the
over-refusal rule.

DEFERRAL GATE

Mechanical scan over the full branch diff: two hit classes, both false
positives with evidence. One "unimplemented" is a self-test case name at
watdiv-queries.py:959 asserting that a distribution the lane does not
implement hard-fails naming it, with no fallback -- the assertion checks for
FAIL:, normal, and "does not implement". Seventeen "placeholder" hits are
WatDiv's own published template vocabulary for the %vN% tokens the
instantiator fills. The PR body and comments carry zero deferral hits. No
TODO, FIXME, stub, or future-work text is added anywhere on the branch.

CONFLICTS

origin/main merged cleanly with no conflicts. All twelve branch files, the
profile id, the golden digest, the four make targets and the wasm allowlist
entry were verified present after the merge.

IMPLEMENTATION NOTES

A shard's failure must fail the run. Both shard bodies check every fallible
command explicitly rather than relying on errexit, because a backgrounded
subshell called on the left of || has errexit disabled for its whole dynamic
extent -- which is how a failing write in files mode once exited 0, reported
bytes=0, and printed the byte-for-byte guarantee over files that did not
match. For the same reason the size loop does not ask wc to measure a path:
wc succeeds on a directory and answers zero.

A digest is a certificate, so it is published only after the thing it
certifies is proven to exist, and a failed run leaves no certificate behind.
Artifacts are identified rather than measured -- eight bytes are not a pack,
but they are not empty either.

Lane parameters never reach a shell. make exports them and the scripts read
them as environment bytes; exporting alone was not enough, because a dollar
sign is make's own syntax and was expanded before the environment existed, so
each value is frozen as the bytes the operator typed.

Directory printing is off for the whole Makefile, because a target whose
stdout is a payload cannot afford make's banner on that stream, and the
per-target form does not work: the banner is printed before target-specific
variables apply.

VERIFICATION, MEASURED ON THE SHIPPED SURFACE

Every number below was measured by running the built binary or the make targets,
not read from the manifest and not taken from a unit test.

Determinism: three separate processes, including one under a Turkish-dotless-I
locale from a different working directory, produced byte-identical output;
--out and stdout agree. The golden digest was recomputed from the bytes the
shipped binary emitted at the pinned spec and equals the constant in the
source, so the pin binds the production binary rather than a library path.

Shard identity: concatenation equals a whole run byte for byte at 1000/7,
100000/13, 99991/17 (prime over prime), 5/13 and 1/9 -- the last two being
cases where shards outnumber quads. Through make scale-corpus, both pipe mode
and the sorted cat of files mode are byte-identical to a direct whole run.

Mix, over 2,000,000 rows: every row-kind share is within 0.42 per mille of its
pinned value with zero unclassified rows; entity-class shares over the index
space sit inside binomial noise for the sample, confirmed by re-running at
five seeds. Named graphs are 166.718 per mille against a pinned sixth, and the
per-row-kind spread is 2.2 per mille, so graph membership really is
independent of row kind.

Class shapes: 723,052 numeric IRIs are exactly 36 digits with leading zeros
intact and values above 2^64; very-long measures 626-629 bytes; 430,663
irregular IRIs each carry exactly two percent-escapes, first upper-case hex
and second lower-case, decoding only to reserved octets.

RDF 1.2 is present, not nominal: 120,054 triple terms with 120,054 distinct
reifier IRIs, typed literals across four XSD datatypes each lexically valid,
and blank nodes in both families.

Ten billion: a debug build with overflow checks generated 3,000,000 rows over
a 10^10 index space without panicking, reaching a maximum index of
9,999,987,905.

Lanes: make lubm executes 14 of 14 with Q1, Q2 and Q14 returning 4, 0 and 5916
at rung full over the whole 102,737-row corpus. The normalisation is not
cosmetic -- the raw published Q1 parses, executes and returns zero through the
production CLI, while the normalised form returns its four answers. make
watdiv executes 20 of 20 for 434,748 rows over the frozen ten-million-triple
corpus, and the query set reproduces byte for byte at one seed and differs at
another.

Failure paths: fourteen distinct lane failures -- a binary that fails, one that
exits zero writing nothing, a directory in place of a binary, and a shard that
dies mid-run after the specification already verified -- every one exits
non-zero in the lane's own voice and leaves no certificate behind. In pipe mode
a mid-run failure says the stream is incomplete rather than handing a consumer
a truncated corpus.

Refusals: twenty-seven argument cases, every refusal paired with a neighbouring
valid input that still succeeds. No over-refusal was found.

Licensing: nothing non-redistributable is tracked -- no generator, no ontology,
no query file, no corpus. The acquire script proves the build tree is ignored
before it downloads a byte.

WHAT THIS LANE CANNOT DO TODAY

Eleven of LUBM's fourteen queries need entailment and the chase exceeds the
engine's fixed ceilings on the full corpus: the run reports 43,602,006 facts
observed against 131,072 permitted. Those eleven are answered over reduced
rungs and the lane prints both counts rather than hiding a missing row. Three
of fourteen published answers are reproducible over the full corpus today.
That is a real limit on this instrument's value. Resolving it means a
partitioned or incremental closure in purrdf-datalog -- work that needs no new
optionality and raises no ceiling, in a crate this branch does not touch.

PR #335, opened 2026-09-18
feat(330): incremental io::Write serialization sink

paudley:
Closes #330

## Summary

An `io::Write` path beside the byte-vector one, for both families #330 names — the
native codecs and the SPARQL-results serializers — reached from the CLI, and with the
whole-document intermediates between producer and drain removed.

The eager APIs are unchanged and are expressed **through** the streaming body, so byte
identity between the two spellings is a property of the structure rather than something
a test establishes afterwards — though it is tested anyway, across every format.

## The sink

`purrdf_core::sink::TextSink` stages into a bounded 64 KiB window over a pluggable
`ByteDrain`. Ships `WriterDrain` (adapts `io::Write`), `Measure` (a document's length
without the document), `Digest` (its SHA-256 without it), and `Tee` (both in one pass).
The drain is a trait rather than `io::Write` because several real destinations — a JS
callback, a C function pointer, a Python file object — are not `io::Write`.

It exposes no `truncate`, `pop`, `clear` or read-back. A sink cannot un-write a drained
byte, so the affordance that would silently break streaming does not exist, and
`scripts/check-serializer-rewinds.py` keeps it from reappearing on the `&mut String`
buffers the per-codec seam still passes around.

## What this does

**All 13 formats emit incrementally**, and five identical "assembled as a TREE … cannot
emit a prefix" comments are gone — the last of them had a second comment block directly
beneath it reading "Emitted through the sink."

**Three rewinds removed structurally**, each replaced by decide-before-you-emit: the
Turtle header retraction, the SRJ root-brace splice, and the XML escaper's
truncate-on-failure.

**Three refusals hoisted ahead of the first byte** (SRX's CONSTRUCT-graph refusal, TSV's
variable-name check, CSV/TSV's row-width check), each paired with a valid neighbour that
must still succeed, because hoisting a check is exactly the change that silently produces
over-refusal.

## Defects found and fixed

- **The CLI could destroy a symlinked output file's link** on a failed write, leaving the
  half-written bytes in the target — the exact loss the guard's own comment described
  preventing. It asked the descriptor, which `File::create` had already resolved through
  the link.
- **`RdfSerializer::serialize<W: Write>`** — the workspace's writer-shaped seam — built
  the whole document and `write_all`'d it. A `W: Write` parameter that buffers is worse
  than an honest `-> Vec<u8>`: the signature tells the caller the opposite of the truth.
- **The results sink had zero production callers.** Half of #330's Ask shipped
[...]

PR #351, opened 2026-09-20
The benchmark lanes: pin what a digest certifies, and share the laws that were in one lane only

paudley:
## Why

The three comparison lanes are report-only: **no gate runs a lane past step 1**, and no timing they print is asserted. That posture is right for host-dependent numbers, and it means a defect inside a lane is invisible to every gate here — the lane prints a well-formed report about nothing, and only a reader who already knew the answer can tell.

An audit of `lubm-lane.sh`, `lubm-queries.py`, `watdiv-lane.sh`, `watdiv-queries.py` and `benchmark-acquire.py` found 29 defects under that cover. **Thirteen further adversarial passes over this branch's own work** followed; the seven with tallied counts contributed 110 findings, the last of them 21. The majority across every pass were defects in the fixes for the previous round — the same shapes, committed by their own repair. All are closed here.

The dominant pattern is the one `crates/bench/tests/make_bench_lanes.rs` opens by warning about: **laws that hold in one lane and not its sibling.** The warning was accurate, repeatedly, including about this branch.

## Nothing this branch built was enforced by the gate that blocks a merge

CI does not run `make check`; it enumerates each hygiene gate as a named step. That is a reasonable choice and it creates a **second list of the same rule set, which nothing compared.** Every gate added here lived in `make check` alone — including the tracked-path gate, which exists *because* a flag-shaped artifact escaped thirty hygiene scripts, and which therefore could not block the merge that would re-commit one.

The divergence was bidirectional and partly older than this branch:

* `check-terminal-predicates.py` and its `--self-test` ran in `make check` and in **no workflow at all**;
* `check-toolchain-pin.py --self-test` ran in CI and not in `make check`.

`scripts/check-gate-parity.py` takes the *agreement between the two lists* as its subject. A gate's identity includes its arguments (`--self-test` and the bare run are different rules) and `$(MAKE)` recursions are followed, since `make check` recurses into `rdf-core-hygiene` and a gate added there was invisible. **42 gates now agree in both directions**, compared against the workflows whose `on:` block is read to include `pull_request` rather than a hardcoded list — because a hardcoded list is the same defect shape as the two lists being compared. Six gates are one-sided by design — two determinism checks needing the wasm toolchain, the book render needing mdbook, the conformance matrix needing tens of minutes, and the two `uv`-driven emitter oracles — and are registered with their reasons under a register that may only shrink and whose size is pinned, so growth is a visible edit rather than the silent one that left an earlier count of this saying "four".

## The gate caught a divergence it was not written against

Merging `main` produced the only evidence for this PR's thesis that does not come from its own
authors. PR #335 added `scripts/check-serializer-rewinds.py` to `make check` and to **no workflow at
all** — so on `main` today, nothing about serializer rewinds can block a merge. The parity gate
reported it on the merged tree, in work it knew nothing about:

```
FAIL: `make check` and the CI workflows do not run the same gates:
  `scripts/check-serializer-rewinds.py` runs in `make check` and in NO workflow at all
  `scripts/check-serializer-rewinds.py --self-test` runs in `make check` and in NO workflow at all
```

Both forms are now CI steps, per `.baseline`'s "when you see it, you own it", and the count moved
40 → 42. Every other divergence this gate has found was one it was written after; this one it found
on its own.

## What a digest certifies, and what was certifying nothing

**Collation was an unnamed input to a published digest.** `lubm-lane.sh` concatenated in `find | sort` order and published the SHA-256 as "the determinism check", naming four knobs. `sort(1)` obeys `LC_COLLATE`, and it was a fifth:

```
LC_ALL=C             University0_1.owl  University0_10.owl  University0_2.owl
LC_ALL=en_US.UTF-8   University0_10.owl University0_1.owl   University0_2.owl
```

Two hosts differing only in environment built a different corpus and published a different digest. Pinned in `lane-common.sh`. **Demonstrated:** `make lubm` under both collations yields identical corpus and query-set digests.

**A reuse stamp certified the container, not the corpus.** The WatDiv gate compared the stamp to the *tarball's* digest and otherwise required only non-emptiness. **Demonstrated:** the 1.5 GB corpus was replaced with one byte; the lane detected it, re-extracted, and returned the byte-identical original digest.

**And no digest was compared to a pin** — every one was compared to a stamp the arena itself wrote, which detects later change but cannot detect a first extraction that was already wrong, because that extraction writes the record. `WORKLOAD_PINS` holds **28 keys**, and the last trust-on-first-use in either lane is closed: `saved.txt`, the entity census, which drives every candidate pool and therefore every row count now asserted. It had been *annotated* as "a narrow gap rather than an open door" rather than closed — inside a change whose whole subject is pin-versus-stamp.

## A knob value bash could not carry was silently replaced

`lane_require_uint` validated and normalised together, and the normalisation was the second half of the same bug. Bash integers are signed 64-bit and wrap without a word:

```
9223372036854775808  -> -9223372036854775808   (a negative seed, from digits)
18446744073709551617 ->                    1
```

Both pass `^[0-9]+$`. The substituted number was handed to the generator, printed in the SUMMARY as the requested run, and folded into the pin key — so `SCALE_QUADS=18446744073709551617` published a one-quad corpus under the name of a request for eighteen quintillion. The range is now judged before any arithmetic, with `2^63-1` — the boundary itself — as the executed neighbour.

## The only engine-output assertion WatDiv had was disabled by the failure most likely to need it

The aggregate row total was asserted only when every query executed. Defensible in itself; the consequence was not. One CANNOT-EXECUTE query printed `NOT checked against a pin`, a full SUMMARY and exit 0 — so a pack that broke one query and changed the answers to others passed, where before this branch's own change it had failed.

All twenty per-query row counts are now pinned, the shape LUBM already used for Q1/Q14. A non-executing query fails against **its own** pin by name; the other nineteen are still asserted. The self-test asserts the twenty sum to the aggregate, because both sides of that comparison live in one file and nothing outside it can check them.

A missing pin and a **broken** pin lookup were also one observable — both behind `2>/dev/null`, both a silent skip, and a skip is indistinguishable from a pass. The lookup now exits 2 for "recorded nowhere"; anything else is a lane failure quoting the status.

## Defects this branch introduced and then had to fix

Recorded because the branch's whole argument is that unexamined claims rot.

- **A forgeable digest.** `lane_query_set_digest` streamed `name‖NUL‖content` unframed, so `{a: "b\0c"}` and `{a: "", b: "c"}` produced the *identical* digest. Now a manifest of per-file digests.
- **"Both holes closed" was false.** The stand-in refusal was claimed to stop a tracked artifact. Both lanes pass the query TEXT last, so `standin query … 'SELECT ?s WHERE { ?s ?p ?o } LIMIT 1'` still wrote the same 71 bytes into the CWD — under a name the tracked-path gate *accepts*. The sibling `convert_stand_in` had no guard at all. It survived because the refusal had **no observing test**.
- **A sweep that destroyed live work.** Added to tidy abandoned scratch, it deleted a concurrent run's in-flight file, whose `os.replace` then raised a bare `FileNotFoundError`.
- **A diagnosis no test reached, wrong twice.** It printed `'''path'''` (a literal-quote idiom pasted into a context that did not need it) and advised checking for something that cannot reach that arm.
- **The ordering defect, fixed in one lane and not the other.** A per-query oracle ahead of the vacuous-run guard made the vacuous diagnosis unreachable at the only knobs whose wording applies to it.
- **A false safety claim about a redirect.** "A later non-zero status is the generator's" — falsified by `gs=0; (echo hi) >>/nonexistent/log 2>&1 || gs=$?`, which sets `gs=1`.
- **Six copies of one constant** under two names across five files, two already drifted — to a quarter and a sixty-fourth of the shared size. The gate written to stop the next one found the sixth on its first run.
- **A gate whose derivation did not survive measurement.** The tracked-path gate refused leading whitespace and accepted an interior space, both on argument-splitting grounds. Measured: a glob passes all whitespace intact, and the shape that actually *splits* is the interior space it accepts.
- **A fabricated claim about a registry.** A commit message credited the issue-refs registry with absorbing two phrases it had never been shown — because `.sh` was scanned by nothing at all, in a repository whose lanes, release scripts and hygiene scripts are shell.

## Verification

Every number below is re-derived from HEAD.

- **`make check` exit 0**, unpiped, reading make's own exit code: **145** `OK:` lines, **10,334** workspace tests. Eight new offline self-tests run inside it, plus one that had only ever run in CI; none needs network.
- **`make lubm` EXIT=0** end to end — `14 of 14`, `comparable rows 3 of 14`, corpus pin `b3fbfcc8…` matched, query-set pin matched, `oracle: Q1 and Q14 match LUBM's published answers`.
- **`make watdiv` EXIT=0** end to end on a clean arena — 10,916,457 triples, `20 of 20`, `answers: every one of the 20 queries matched its recorded row count`, 434,748 rows.
- **All three answer-report branches executed**, because a reporting branch is where a skip hides:
  - *matched* — the run above;
  - *refused* — C1's pin set to 17 against an engine returning 16: `C1 returned 16 rows; the pin for watdiv.10M.seed0.rows.C1.purrdf-2.0.2 is 17`, exit non-zero, **no SUMMARY published**;
  - *no pin recorded* — `WATDIV_SEED=3`, which is the over-refusal neighbour and a DISTINGUISHING one: it exits 0 reporting `none is recorded under watdiv.10M.seed3.rows.<id>.purrdf-2.0.2`, and it produces a different query-set digest (`1d48c1aa…` vs `2fabc0ef…`) and a different total (434,661 vs 434,748) — so the seed demonstrably reaches the answers, and this branch cannot be confused with a pin that was silently skipped.
- **The census pin refuses what a stamp cannot see.** Corrupting `saved.txt` alone is caught by the arena's own stamp and self-heals by re-extracting, which is correct and is *not* what the pin is for. With the stamp edited to AGREE with the corrupted census — a first extraction that was already wrong, certified by its own record — the lane printed `reusing the extraction already stamped with this tarball's digest` and the pin still refused, naming both digests.
- **The vacuous-run diagnosis demonstrated in both directions** with a stand-in that answers every query zero rows while delegating conversion to the real binary, so the corpus still matches its pin. Before the reordering the lane died on the oracle's message; after it, on `all 14 queries executed and every one matched zero rows`.
- **Corpus substitution**: one byte changed in 1.5 GB — detected, re-extracted, identical digest restored.
- **A version banner no pin key can be built from is refused at step 2**, before anything is fetched: a stand-in reporting `purrdf 2.0.2 (a1b2c3d)` — a release that added a git hash — is rejected quoting the banner, because that one change would otherwise make all 22 version-keyed pins unreachable and every pin check report "not recorded" while exiting 0. `purrdf 2.0.2` and `purrdf 2.0.2-alpha+build.1` are both accepted.
- **A short HTTP body retries instead of quarantining.** A server declaring 100 bytes and sending 10 produces three attempts and a refusal naming the *transfer*; a complete response installs on the first attempt; a malformed `Content-Length: 1,024` is quoted verbatim rather than raising, and leaves no scratch file behind.
- **Each errno gets its own diagnosis.** An unreadable query file says the path EXISTS and that this is not a concurrent run; a missing directory says it is; anything else says the cause is not established. Previously all three claimed a concurrent run.
- `cargo test -p purrdf-bench`: **140** tests across six binaries — `lane_common_laws` 35, `make_bench_lanes` 15.
- **Mutation-checked, not trusted for going green.** Every fix above has its specific mechanism neutered and its own test confirmed red, including: restoring the concatenating digest; restoring `$((10#…))`; restoring the two-shape stand-in blocklist (which reproduces the committed artifact, `created ["SELECT ?s WHERE { ?s ?p ?o } LIMIT 1"]`); restoring the blanket `.part.*` sweep; dropping the shell scanner (prints `BLIND`); changing one per-query pin by one.

`docs/design/purrdf-bench-lane-laws.md` records the durable part — what a digest certifies, the traps, the shared-law inventory, and now two laws that were defects in *both* lanes: a scratch failure must not be published as a failure of the binary under test, and the broadest diagnosis runs first.

## The licence offer, and the gate that now holds it

The MulanPSL-2.0 merge rewrote 1707 SPDX identifiers and left the offer a human reads at two licences: 20 prose blocks, 23 badges, a gutted README section, and — found later — two crates publishing a README with no licence offer at all. `README.md` itself and 19 crate READMEs rendered on crates.io.

This is worth reading as a sequence, because it is the branch's own thesis applied to the branch:

1. The first fix corrected the 20 blocks and said outright "nothing yet gates the human-readable offer, so this is a fix without a guard". That same commit **introduced a new prose defect into 17 of them** — "at your option" left in twice, a dangling sentence on 17 crates.io pages — and nothing caught it.
2. The second said "Naming the gap did not substitute for having the guard" and built the guard. It left four more: a tri-licensed crate offering two licences, a publishable README with no offer, and two with a licence sentence but no heading.
3. The guard then had four defects of its own, including requiring an explicit `readme =` key when **cargo auto-discovers `README.md`** — so the two crates that published no offer at all were invisible to the gate written for exactly that.

`check-licenses.py` now requires every crate cargo publishes a README for to carry a licence section naming **every term** of the offer. Version-aware (a section naming `Apache-1.1` was previously accepted), scoped to the section rather than the body (a README's own SPDX header otherwise satisfied a section naming two of three), and it refuses a term named in order to withdraw it. It also no longer writes the tree it judges — the self-test had been mutating three tracked files including the root `Cargo.toml`, restoring them in a `finally` that does not run on SIGKILL.

**Still ungated, stated rather than implied:** the root `README.md`, `README.zh-Hans.md`, `bindings/python/README*.md`, and the two crates that are `publish = false`. Those are outside the "what cargo publishes to crates.io" rule the gate enforces, and extending it to them is real work, not a line.

### Known and not fixable in place

An early commit message states 10,759 tracked paths where the tree had 10,760, measured in an index state that was never committed. The live figure is **10,778** and the gate computes it rather than restating it, so nothing stale ships; the commit message cannot be corrected without rewriting pushed history, which is not worth doing for one digit.


paudley:
All three review findings were real and all three were defects in this branch's own work, so all three are fixed in `ca4eaab4` rather than argued with.

**The unreadable query file is the one that mattered.** Both loops called `lane_require_query_file` only when the path was not a regular file, so a file that exists with bytes but cannot be read skipped the guard entirely, reached `cat` under `set +e`, and became an empty query the engine was then blamed for rejecting — the exact misdiagnosis the guard was added to stop, arriving through a door I had left open. It now also requires readability and runs on every iteration. The concurrency question is still asked first, since a vanished file is still most likely a concurrent run. Proven on all four states: readable non-empty accepted; empty, unreadable and missing each refused.

**The root-creatable arena is correct and I had not considered uid 0.** `uncreatable_arena` now puts a regular file in the scratch directory and targets a child of it, so `mkdir` fails with `ENOTDIR` for every uid — the kernel checks the parent's type, not permission. Verified `ENOTDIR` directly. Applied to all four sites, including the two that predate this branch: a latent failure is not improved by having been there longer.

**The doc misdescribing its own fix** is the failure shape this branch is about, so thank you for catching it. It said stderr's *first line* goes in the detail column; `lane_flatten_detail` preserves every non-blank line and joins them, which is the whole point — the part that matters is often on a later line.

`make check` exits 0 unpiped after all three; `make_bench_lanes.rs` still 13/13.

paudley:
feat(bench): make the comparison lanes falsifiable, and gate the gates

The three benchmark comparison lanes were report-only: no gate ran one past step 1, and no
number they printed was asserted. That posture is right for host-dependent timings and it
meant a defect inside a lane was invisible to every gate — the lane printed a well-formed
report about nothing, and only a reader who already knew the answer could tell.

An audit of the five lane programs found 29 defects under that cover. Sixteen further
adversarial passes over this branch's own work found more, the majority of them defects in
the fixes for the previous round. All are closed.

## What a digest certifies now

**Collation was an unnamed input to a published digest.** `lubm-lane.sh` concatenated in
`find | sort` order and published the SHA-256 as "the determinism check", naming four knobs.
`sort(1)` obeys `LC_COLLATE`, and it was a fifth: under `LC_ALL=C`, `University0_1.owl`
sorts before `University0_10.owl`; under a UTF-8 collation they reverse. Two hosts differing
only in environment built a different corpus and published a different digest. Pinned in
`lane-common.sh`. Demonstrated: `make lubm` under both collations yields identical digests.

**A reuse stamp certified the container, not the corpus.** The WatDiv gate compared the stamp
to the tarball's digest and otherwise required only non-emptiness. Demonstrated: the 1.5 GB
corpus was replaced with one byte; the lane detected it, re-extracted, and returned the
byte-identical original digest.

**And no digest was compared to a pin** — every one was compared to a stamp the arena itself
wrote, which detects later change but cannot detect a first extraction that was already
wrong, because that extraction writes the record. `WORKLOAD_PINS` holds 28 keys, 22 of them
keyed on the binary version because they are the serializer's output. The last
trust-on-first-use is closed: `saved.txt`, the entity census, which drives every candidate
pool and therefore every row count now asserted.

**A knob value bash could not carry was silently replaced.** Bash integers are signed 64-bit
and wrap without a word: `9223372036854775808` became a NEGATIVE seed, `18446744073709551617`
became 1. Both pass `^[0-9]+$`, and the substituted number was handed to the generator,
printed as the requested run, and folded into the pin key. The range is judged before any
arithmetic, with 2^63-1 — the boundary itself — as the executed neighbour.

**The only engine-output assertion WatDiv had was disabled by the failure most likely to need
it.** The aggregate row total was asserted only when every query executed, so one
CANNOT-EXECUTE query printed "NOT checked against a pin" and exit 0 — a pack that broke one
query and changed the answers to others passed. All twenty per-query counts are pinned now;
a non-executing query fails against its own pin by name and the other nineteen still assert.

## The gates, and why they are themselves gated

`scripts/check-gate-parity.py` takes the AGREEMENT between `make check` and the pull-request
workflows as its subject. CI does not run `make check`; it enumerates each gate as a named
step, so the two lists encode one rule and nothing compared them. It found three pre-existing
divergences on its first run — `check-terminal-predicates.py` and its `--self-test` ran
locally and in no workflow at all, and `check-toolchain-pin.py --self-test` ran in CI and not
locally — and a fourth hiding behind a workflow `make` step. 42 gates now agree in both
directions. Six are one-sided by design, registered with their reasons under a register that
may only shrink and whose size is pinned.

`scripts/check-stream-chunk.py` refuses a streamed read whose chunk size is written out
rather than named. Six copies of that number lived under two names across five files, two
already drifted — to a quarter and a sixty-fourth of the shared size — and because every
chunk size produces a correct digest, nothing reported it.

`scripts/check-tracked-paths.py` refuses a tracked path that misrepresents itself to the
tools that read it: a component beginning with `-`, which a glob hands to a command as a
FLAG; a character that does not render as what it is; or a path colliding with another under
Unicode normalisation or case folding, which are the same file on APFS and HFS+. A
flag-shaped artifact had been committed at the repository root and swept past roughly thirty
hygiene scripts, because they all walk a file list rather than a glob.

`scripts/check-licenses.py` gained a first-party SPDX header rule and a published-README
rule, both reading the licence expression from `Cargo.toml` rather than restating it.

## The gate caught a divergence it was not written against

Merging `main` produced the only evidence for this change's thesis that does not come from
its author. PR #335 added `scripts/check-serializer-rewinds.py` to `make check` and to no
workflow at all, so nothing about serializer rewinds could block a merge. The parity gate
reported it on the merged tree, in work it knew nothing about, and both forms are now CI
steps — per `.baseline`'s "when you see it, you own it". The count moved 40 to 42.

## Files

Added: `scripts/check-gate-parity.py`, `scripts/check-stream-chunk.py`,
`scripts/check-tracked-paths.py`, `scripts/lane_chunk.py` (one definition of the streamed
chunk size and of fsync-before-rename), `crates/bench/tests/lane_common_laws.rs` (35 tests
proving the shared helpers directly, offline), `docs/design/purrdf-bench-lane-laws.md` (the
durable laws, with an inventory table now compared against `lane-common.sh` in both
directions by a test).

Modified: the three lanes and `lane-common.sh` (24 shared helpers, up from 15);
`benchmark-acquire.py`, `watdiv-queries.py`, `lubm-queries.py`, `cnschema-probe.py`,
`check-licenses.py`, `check-issue-refs.py`, `check-doc-claims.py`; `Makefile` and
`ci.yaml` (both lists, in agreement); `docs/BENCHMARKS.md`; 22 READMEs and `CHANGELOG.md`
for the licence offer.

## `.goals` compliance

- SUBSUME, EXTEND, ENHANCE — met. No inferior duplicate survives: six copies of one constant
  folded to one, `require_pin` consolidated from four implementations to one, the per-lane
  `write_checked`/`mkdir_checked` wrappers retained deliberately as adapters that bind the
  knob name (recorded in the design doc, with the reason).
- LOW/NO OPTIONALITY, HARD FAILS — met. `lane-common.sh` hard-fails without `LANE`/
  `LANE_BINARY`; "recorded nowhere" (exit 2) and "the lookup is broken" are separate states;
  no swallowed error found by any pass.
- MAXIMAL UTILITY — met, and this is where the branch cost the most: three separate
  over-refusals were introduced and closed (every emoji path, then the keycap sequences, then
  a legitimate `#` comment documenting an idiom). Each was caught by executing a valid
  neighbour, which CLAUDE.md requires and which the first attempt at each rule lacked.
- MAXIMAL PERFORMANCE / PORTABILITY — no regression; no Cargo feature added; wasm untouched.
- RUST-FIRST and RDF1.2 — not applicable; this is tooling and lane harness work.

## Deferral gate

Mechanical scan over the full diff: one hit, `"comment it out for now"` in
`check-gate-parity.py` — a quoted phrase naming the failure mode the code refuses, and the
rationale for why a commented-out gate must not count. PR body and comments: one hit, a
CodeRabbit report row (`Out of Scope Changes check | Passed`), not branch content. Both are
false positives with the evidence above. No `TODO`, `FIXME`, `unimplemented`, `placeholder`,
`deferred`, `follow-up` or `future work` on the `+` side.

## Verification

- `make check` exit 0, unpiped, reading make's own exit code: 145 `OK:` lines, 10,334 tests.
- `make lubm` EXIT=0 — `14 of 14`, `comparable rows 3 of 14`, corpus and query-set pins
  matched, `oracle: Q1 and Q14 match LUBM's published answers`.
- `make watdiv` EXIT=0 — 10,916,457 triples, `20 of 20`, every one of the twenty per-query
  row counts matched against its own pin, 434,748 rows.
- All three answer-report branches executed, because a reporting branch is where a skip
  hides: matched; refused (C1 pinned to 17 against an engine returning 16 — exit non-zero, no
  SUMMARY published); and no-pin-recorded (`WATDIV_SEED=3`, which is also the over-refusal
  neighbour and a distinguishing one, since seed 3 yields a different query-set digest and a
  different total).
- The census pin refuses what a stamp cannot see: with the stamp edited to AGREE with a
  corrupted census, the lane printed "reusing the extraction already stamped with this
  tarball's digest" and the pin still refused.
- The vacuous-run diagnosis demonstrated both ways with a stand-in answering every query zero
  rows while delegating conversion to the real binary, so the corpus still matched its pin.
- Mutation-checked throughout, not trusted for going green: every fix has its own mechanism
  neutered and its own test confirmed red. Restoring the concatenating digest reddens three
  tests; restoring `$((10#…))` reddens the boundary test; the two-shape stand-in blocklist
  reproduces the committed artifact `created ["SELECT ?s WHERE { ?s ?p ?o } LIMIT 1"]`.

## What a reviewer should know

The merge commit `07fc6f4b` carries 67 lines of hand-written work — `collisions()` and the
variation-selector rule in `check-tracked-paths.py`, authored while resolving the merge rather
than in their own commit. That is poor practice and it is recorded here because a reviewer
reading a 59-file merge would not find it.

Sixteen adversarial passes ran against this branch and none returned clean on its first look
at a commit. The known defects are closed and mutation-checked; what remains documented is
latent, with zero live occurrences. Three rules in this branch had to be corrected two or
three times each — the variation-selector rule failed in alternating directions three times
(every emoji refused, then keycaps refused, then a bare selector accepted) before the fix
encoded the structural question rather than adjusting the predicate's scope. That history is
in the commit messages deliberately.

## Completion-adversary: PASS

The stage-3 completion-adversary ran five times against the merge-into-main result and
returned PASS on the fifth (`b65cc8fb`), with the remaining items stated as cosmetic or
latent with zero live occurrences. Two of those it called "worth a follow-up commit if
convenient" were closed here rather than left, so nothing it raised is outstanding.

Its per-criterion demonstrations, on the production surface:

- **Escape translation matches the shell.** It replaced `python3` on `PATH` with a recorder
  and ran all eight embedded-program fixtures under real `bash`: every extracted body was
  byte-identical to what the shell actually hands Python, modulo a trailing newline on
  heredocs. That is the oracle check the gate cannot perform on itself.
- **Coverage on the real tree**: 17 of 17 live embedded sites extracted, all parsing.
- **Every repair reddens when reverted**, checked by reverting each in a scratch copy:
  the single-line body pattern, the per-launcher unescaping, the over-unescaping correction,
  the `<<-` tab stripping, the exact-delimiter comparison, the raw-text masking, and the
  heredoc-opener-inside-a-body skip.
- **Shell claims verified against bash rather than asserted**: a backslash inside double
  quotes is removed before `$`, a backtick, `"` and a backslash and KEPT before `_`, `n`,
  `t`; `<<-` strips tabs and not spaces; a space-indented delimiter does not close a `<<-`
  heredoc; a tab-indented delimiter does not close a plain one.
- **Numbers re-derived**: 10,778 tracked paths; 42 gates agreeing with 6 registered
  exemptions; 28 `WORKLOAD_PINS` keys; 140 bench tests (`lane_common_laws` 35,
  `make_bench_lanes` 15); `cargo clippy -p purrdf-bench --all-targets -- -D warnings` exit 0.
- **`.deficiencies`**: notice and marker only, no entries, in both trees.

## Verification at the merge commit

`make check` exit 0, unpiped: 146 `OK:` lines. Both lanes were verified EXIT=0 on the
merged code, not on an ancestor, because this round changed `lane-common.sh`'s digest path.

